PR #19, opened 2020-12-25
Encrypted Account Key in Database

sugargoat:
### Motivation

It is best practice for sensitive material to be encrypted at rest. This PR encrypts the account key in the DB, and provides a path forward for existing DBs without encrypted account keys.

Inspired by the work in [MobileCoin 585](https://github.com/mobilecoinfoundation/mobilecoin/pull/585) and [MobileCoin 577](https://github.com/mobilecoinfoundation/mobilecoin/pull/577).

### In this PR
* Adds encryption to WalletDb, including in-memory store of password hash
* Adds `set_password`, `unlock`, and `change_password` to the wallet service
* Updates service endpoints that require DB to be unlocked
* Adds tests to cover new functionality
* Allows passing `password_hash` or `password` and uses argon2 to hash the `password` if provided
* Prevents attempting to submit transactions in offline mode
* Fixes a syncing bug for change txos

[WS-6: Encrypt Account Key At Rest](https://mobilecoin.atlassian.net/browse/WS-6)

### TODO
- [x] Clean up FIXMEs
- [x] Update README


full-service/src/db/mod.rs
```diff
@@ -68,4 +98,259 @@ impl WalletDb {
...
+                    self.logger,
+                    "Database has never been locked. Setting password with new password."
+                );
+                self.set_password_hash(new_password_hash, &conn)?;
+            }
+            EncryptionState::Encrypted => {
+                log::debug!(self.logger, "Database is locked. Verifying old password.");
+                // Attempt to decrypt the test value to confirm if password is correct
+                let expected_val = Self::encrypt(ENCRYPTION_VERIFICATION_VAL, old_password_hash)?;
+                if EncryptionIndicator::verify_password(&expected_val, &conn)? {
```
eranrund, line 185: What happens to existing accounts when the password is changed?

full-service/src/service/wallet.rs
```diff
@@ -190,6 +214,15 @@ pub enum JsonCommandResponse {
     },
 }
 
+// Helper method to escape quotes for json responses.
+fn format_error(e: WalletServiceError) -> String {
+    format!(
```
eranrund, line 219: Is it not possible to use the json! macro to do this properly? This looks like a problematic hack.
https://github.com/serde-rs/json#constructing-json-values

full-service/src/service/wallet.rs
```diff
@@ -33,6 +34,20 @@ pub struct WalletState<
 #[serde(tag = "method", content = "params")]
 #[allow(non_camel_case_types)]
 pub enum JsonCommandRequest {
+    set_password {
+        password: Option<String>,
```
eranrund, line 38: Whats the difference between password and password_hash?

full-service/src/service/wallet_impl.rs
```diff
@@ -98,16 +106,129 @@ impl<
...
             submit_node_offset: Arc::new(AtomicUsize::new(rng.next_u64() as usize)),
+            offline,
             logger,
         }
     }
 
+    // Helper method to expand password to password hash using argon2.
+    fn get_password_hash(
+        &self,
+        password: Option<String>,
```
eranrund, line 117: Why bother with supporting both? This is a pattern I have never seen anywhere else. I think should just accept `password` and always hash it.

full-service/src/service/sync.rs
```diff
@@ -291,6 +291,18 @@ fn sync_thread_entry_point(
...
                     Err(SyncError::AccountNotFound) => {}
 
+                    Err(SyncError::Database(WalletDbError::NoDecryptionKey)) => {
+                        log::trace!(
+                            logger,
+                            "No decryption key. Cannot sync account {:?} Please unlock database.",
+                            account_id,
+                        );
+                        // Queue the account again.
+                        sender
```
eranrund, line 301: Wouldn't this result in 100% cpu usage?

eranrund (commented):
I'm having a hard time understanding whats going on here. Why are sometimes the password hash is passed around, and sometimes it is obtained from the wallet db? Why is this behavior not fully encapsulated in the WalletDb object similarly to how this was done in `mobilecoind`?
It appears that this touches many parts of the code, where I would've expected it to have a much smaller footprint since there is only 1 field that ends up being encrypted.
I'm also confused with the `password`/`password_hash` stuff. I would strongly prefer if that went away and we always just hashed whatever comes from the user over the API.

sugargoat:
@eranrund - thank you for the review!

Address a few of your high level comments:

### Code Footprint and Code Organization
I think one reason this touches so much is because of the architecture with the wallet, where I've broken out the database models into traits. Anywhere that uses the account (which has to be used for scanning, building transactions, and querying various things about txos) now has to make sure it has the password hash to decrypt it. Essentially all those actions are happening right above the DB layer, but it's distributed across multiple files in the db crate because of breaking them out by model.

I've come to desire a refactor on the service side as well, so that there are multiple "services" in the service crate, corresponding to the types of actions that make sense to be grouped together (e.g., the password service would be a good first candidate). I actually experimented with this design in the upcoming gift code PR, and I like it a lot better than the current spaghetti-code wallet_impl.rs file. It also makes it easier to test things logically.

### Password Hash Management
Re: whether the password is gotten from the db or passed in, that mostly has to do with whether we're at the DB or service layer. The DB models take the password hash, while the service (which has an instance of the DB) gets the password hash from `self.wallet_db.get_password_hash()?`. I think I was a little inconsistent in the tests, since the tests both have the hash defined as well as the wallet_db object. I will try to clean that up so that the tests always get the hash from the wallet_db rather than passing it directly.

### Password vs Password Hash
Re: password vs password_hash, I agree that it's a little weird to support both. The main reason for doing so is that the GUI front-end wallet wants to do the hash, and then pass that back to full-service on the backend, without it getting hashed again (mostly to not double the time to hash the password, since part of argon2's value add is being slow to hash). However, I didn't think it would be appropriate to require users to hash their password outside of full-service if they were using full-service without a GUI front-end.

Ultimately, I actually wanted to just go ahead and use SQLite's encryption - because SQLite comes with encryption and password protection out of the box. But I wasn't quite able to plug that in. It may be worth revisiting, so that we don't have to think about managing it in the DB ourselves, per field.

eranrund:
Thank you for the detailed reply!

With regards to passing around the hash - I am still not convinced that is a reasonable solution. Everything that touches the DB, either reading from it or writing from it, passes around a connection object so that it could access the database. There are two alternatives to prevent having to pass the password around all over the place:
1) Instead of passing `conn: &PooledConnection<ConnectionManager<SqliteConnection>>`, we can pass an object that encapsulates a connection + the password, or the connection + a reference (via Arc perhaps) to the top-level DB object
2) We could create a new connection type, that encapsulates `conn: &PooledConnection<ConnectionManager<SqliteConnection>>` inside it, and implement all the necessary Diesel traits to forward the calls to the inner conn. This object would also include details about potential encryption, or a reference to the root object.

With regards to Password vs Password Hash - I don't see a problem hashing twice. This operation happens only once, when the user is providing a password over GRPC and simplifies this entire thing. Since the selling point is making the hash expensive, if it is a bit more expensive that is fine. This happens only when the password is being moved over GRPC which happens very rarely. I'm also not convinced that it is not okay to just receive whatever over GRPC and store the expand_password result, which is cheap and hashed.

sugargoat:
@eranrund -  I really like both proposals for improving the password-hash passing. I'll look at those today.

@davidyhall - would you be ok with simplifying things on the backend by making it treat the hash from the client as the password, and hashing it again?

full-service/src/service/wallet.rs
sugargoat: I ran into issues deriving `serde::Serialize` for WalletServiceError because several of the wrapped errors from external crates don't implement `serde::Serialize`. I will wrap it in `json!().to_string()` to make it clear that it's json, but will still need to do the annoying quote replacement hack. The reason for returning both the stringified `error` and the details from `error.to_string()` is to make sure to get the full formatted displaydoc output.

full-service/src/service/wallet_impl.rs
davidyhall: We want the mobile and desktop wallet to allow configurable Argon2 parameters so the user can control time of encryption (2s, 5s, 10s, etc.) This is important based on the processing speed of their devices (specifically some phones).

That logic doesn't seem right within Full-Service.

The docs should say that 99% of users will want to give it the `password: <string>` option. And you should only use the `passwordHash: <hex encoded string>` if you know why you're using it.

sugargoat:
@eranrund - looking at the current organization, the WalletDb actually only has 3 members - the conn, the password_hash, and the logger. So I think it is already the encapsulated item you suggested. Maybe all that needs to change is where we're passing both the conn and the password hash, just pass the wallet_db. However, I actually don't like that, because I do like it being explicit which methods need to use the password, since some methods (such as submit_tx) don't need the password. This is important to distinguish which methods are useful in offline flows. I'll keep thinking about this.

full-service/src/service/sync.rs
sugargoat: Ah yes - I wanted to make sure the account was queued again so that it would sync when the DB was unlocked, but the account should get picked up in the next iteration of the SyncThread's start loop. I removed this queue step and and verified that the account gets picked up.

davidyhall (commented):
Can you exposed a call to see if the DB has an encrypted? This is useful for GUI wallets, and to prompt either an unlock screen or a create/import screen.

sugargoat:
@davidyhall - done!

full-service/src/service/wallet.rs
sugargoat: Actually, the json macro handles the quotes!

full-service/src/db/mod.rs
sugargoat: That's handled at the service layer - see `PasswordService::change_password`

sugargoat:
@eranrund - I went down the path of passing the WalletDb as opposed to the connection and the password hash, which I also figured would have the nice effect of making it easier to log all of the Db operations from the models.

However, the problem I'm running into is with the connection pool object. The way the models are currently designed, each model operation takes a `&conn`, so that you can potentially do multiple operations within a single database transaction (which is created off of a `conn` object). By making all of these models take the `wallet_db` instead, it means they are calling into the `wallet_db` multiple times to "`get_conn`," which then exhausts the pool and results in timeouts trying to get connections to the db. I also don't expect the db transaction to work quite right spawning new db connections inside the transaction. I tried adding an "active_conn" to the wallet_db, but that ended up complicating things even further.

I pushed my current progress on this front to [sugargoat/refactor-db-conns](https://github.com/sugargoat/full-service/tree/refactor-db-conns), but going down this road would mean a major paradigm shift for the design of the wallet, and I'm not sure it's worth it, when I really like the current properties of being able to orchestrate the model operations together in arbitrary ways inside of a db transaction. I also do still appreciate having the password_hash parameter as a clear indicator when a model operation will be interacting with decrypted account key data. This is also useful when composing multiple operations together.

Deeply appreciate your thoughts on this!

eranrund:
@sugargoat the idea is to create a new object that points at a connection and the wallet db for using common functionality such as getting the current password. You are right that passing the wallet db itself won't work since you have to pass a specific connection and adding an "active connection" field is not a good pattern and will break with threads and such.

What I was suggesting is creating a new object:
```
struct WalletDbConn {
pub conn: PooledConnection<ConnectionManager<SqliteConnection>>
pub wallet_db: Arc<WalletDb>
}
```
and passing that around instead of the `conn: &PooledConnection<ConnectionManager<SqliteConnection>>` - so essentially the change would be to
1) Add this object somewhere
2) Search-replace `conn: &PooledConnection<ConnectionManager<SqliteConnection>>` with `conn: &WalletDbConn` and inside the methods use `conn.conn` and `conn.wallet_db.get_password() / conn.wallet_db.is_encrypted() / etc` where needed
3) Replace `pool.get_conn` with something that wraps that with the new `WalletDbConn`.

There might be small nuances I am missing but that is the general idea.

eranrund:
It might also be easier to separate the encryption stuff into its own object and then wrap that inside an Arc and have it live inside the WalletDb, and have the WalletDbConn point to a connection and the Arc'd EncryptionProvider thing.

sugargoat:
@eranrund - Ok I see - thank you! I will give that a shot!

full-service/src/db/mod.rs
```diff
@@ -43,15 +47,28 @@ pub fn b58_decode(b58_public_address: &str) -> Result<PublicAddress, WalletDbErr
     Ok(PublicAddress::try_from(pubaddr_proto).unwrap())
 }
 
+/// A struct encapsulating a single connection, which can also be used for encryption and logging.
```
sugargoat, line 50: @eranrund - please take a look at the WalletDbConnManager and let me know what you think!

full-service/src/db/encryption_provider.rs
```diff
@@ -0,0 +1,256 @@
+// Copyright (c) 2020 MobileCoin Inc.
```
sugargoat, line 1: @eranrund - refactored the encryption_provider functionality out of the wallet_db. LMK what you think!

full-service/src/service/password_manager.rs
```diff
@@ -0,0 +1,440 @@
+// Copyright (c) 2020 MobileCoin Inc.
```
sugargoat, line 1: @eranrund - I also refactored the password management functionality that was originally in the wallet_service.impl into its own service trait. LMK what you think about this pattern, because I'd like to use it going forward with gift codes, and eventually refactor the other logical services out of wallet_service.impl.

full-service/src/service/password_manager.rs
```diff
@@ -0,0 +1,440 @@
...
+    fn is_locked(&self) -> Result<Option<bool>, PasswordServiceError>;
+
+    /// Utility method to call on service methods which require the database to be unlocked
+    fn verify_unlocked(&self) -> Result<(), PasswordServiceError>;
+}
+
+impl<T, FPR> PasswordService for WalletService<T, FPR>
+where
+    T: BlockchainConnection + UserTxConnection + 'static,
+    FPR: FogPubkeyResolver + Send + Sync + 'static,
```
eranrund, line 119: Is this needed?

full-service/src/service/password_manager.rs
eranrund: I think this pattern is good. I will review the file more closely once the _hash stuff goes away.

full-service/src/service/wallet.rs
```diff
@@ -122,6 +139,18 @@ pub enum JsonCommandRequest {
...
+        success: bool,
+    },
+    unlock {
+        success: bool,
+    },
+    change_password {
+        success: bool,
+    },
+    is_locked {
+        is_locked: Option<bool>,
```
eranrund, line 152: Why is this optional?

full-service/src/service/wallet.rs
```diff
@@ -196,6 +225,11 @@ pub enum JsonCommandResponse {
     },
 }
 
+// Helper method to format displaydoc errors in json.
+fn format_error<T: std::fmt::Display + std::fmt::Debug>(e: T) -> String {
+    json!({"error": format!("{:?}", e), "details": e.to_string()}).to_string()
```
eranrund, line 230: Yay!

eranrund: I actually think error and details should be switched because the debug formatting will have more information.

full-service/src/service/wallet_impl.rs
```diff
@@ -57,39 +58,43 @@ pub struct WalletService<
     T: BlockchainConnection + UserTxConnection + 'static,
     FPR: FogPubkeyResolver + Send + Sync + 'static,
 > {
-    wallet_db: WalletDb,
+    pub wallet_db: WalletDb,
```
eranrund, line 61: Whats accessing this externally?

full-service/src/service/wallet_impl.rs
```diff
@@ -57,39 +58,43 @@ pub struct WalletService<
...
     ledger_db: LedgerDB,
     peer_manager: McConnectionManager<T>,
     network_state: Arc<RwLock<PollingNetworkState<T>>>,
     fog_pubkey_resolver: Option<Arc<FPR>>,
     _sync_thread: SyncThread,
     /// Monotonically increasing counter. This is used for node round-robin selection.
     submit_node_offset: Arc<AtomicUsize>,
-    logger: Logger,
+    offline: bool,
+    pub logger: Logger,
```
eranrund, line 70: Whats accessing this externally?

full-service/src/service/wallet_impl.rs
```diff
@@ -198,41 +207,42 @@ impl<
...
-            )?;
-            Ok(decorated_account)
-        })?)
+        Ok(conn_manager
+            .conn
+            .transaction::<JsonAccount, WalletServiceError, _>(|| {
+                Account::get(&AccountID(account_id_hex.to_string()), &conn_manager.conn)?
+                    .update_name(name, &conn_manager.conn)?;
+
+                let local_height = self.ledger_db.num_blocks()?;
```
eranrund, line 231: Its pretty weird to have all this stuff inside a method called `update_account_name`. Why is this here? Is this duplicated from anywhere?

eranrund: Maybe there should be a `self.get_decorated_account(account_id)` if you want to return the entire account as a response.

eranrund (commented):
This is going in a good direction! Thank you for continuously iterating on this!

full-service/src/db/account.rs
```diff
@@ -231,42 +278,76 @@ impl AccountModel for Account {
...
+                let unspent = Txo::list_by_status(
+                    &account_id_hex.to_string(),
+                    TXO_UNSPENT,
+                    &conn_manager.conn,
+                )?
                 .iter()
                 .map(|t| t.value as u128)
                 .sum::<u128>();
-            let pending = Txo::list_by_status(&account_id_hex.to_string(), TXO_PENDING, conn)?
+                let pending = Txo::list_by_status(
```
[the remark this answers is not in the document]
sugargoat: Yeah - I think they should all eventually take conn_manager/conn_context, but if it's ok to reduce the footprint here, I'd like to introduce that on-demand as we encrypt more fields.

full-service/src/db/encryption_indicator.rs
```diff
@@ -0,0 +1,151 @@
...
+        encrypted_verification_value: &[u8],
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<(), WalletDbError> {
+        use crate::db::schema::encryption_indicators as encryption_table;
+
+        Ok(conn.transaction::<(), WalletDbError, _>(|| {
+            let verification_val_insertable = encrypted_verification_value.to_vec();
+            let new_indicator = NewEncryptionIndicator {
+                encrypted: true,
+                verification_value: Some(&verification_val_insertable),
```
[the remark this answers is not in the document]
sugargoat: It does look like the `#[derive(Insertable)]` "NewEncryptionIndicator" struct can take `&'a [u8]` - I had only seen examples with vecs. Will fix that up.

full-service/src/db/mod.rs
```diff
@@ -68,4 +85,170 @@ impl WalletDb {
...
+    pub fn change_password(
+        &self,
+        old_password_hash: &[u8],
+        new_password_hash: &[u8],
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<(), WalletDbError> {
+        Ok(self
+            .encryption_provider
+            .change_password(old_password_hash, new_password_hash, &conn)?)
+    }
```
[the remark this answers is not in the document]
sugargoat: This is the bottom of the change_password stack, so the re-encryption occurs at the password_manager service level, in `PasswordManager::change_password`

full-service/src/service/wallet_impl.rs
sugargoat: This unfortunately was a requirement when I refactored out the PasswordManager service to its own trait. Maybe I'm missing some rust trick so that I can access the private members of the WalletService while implementing the PasswordManager trait for it?

full-service/src/service/wallet.rs
sugargoat: Perhaps I'm overloading the Option and this should be an enum - the meaning is:

None => Never been locked (no password set)
Some(true) => Is locked
Some(false) => Is unlocked

full-service/src/db/txo.rs
```diff
@@ -559,13 +591,19 @@ impl TxoModel for Txo {
                     txo_details.received_to_account = Some(account_txo_status.clone());
                     // Get the subaddress details if assigned
                     let assigned_subaddress = if let Some(subaddress_index) = txo.subaddress_index {
-                        let account: Account =
-                            Account::get(&AccountID(account_txo_status.account_id_hex), conn)?;
-                        let account_key: AccountKey = mc_util_serial::decode(&account.account_key)?;
+                        let account: Account = Account::get(
```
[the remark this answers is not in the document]
sugargoat: I believe there are some instances where I want to get the Account object for other fields, but I don't need to decrypt the key -> will do a quick audit to make sure that's still the case.

full-service/src/db/encryption_indicator.rs
```diff
@@ -0,0 +1,151 @@
...
+    QueryDsl, RunQueryDsl,
+};
+
+#[derive(Debug)]
+pub enum EncryptionState {
+    /// Database has never been locked, and set_password should be called.
+    Empty,
+    /// Database is encrypted.
+    Encrypted,
+    /// Database is unencrypted. This is for databases that existed before we added encryption.
```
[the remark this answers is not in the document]
sugargoat: This is intended for databases that exist today which have contents that have not been encrypted. Once the user sets the password, then it never goes back to the unencrypted paradigm.

The encryption is mandatory only if using the interfaces which require encryption. It is also the intent that you could run full-service for methods that don't require encryption and never have to set a password (e.g. submit-transaction, in an offline flow).

full-service/src/service/wallet.rs
sugargoat: It's sort of counter-intuitive. The debug only prints the name of the error, and the to_string() prints the full displaydoc string.

eranrund: Ha, yeah, I can see that being the case. My mistake then :)

full-service/src/db/encryption_indicator.rs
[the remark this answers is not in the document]
eranrund: 1. Are there databases that exist today?
2. I am not sure I understand `The encryption is mandatory for new full-service instances only if using the interfaces which require encryption. ` - does that mean that in order to add an account to a new database I must first set up encryption?

full-service/src/service/wallet.rs
eranrund: Yeah, this should be an enum

full-service/src/service/wallet_impl.rs
eranrund: Can you point me at where this is being used in the trait implementation?

full-service/src/db/encryption_indicator.rs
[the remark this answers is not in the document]
sugargoat: 1. There may be, though I'm not in contact with any current users of full-service.
2. Your understanding is correct - before adding an account, you will need to set the password.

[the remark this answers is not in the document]
eranrund: This project is early enough in its life that I would argue to not worry about backward support in favor of simplifying a whole bunch of logic.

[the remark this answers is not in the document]
davidyhall: As long as there is a very clear migration path, rare breaking changes can be tolerated.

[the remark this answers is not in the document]
eranrund: If we really think this is a concern we can provide a separate utility that migrates the old db to the new one. Realistically, its very unlikely anyone is using this in the wild and we're introducing code here that we'll just have to keep maintaining for what I think is no good reason.

full-service/src/db/encryption_provider.rs
```diff
@@ -0,0 +1,305 @@
...
+        Ok(cur_password_hash.clone())
+    }
+
+    /// Never locked is considered true, because some functionality of the wallet is disabled.
+    pub fn is_locked(
+        &self,
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<bool, WalletDbError> {
+        match self.get_locked_status(conn)? {
+            LockedStatus::IsLocked => Ok(true),
```
sugargoat, line 109: IsLocked -> Locked everywhere

full-service/src/db/account.rs
```diff
@@ -231,42 +278,76 @@ impl AccountModel for Account {
...
+                }
+                EncryptionState::Unencrypted => mc_util_serial::decode(&self.account_key)?,
+                EncryptionState::Encrypted => {
+                    if !conn_manager.encryption_provider.is_unlocked()? {
+                        return Err(WalletDbError::NoDecryptionKey);
+                    }
+                    let decrypted_account_key = conn_manager
+                        .encryption_provider
+                        .decrypt(&self.account_key)?;
+                    mc_util_serial::decode(&decrypted_account_key)?
```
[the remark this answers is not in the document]
sugargoat: Done!

full-service/src/service/wallet_impl.rs
sugargoat: Here's an example: https://github.com/mobilecoinofficial/full-service/blob/880f94b73ffd8c4fb846430d1011367cc8758c76/full-service/src/service/password_manager.rs#L121

full-service/src/service/wallet_impl.rs
sugargoat: Here's an example where the trait impl is using it: https://github.com/mobilecoinofficial/full-service/blob/880f94b73ffd8c4fb846430d1011367cc8758c76/full-service/src/service/password_manager.rs#L126

full-service/src/db/encryption_indicator.rs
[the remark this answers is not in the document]
sugargoat: I updated this enum variant description to: 

```    
/// Database is unencrypted. This is an invalid state where some accounts are not encrypted.
```

And wherever that enum variant is returned, it's an error, with instructions to migrate.

full-service/src/db/encryption_indicator.rs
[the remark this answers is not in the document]
sugargoat: Also fixed up the other New* models.

full-service/src/db/encryption_provider.rs
```diff
@@ -0,0 +1,256 @@
...
+        password_hash: &[u8],
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<(), WalletDbError> {
+        // No need to check db state if we're already unlocked
+        if self.is_unlocked()? {
+            return Ok(());
+        }
+
+        // Check whether encrypted, and if so, then attempt to unlock
+        match EncryptionIndicator::get_encryption_state(&conn)? {
```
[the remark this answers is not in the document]
sugargoat: Now using the "get_locked_status" to properrly handle database locked state in set_password

full-service/src/db/txo.rs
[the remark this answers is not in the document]
sugargoat: One example is in sync: https://github.com/mobilecoinofficial/full-service/blob/880f94b73ffd8c4fb846430d1011367cc8758c76/full-service/src/service/sync.rs#L335

We don't need the decrypted account key, but we need some other fields from the account.

full-service/src/service/password_manager.rs
sugargoat: I believe so - without it, I get "Wrong number of type arguments"

full-service/src/db/encryption_provider.rs
```diff
@@ -0,0 +1,256 @@
...
+        &self,
+        password_hash: &[u8],
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<(), WalletDbError> {
+        {
+            let mut cur_password_hash = self.password_hash.lock()?;
+            *cur_password_hash = password_hash.to_vec();
+        }
+        // Encrypt the verification value and set in the DB
+        let verification_value =
```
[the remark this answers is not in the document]
sugargoat: Good catch - I switched the order so that we only set the password hash if encrypting the test value succeeded.

full-service/src/service/wallet_impl.rs
eranrund: Ha, I'm surprised that is the case. I guess Rust considers it a separate module if its in a separate file.

README.md
```diff
@@ -48,13 +48,165 @@ MobileCoin Full Service is available under open-source licenses. Look for the [L
...
+  "error": "Database(PasswordFailed)",
+  "details": "Error interacting with the database: Password failed"
+}
+```
+
+#### Is Locked
+
+Check whether the database is currently locked.
+
+```sh
```
eranrund, line 193: This is not working for me:
```
curl -s localhost:9090/wallet \
  -d '{
        "method": "is_locked",
      }' \
  -X POST -H 'Content-type: application/json'

            <!DOCTYPE html>
            <html lang="en">
            <head>
                <meta charset="utf-8">
                <title>422 Unprocessable Entity</title>
            </head>
            <body align="center">
                <div role="main" align="center">
                    <h1>422: Unprocessable Entity</h1>
                    <p>The request was well-formed but was unable to
                be followed due to semantic errors.</p>
                    <hr />
                </div>
                <div role="contentinfo" align="center">
                    <small>Rocket</small>
                </div>
            </body>
            </html>
```

eranrund:
```
2021-01-21 19:09:05.452479 UTC ERRO Couldn't parse JSON body: Error("unknown variant `is_locked`, expected one of `set_password`, `unlock`, `change_password`, `get_locked_status`, `create_account`, `import_account`, `get_all_accounts`, `get_account`, `update_account_name`, `delete_account`, `get_all_txos_by_account`, `get_txo`, `get_wallet_status`, `get_balance`, `create_address`, `get_all_addresses_by_account`, `send_transaction`, `build_transaction`, `submit_transaction`, `get_all_transactions_by_account`, `get_transaction`, `get_transaction_object`, `get_txo_object`, `get_block_object`, `get_proofs`, `verify_proof`", line: 2, column: 29), mc.app: full-service, mc.module: rocket_contrib::json, mc.src: /Users/eran/.cargo/registry/src/github.com-1ecc6299db9ec823/rocket_contrib-0.4.6/src/json.rs:153
```

eranrund (commented):
This is great progress. It is not working great locally (communicating details over Slack).
I also do not think this can be considered production ready until the pool max connections issue is resolved.
Somewhat related, I think the repo should have a big notice in the README saying this is not yet ready for production, just in case someone gets ambitious.

README.md
sugargoat: Ah thank you - need to update the documentation. It's `get_locked_status` now

full-service/src/bin/main.rs
```diff
@@ -109,6 +109,7 @@ fn main() {
             network_state,
             config.get_fog_pubkey_resolver(logger.clone()).map(Arc::new),
             config.num_workers,
+            config.offline,
```
mfaulk, line 112: What is the use case for offline mode? I don't really understand what that is for.

full-service/src/db/encryption_indicator.rs
```diff
@@ -0,0 +1,177 @@
+// Copyright (c) 2020 MobileCoin Inc.
+
+//! DB impl for the Locked model.
+//!
+//! There are two "locked" states for a DB, in order to support a mode in which you can
+//! still submit transactions without accounts (and therefore without a password).
```
mfaulk, line 6: Can you say more about what the idea here is? For example, when/how would a user submit transactions without access to the accounts stored in the wallet?

full-service/src/db/encryption_indicator.rs
```diff
@@ -0,0 +1,177 @@
...
+    QueryDsl, RunQueryDsl,
+};
+
+#[derive(Debug)]
+pub enum EncryptionState {
+    /// Database has never been locked, and set_password should be called.
+    Empty,
+    /// Database is encrypted.
+    Encrypted,
+    /// Database is unencrypted. This is an invalid state where some accounts are not encrypted.
```
mfaulk, line 28: When would the database be in the `Unencrypted` state, if that is an invalid state?

full-service/src/db/encryption_indicator.rs
```diff
@@ -0,0 +1,177 @@
+// Copyright (c) 2020 MobileCoin Inc.
+
+//! DB impl for the Locked model.
```
mfaulk, line 3: I generally expect the word "locked" to be associated with a mutex, but here it seems to mean something like "encrypted-at-rest" or "the user has authenticated".

chrisgerda (approved):

chrisgerda (approved):

full-service/src/db/encryption_provider.rs
```diff
@@ -0,0 +1,304 @@
...
+            &self.get_password_hash()?,
+        )?)
+    }
+
+    /// Encrypt data.
+    pub fn encrypt_with_password(
+        plaintext_bytes: &[u8],
+        password_hash: &[u8],
+    ) -> Result<Vec<u8>, WalletDbError> {
+        let (key, nonce) = Self::expand_password(password_hash)?;
```
mfaulk, line 228: This looks like it uses the same key and nonce for every encryption done for a given password, and I believe AES-GCM breaks if a (key,nonce) pair is ever reused, e.g. https://chris-wood.github.io/2016/12/25/GCM-SIV.html .

If so, we could choose a more nonce-misuse-resistant algorithm (AES-GCM-SIV, ChaCha20?), or be careful to use a distinct nonce for each encryption. I think its common to use a counter as the nonce (incremented for each encryption, so always unique), although that will take more book-keeping on the DB side to maintain a global counter for every encryption.

full-service/src/service/password_manager.rs
```diff
@@ -0,0 +1,377 @@
...
+        old_password: String,
+        new_password: String,
+    ) -> Result<bool, PasswordServiceError> {
+        // Make sure DB is unlocked to continue.
+        self.unlock(old_password.clone())?;
+
+        let old_password_hash = self.get_password_hash(old_password)?;
+        let new_password_hash = self.get_password_hash(new_password)?;
+
+        // Re-encrypt all of our accounts with the new password hash
```
mfaulk, line 170: Having to read, decrypt, encrypt, and write every record in the DB won't scale well. That may be OK for accounts (assuming that a wallet will only have one or a few accounts), but this approach probably wouldn't be a good fit if we want to later do encryption-at-rest for larger wallet datasets, like received outputs.

I think it's more standard to use a data-encryption-key (DEK) and a key-encryption-key (KEK) to solve this problem (aka "envelope encryption"), where the "master" KEK is used to decrypt the DEK, and then the DEK is used to encrypt/decrypt each piece of stored data. Then, changing the password is equivalent to decrypting the DEK using the old KEK and re-encrypting it using the new KEK (which is derived from the new password). The DEK doesn't change, so it is not necessary to re-encrypt each piece of stored data.

sugargoat:
We're going to close this for now and revisit the idea with envelope encryption. The aspects of this PR that can be scavenged will appear in subsequent PRs.

PR #21, opened 2021-01-08
Gift codes

sugargoat:
### Motivation

Gift codes are miniature accounts that contain a single Txo, for use as delivering gifts of MOB.

### In this PR
* Adds the Gift Code data model
* Adds the Gift Code Service, including create, consume, and check_status.

[WS-50](https://mobilecoin.atlassian.net/browse/WS-50)


API.md
```diff
@@ -375,6 +376,39 @@ Txo Spent from One Account to Another in the Same Wallet
...
+| object | string, value is "gift_code" | String representing the object's type. Objects of the same type share the same value
+
+#### Example Object
+
+```
+{
+  "object": "account",
+  "gift_code": "1916a9b3...",
+  "entropy": "1916a9b3...",
+  "value": "2000000000000",
```
eranrund, line 401: Having the value feels questionable to me since there's no guarantee it actually matches the TxOut.

README.md
```diff
@@ -48,13 +48,150 @@ MobileCoin Full Service is available under open-source licenses. Look for the [L
...
+If you receive the following error, it means you are attempting an action that requires the database to be unlocked:
+
+```sh
+{
+  "error": "Database(NoDecryptionKey)",
+  "details": "Error interacting with the database: No decryption key in database."
+}
+```
+
+#### Change Password
```
eranrund, line 156: I don't think this is available in this branch...

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+    /// Returns:
+    /// * JsonSubmitResponse from submitting the gift code transaction to the network
+    /// * Entropy of the gift code account, hex encoded
+    fn build_and_submit_gift_code(
+        &self,
+        from_account_id: String,
+        value: String,
+        name: Option<String>,
+        input_txo_ids: Option<&Vec<String>>,
+        fee: Option<String>,
```
eranrund, line 77: Pretty weird that this is a `String` in Rust world.

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+    UnexpectedValueInGiftCodeTxo(u64),
+}
+
+pub trait GiftCodeService {
+    /// Builds a new gift code.
+    ///
+    /// Building a gift code requires the following steps:
+    ///  1. Create a new account to receive the funds
+    ///  2. Send a transaction to the new account
+    ///  3. Wait for the transaction to land
```
eranrund, line 65: It doesn't look like the implementation of `build_and_submit_gift_code` is waiting for anything, it just submits the transaction.

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+        )?;
+
+        Ok((submit_response, entropy_str))
+    }
+
+    fn register_gift_code(
+        &self,
+        transaction_id_hex: String,
+        gift_code_entropy: String,
+        poll_interval: Option<u64>,
```
eranrund, line 190: Its implicit that these are seconds. This should be a `Duration`

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+            .iter()
+            .map(|g| JsonGiftCode {
+                object: "gift_code".to_string(),
+                gift_code: g.gift_code_b58.clone(),
+                entropy: hex::encode(
+                    &g.get_decrypted_entropy(
+                        &self
+                            .wallet_db
+                            .get_password_hash()
+                            .expect("Could not get password hash"),
```
eranrund, line 300: I don't think expect is appropriate.

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+                    .expect("Could not decrypt entropy"),
+                ),
+                value: g.value.to_string(),
+                memo: g.memo.clone(),
+            })
+            .collect())
+    }
+
+    fn check_gift_code_status(&self) -> Result<(), WalletServiceError> {
+        Ok(())
```
eranrund, line 311: What is this for?

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+            "Consuming gift code {:?} to account_id {:?} at address {:?}",
+            gift_code_b58,
+            account_id_hex,
+            assigned_subaddress_b58
+        );
+
+        // Get the components of the gift code from the printable wrapper
+        let wrapper =
+            mc_mobilecoind_api::printable::PrintableWrapper::b58_decode(gift_code_b58.clone())
+                .unwrap();
```
eranrund, line 334: Unwrap is not reasonable here imo.

eranrund: This applies to all the other unwraps that are sprinkled in this method.

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+                    json_gift_code_account.account_id,
+                    json_gift_code_account.offset_count,
+                )
+            }
+            Err(e) => return Err(e.into()),
+        };
+
+        // Note that we now need to allow the sync thread to catch up for this TXO so that we can
+        // make sure the subaddress is assigned, rendering the Txo spendable.
+        std::thread::sleep(std::time::Duration::from_secs(5));
```
eranrund, line 388: I don't think this is reasonable, what happens if it takes more than 5 seconds?

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+            );
+        }
+        if txos[0].txo.value as u64 != value {
+            return Err(GiftCodeServiceError::UnexpectedValueInGiftCodeTxo(
+                txos[0].txo.value as u64,
+            )
+            .into());
+        }
+
+        log::info!(self.logger, "\x1b[1;33m GOT TXOS = {:?}\x1b[0m", txos);
```
eranrund, line 425: We don't do colors in the log messages

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+            TransactionLog::get(&submit_response.transaction_id, &self.wallet_db.get_conn()?)?;
+
+        // Add the consumed gift code to our gift code store. If we also own this gift code, then update.
+        match GiftCode::get(&gift_code_b58, &self.wallet_db.get_conn()?) {
+            Ok(gc) => {
+                log::info!(self.logger, "Updating existing gift code consume_log_id");
+                gc.update_consume_log_id(transaction_log.id, &self.wallet_db.get_conn()?)?
+            }
+            Err(WalletDbError::GiftCode(GiftCodeDbError::GiftCodeNotFound(_))) => {
+                log::info!(self.logger, "Registering gift code");
```
eranrund, line 454: This should have more information like the b58 or the id or anything that makes it easier to debug, this applies to other log messages in this method.

eranrund (changes requested):
I think the gift code stuff looks mostly reasonable but to be honest it is a lot of code so I can't guarantee I didn't miss anything or that this solves whatever it is this is set to solve. The issues I have with this PR are:
1) the JSON String/u64 spills into places it shouldn't.
2) This includes all the encryption stuff/password hash stuff that I wasn't happy about, making it both harder to review, and contain code that I think is not optimal
3) Too liberal use of unwraps/expects
I'm trusting you that the happy flow actually works as expected (both in cases a person gifts themselves for whatever reason, and when two unrelated instances of the service gift each-other), I haven't tested it.

full-service/src/service/gift_code.rs
```diff
@@ -0,0 +1,644 @@
...
+        input_txo_ids: Option<&Vec<String>>,
+        fee: Option<String>,
+        tombstone_block: Option<String>,
+        max_spendable_value: Option<String>,
+    ) -> Result<(JsonSubmitResponse, String), WalletServiceError>;
+
+    /// After a gift code has been built and submitted, this method polls for the funded Txo to
+    /// hit the ledger, and then constructs the gift code given the entropy, txo_public_key, and
+    /// the memo.
+    fn register_gift_code(
```
eranrund, line 85: This name is confusing to me

sugargoat:
Closing for now - will re-open when this is ready for review.

PR #23, opened 2021-01-15
WIP: Mirror code for full-service

tsegaran:
Soundtrack of this PR: [Hall of mirrors](https://www.youtube.com/watch?v=G0DQLBzUEKw)

This is a work in progress, several important features are not implemented but the structure of the code should be correct.

### Motivation

In some setups, for security reasons, it's necessary to run full service with restricted inbound network access while still allowing for certain commands to be run from a different set of servers. We solve this with a mirror service, the private portion running alongside full service, which polls a public service for requests.

### In this PR
* Adds private mirror service to be run alongside full service. The private service does not accept inbound requests.
* Adds public mirror service which accepts and queues inbound requests and be polled by the private mirror service
* Sets up a whitelist and change rules for what the private mirror will accept


christian-oudard:
Development is continuing here: https://github.com/mobilecoinofficial/mobilecoind-mirror/pull/13

PR #24, opened 2021-02-05
[WS-31] Mocking WalletService

mfaulk:
Soundtrack of this PR: [Max Cooper - Repetition](https://youtu.be/nO9aot9RgQc)

### Motivation

Mock objects make it easy to unit test individual components without laboriously constructing and initializing real dependencies. This is particularly handy for testing API-layer functionality.

### In this PR
[x] Extracts a Wallet trait,
[x] Uses mockall to generate MockWallet,
[x] Updates WalletState to hold an instance of Wallet,
[x] Updates wallet API unit tests to use mocked WalletState,

This PR is smaller than it looks because it contains some refactors:
- Splits error.rs into separate files so that the errors are defined closer to the code that generates them.
- Moves "API-related" things into a new api module (e.g. wallet.rs --> api/wallet.rs).
- Moves the large enums JsonCommandRequest and JsonCommandResponse into separate files (api/json_rpc_request.rs and api/json_rpc_response).
- Splits the JSON-catch-all file decorated_types.rs into separate files under the `api` module.

### Future Work
* Move/Add unit tests that test the wallet service independent of the API.
* Try to encapsulate JSON and String-y data into a JSON-RPC API Layer, so that the application layer (i.e. WalletService) works with normal Rust objects. That would make developing and testing the application layer easier, and would also make it easy to add other API implementations (gRPC, command line, GUI).

PR #30, opened 2021-02-11
WS-78: TxProposal Stringify

sugargoat:
Soundtrack of this PR: [Undone](https://www.youtube.com/watch?v=LHQqqM5sr7g)

### Motivation

The JsonTxProposal object has several u64s in various objects contained therein. This change introduces a StringifiedJsonTxProposal which has all internal u64s represented as strings.

### In this PR
* StringifiedJsonTxProposal and StringifiedUnspentTxOut

[WS-78](https://mobilecoin.atlassian.net/browse/WS-78)

### TODO for this PR
- [ ] Update mobilecoin submodule once [upstream PR](https://github.com/mobilecoinfoundation/mobilecoin/pull/719) is merged

### DIFF
--- a/full-service/src/service/decorated_types.rs
+++ b/full-service/src/service/decorated_types.rs
@@ -14,6 +14,7 @@ use mc_mobilecoind_json::data_types::{
 };
 use serde_derive::{Deserialize, Serialize};
 use serde_json::Map;
+use std::convert::TryFrom;
 
 #[derive(Deserialize, Serialize, Default, Debug)]
 pub struct JsonCreateAccountResponse {
@@ -313,17 +314,31 @@ impl From<&JsonUnspentTxOut> for StringifiedJsonUnspentTxOut {
     }
 }
 
-impl From<&StringifiedJsonUnspentTxOut> for JsonUnspentTxOut {
-    fn from(src: &StringifiedJsonUnspentTxOut) -> Self {
-        Self {
+impl TryFrom<&StringifiedJsonUnspentTxOut> for JsonUnspentTxOut {
+    type Error = String;
+
+    fn try_from(src: &StringifiedJsonUnspentTxOut) -> Result<JsonUnspentTxOut, String> {
+        Ok(Self {
             tx_out: src.tx_out.clone(),
-            subaddress_index: src.subaddress_index.parse::<u64>().unwrap(),
+            subaddress_index: src
+                .subaddress_index
+                .parse::<u64>()
+                .map_err(|err| format!("Failed to parse u64 from subaddress_index: {}", err))?,
             key_image: src.key_image.clone(),
             value: src.value.clone(),
-            attempted_spend_height: src.attempted_spend_height.parse::<u64>().unwrap(),
-            attempted_spend_tombstone: src.attempted_spend_tombstone.parse::<u64>().unwrap(),
+            attempted_spend_height: src.attempted_spend_height.parse::<u64>().map_err(|err| {
+                format!("Failed to parse u64 from attempted_spend_height: {}", err)
+            })?,
+            attempted_spend_tombstone: src.attempted_spend_tombstone.parse::<u64>().map_err(
+                |err| {
+                    format!(
+                        "Failed to parse u64 from attempted_spend_tombstone: {}",
+                        err
+                    )
+                },
+            )?,
             monitor_id: src.monitor_id.clone(),
-        }
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


full-service/src/service/decorated_types.rs
```diff
@@ -284,3 +286,92 @@ pub struct JsonProof {
...
+}
+
+impl From<&StringifiedJsonUnspentTxOut> for JsonUnspentTxOut {
+    fn from(src: &StringifiedJsonUnspentTxOut) -> Self {
+        Self {
+            tx_out: src.tx_out.clone(),
+            subaddress_index: src.subaddress_index.parse::<u64>().unwrap(),
+            key_image: src.key_image.clone(),
+            value: src.value.clone(),
+            attempted_spend_height: src.attempted_spend_height.parse::<u64>().unwrap(),
```
eranrund, line 322: I think `unwrap` is unacceptable in this context, this should be a `try_from`

full-service/src/service/decorated_types.rs
```diff
@@ -284,3 +286,92 @@ pub struct JsonProof {
...
+    }
+}
+
+#[derive(Deserialize, Serialize, Default, Debug)]
+pub struct StringifiedJsonTxProposal {
+    pub input_list: Vec<StringifiedJsonUnspentTxOut>,
+    pub outlay_list: Vec<JsonOutlay>,
+    pub tx: JsonTx,
+    pub fee: String,
+    pub outlay_index_to_tx_out_index: Vec<(String, String)>,
```
eranrund, line 335: Why do these need to be strings? They will never exceed a very small number.

full-service/src/service/wallet.rs
```diff
@@ -1064,4 +1067,50 @@ mod tests {
         let value = txo.get("value_pmob").unwrap().as_str().unwrap();
         assert_eq!(value, "42000000000000");
     }
+
+    /*
```
eranrund, line 1071: Leftovers?

eranrund (changes requested):
I'm not sure why this is needed, but regardless of that I don't think it is okay to use `unwrap` on data that can come from users/the outside world.

full-service/src/service/decorated_types.rs
sugargoat: I think we need to keep it consistent - they are u64 in the tx_proposal object, and we are categorically saying that all u64s are strings in json

eranrund: Fair, although I'd argue that maybe we should change them to u16 or u32 since there's no world in which there are that many outlays/txouts...

sugargoat: Fair will fix up in [719](https://github.com/mobilecoinfoundation/mobilecoin/pull/719)

full-service/src/service/decorated_types.rs
```diff
@@ -284,3 +286,92 @@ pub struct JsonProof {
     pub txo_id: String,
     pub proof: String,
 }
+
+#[derive(Deserialize, Serialize, Default, Debug)]
+pub struct StringifiedJsonUnspentTxOut {
```
mfaulk, line 291: This doesn't feel good to me. I generally prefer things with real types over String-y data, so "StringifiedJson..." feels like a double red flag.

I may be missing the point, but I think I'd rather take the approach that these "decorated types" are only used to translate json-rpc API requests and responses to/from JSON. In that case, it seems like these types can have more strongly typed fields. It might also be worth using something like https://github.com/Keats/validator to enforce constraints on user-supplied data.

mfaulk: Tangentially, I started moving the decorated_types into separate files under a new json_rpc module in the mock wallet PR, for example:

https://github.com/mobilecoinofficial/full-service/pull/24/files?file-filters%5B%5D=.rs&hide-deleted-files=true#diff-7456eb03019837a851279b592b4477613c877fb2592dd07d364bf8b831b6bbbd

As future work, I was sort of imagining replacing the String-y fields with stricter fields, adding more tests and fuzzing around coversion to/from JSON, etc.

full-service/src/service/decorated_types.rs
```diff
@@ -8,7 +8,9 @@ use crate::db::{
     txo::TxoDetails,
 };
 use chrono::{TimeZone, Utc};
-use mc_mobilecoind_json::data_types::{JsonTxOut, JsonTxOutMembershipElement};
+use mc_mobilecoind_json::data_types::{
```
mfaulk, line 11: It seems odd to me that full-service depends on mobilecoind_json. Full-service is supposed to replace mobilecoind and mobilecoind_json, right? Maybe more importantly, its awkward that the api of this service is partly defined in a different repo.

sugargoat: Yeah - the goal here was to make TxProposal cross-compatible with the wallets, so that wherever you built your TxProposal, you could submit it with a compatible wallet.

full-service/src/service/decorated_types.rs
sugargoat: Let's chat about this - the issue is that json and javascript don't support u64, so if you naively try to return u64s, you could lose precision. The rust types are the strongly typed structs.

mfaulk: Hmm. So, I think its fine to represent a 64-bit number as a String in JSON, but I don't think we need (or want) to use String on the Rust side for numbers -- even in these objects that basically exist to define the JSON schema of the API. It seems like Serde has the flexibility to serialize/deserialize a u64 field to/from a String in JSON by using field attributes, https://serde.rs/field-attrs.html. It looks like people have already fought this battle, and there is code for doing what we want:

https://github.com/serde-rs/json/issues/412
https://github.com/vityafx/serde-aux/blob/master/src/field_attributes.rs#L108

It looks like there are a number of places (pardon the pun) in this file where integers are represented as Strings. I'm sure we could manually back and forth between integers and strings (as well as documenting what each String is supposed to encode, and unit testing the conversions, etc), but that sounds tedious and error-prone compared to using Serde's field attributes to declare that a u64 should be converted to/from a JSON String.

sugargoat: That's neat! We should probably use those field attrs in mobilecoind-json as well.

I created a ticket for this ([FS-90](https://mobilecoin.atlassian.net/browse/FS-90), as I think this big of a change is outside the scope of this PR (for consistency, all types should be made to use the field attrs, as well as the mobilecoind-json types we are importing).

I also created a ticket for re-implementing the json types from mobilecoind-json, rather than importing them. However, I do think we should discuss, as it is nice to have a consistent object API across our wallet implementations. [FS-91](https://mobilecoin.atlassian.net/browse/FS-91).

For the purposes of this PR (to unblock the electron wallet integration), I think we should move forward with this bandaid, and we can do full surgery later, possibly as part of the PR that is refactoring the json types. The refactor should not have any affect on downstream.

full-service/src/service/decorated_types.rs
sugargoat: Done!

eranrund (approved):
Thanks for getting rid of the unwraps. I agree the correct way is to use the serde attributes, but since @mfaulk is away and won't get to that this week and this is a relatively small PR, I think it can get merged. I do hope that this does get properly addressed in the near future.

PR #33, opened 2021-02-12
Add rustfmt.toml

mfaulk:
More consistent project formatting via rustfmt

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,4 +1,7 @@
 merge_imports = true
+edition = "2018"
+wrap_comments = true
+
 
 ignore = [
     "mobilecoin",   # The mobilecoin submodule is a dependency of this project.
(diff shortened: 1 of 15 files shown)


rustfmt.toml
```diff
@@ -0,0 +1,5 @@
+merge_imports = true
+
+ignore = [
+    "mobilecoin",   # The mobilecoin submodule is a dependency of this project.
```
mfaulk, line 4: I **think** this is the right thing to do?

eranrund: I think so.
I would suggest adding these two:
```
edition = "2018"
wrap_comments = true
```

@jcape has been using them in some places and I have created a ticket for myself to put that in the root `rustfmt.toml`

eranrund (approved):

sugargoat (approved):

PR #36, opened 2021-02-17
FS-92: JSON Errors

sugargoat:
### Motivation

The current error formatting is clunky and doesn't guarantee proper json. This PR fixes that.

### In this PR
* Introduces format_error helper at jsom-rpc layer
* Fixes up displaydoc for errors to add `:` before wrapped error.

[FS-92](https://mobilecoin.atlassian.net/browse/FS-92)

### DIFF
--- a/full-service/src/error.rs
+++ b/full-service/src/error.rs
@@ -7,13 +7,13 @@ use displaydoc::Display;
 #[derive(Display, Debug)]
 #[allow(clippy::large_enum_variant)]
 pub enum WalletServiceError {
-    /// Error interacting with the DB {0}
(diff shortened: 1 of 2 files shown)


full-service/src/error.rs
```diff
@@ -231,27 +231,39 @@ impl From<prost::DecodeError> for WalletDbError {
     }
 }
 
+impl<T> From<std::sync::PoisonError<T>> for WalletDbError {
```
sugargoat, line 234: Shouldn't be here - removing

voloshyn (approved):
LGTM

PR #40, opened 2021-02-18
FS-99: Address pay-self edge case and test

sugargoat:
### Motivation

The current Txo processing did not account well for the case where an account is sending transactions to itself. This PR fixes that case, and updates the txo lifecycle test to be more complete.

### In this PR
* Addresses pay-to-self edge case for Txo processing
* Updates `test_txo_lifecycle` to account for this case
* Adds a `list_txo_by_type` method in the Txo db crate. We may expose this in the future at the API level, but for now it's only used in the test.

[FS-99](https://mobilecoin.atlassian.net/browse/FS-99)

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -113,9 +113,9 @@ pub trait TransactionLogModel {
     /// txos involved in the transaction.
     ///
     /// Note: We expect transactions created with this wallet to have one
-    /// recipient, with the       rest of the minted txos designated as
-    /// change. Other wallets may choose to behave       differently, but
-    /// our TransactionLogs Table assumes this behavior.
+    /// recipient, with the rest of the minted txos designated as change. Other
+    /// wallets may choose to behave differently, but our TransactionLogs Table
+    /// assumes this behavior.
     fn log_submitted(
         tx_proposal: TxProposal,
         block_count: u64,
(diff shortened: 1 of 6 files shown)


full-service/src/db/transaction_log.rs
```diff
@@ -678,7 +681,7 @@ mod tests {
                 .subaddress_index,
             0
         );
-        assert!(input_details.spent_from_account.is_none());
+        assert!(input_details.secreted_from_account.is_none());
```
sugargoat, line 684: Updated this field name as "spent_from" is ambiguous.

full-service/src/db/txo.rs
```diff
@@ -55,7 +55,16 @@ pub struct TxoDetails {
     pub txo: Txo,
     pub received_to_account: Option<AccountTxoStatus>,
     pub received_to_assigned_subaddress: Option<AssignedSubaddress>,
-    pub spent_from_account: Option<AccountTxoStatus>,
+    pub secreted_from_account: Option<AccountTxoStatus>,
+}
+
+#[derive(Debug, Clone)]
+pub struct ProcessedTxProposalOutput {
```
sugargoat, line 62: Added to make return type less cumbersome.

full-service/src/db/txo.rs
```diff
@@ -364,12 +408,12 @@ impl TxoModel for Txo {
...
-        Ok((
-            outlay_receiver,
-            txo_id.to_string(),
-            total_output_value as i64,
-            transaction_txo_type.to_string(),
-        ))
+        Ok(ProcessedTxProposalOutput {
+            recipient: outlay_receiver,
+            txo_id: txo_id.to_string(),
+            value: log_value as i64,
```
sugargoat, line 414: Previously, this always returned the value of the overall transaction, but if we are calling this method (`create_minted`) on a change value, then the correct value to return is the value of the change.

full-service/src/db/txo.rs
```diff
@@ -561,36 +625,39 @@ impl TxoModel for Txo {
...
-                    };
-                    txo_details.received_to_assigned_subaddress = assigned_subaddress;
                 }
                 _ => {
                     return Err(WalletDbError::UnexpectedTransactionTxoType(
                         account_txo_status.txo_type,
                     ))
                 }
             }
+            // Get the subaddress details if assigned
```
sugargoat, line 648: Moved here because a minted & unspent transaction can have an assigned subaddress (this is when sent to self, or as change)

eranrund (approved):

PR #43, opened 2021-02-22
API v2

sugargoat:
### Motivation

This PR introduces an API v2 with several fixes for the consistency of methods and objects across the API, as well as adhering to the JSON RPC 2.0 spec.

The changes in this PR are carefully implemented to be backward compatible with the current API, so that both are supported simultaneously by the service.

This PR builds on the work by @mfaulk in https://github.com/mobilecoinofficial/full-service/pull/24.


full-service/src/service/api_v2.rs
```diff
@@ -1 +1,131 @@
 // Copyright (c) 2020-2021 MobileCoin Inc.
+
+//! The Wallet API for Full Service. Version 1.
```
mfaulk: "Version 2"

full-service/src/service/wallet.rs
```diff
@@ -36,6 +117,6 @@ pub fn rocket(
     state: WalletState<ThickClient<HardcodedCredentialsProvider>, FogResolver>,
 ) -> rocket::Rocket {
     rocket::custom(rocket_config)
-        .mount("/", routes![wallet_api, wallet_help])
+        .mount("/", routes![wallet_api, wallet_help_v2, wallet_help_v1])
```
mfaulk: I'm not sure what the convention is for JSON-RPC, but I have often seen versioned REST APIs use a URL scheme like /api/v1/foo, /api/v2/foo. I think that more clearly conveys the hierarchy than having things like /wallet --> v2, and /wallet/v1 --> v1.

Maybe something like this would be nice:

```
.mount("/api/v1", routes![wallet_api_v1, wallet_help_v1])
.mount("/api/v2", routes![wallet_api_v2, wallet_help_v2])
```

full-service/src/json_rpc/account_key.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use serde_derive::{Deserialize, Serialize};
+
+/// The AccountKey contains a View keypair and a Spend keypair, used to
+/// construct and receive transactions.
+#[derive(Deserialize, Serialize, Default, Debug, Clone)]
+pub struct AccountKey {
+    /// String representing the object's type. Objects of the same type share
+    /// the same value.
+    pub object: String,
```
mfaulk, line 14: It definitely feels weird using a String field as a type hint, but I guess that's the state of type systems in JSON :(

mfaulk:
This looks like a lot of good stuff! Having a versioned API is great.

Is it really necessary to support the old (v1) API? I'm still thinking of full-service as being in the "pre-release / breaking changes are OK phase", and it would be nice to not have to support multiple APIs at this time.

full-service/src/json_rpc/wallet.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct WalletState<
+    T: BlockchainConnection + UserTxConnection + 'static,
+    FPR: FogPubkeyResolver + Send + Sync + 'static,
+> {
+    /// The Wallet Service implementation.
+    pub service: WalletService<T, FPR>,
+}
+
+/// The route for the Full Service Wallet API.
+#[post("/wallet", format = "json", data = "<command>")]
```
mfaulk, line 37: It feels weird to me that this handles requests to both (and eventually, all?) API versions. It seems like that adds boilerplate (i.e. the if / else blocks, an api_version field to the Requests, etc.), but mostly it seems to couple together the code paths of each API version, when ideally those are as independent as possible.

My vote would be for a top-level module separation (something like json_rpc/v1 and json_rpc/v2), with routes mapped under the top-level URLs /v1/foo and /v2/foo, etc.

full-service/src/db/assigned_subaddress.rs
```diff
@@ -211,6 +217,19 @@ impl AssignedSubaddressModel for AssignedSubaddress {
 
         Ok(matches)
     }
+
+    fn delete_all(
+        account_id_hex: &str,
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<(), WalletDbError> {
+        use crate::db::schema::assigned_subaddresses::dsl::{
+            account_id_hex, assigned_subaddresses,
```
christian-oudard, line 226:
```suggestion
            account_id_hex as schema_account_id_hex, assigned_subaddresses,
```

full-service/src/db/assigned_subaddress.rs
```diff
@@ -211,6 +217,19 @@ impl AssignedSubaddressModel for AssignedSubaddress {
...
+        };
+
+        diesel::delete(assigned_subaddresses.filter(account_id_hex.eq(account_id_hex)))
```
christian-oudard, line 229:
```suggestion
        diesel::delete(assigned_subaddresses.filter(schema_account_id_hex.eq(account_id_hex)))
```

full-service/src/db/assigned_subaddress.rs
christian-oudard: currently masking the argument, causing the function to never work.

full-service/src/service/wallet.rs
sugargoat: We discussed offline and decided we will not be maintaining v1 in a backward-compatible way. In the future though, this versioning scheme may be what we decide to do.

christian-oudard (approved):

PR #44, opened 2021-02-23
FS-108: Fix unique account_txo_status violation on sync

sugargoat:
### Motivation

When sending transactions between accounts in the same wallet (From Alice -> Bob, then from Bob -> Alice), the handling of the Txo was not quite right as it moved from `(minted, secreted)` in Alice's account, to `(received, unspent)` and then to `(received, spent)` in Bob's account. This PR fixes the issues and adds a test at the service layer to verify.

### In this PR
* Fixes Txo management when transactions are sent between accounts in the same wallet.

[FS-108](https://mobilecoin.atlassian.net/browse/FS-108)

### DIFF
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -100,7 +100,7 @@ pub trait AccountModel {
     fn get_by_txo_id(
         txo_id_hex: &str,
         conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
-    ) -> Result<Account, WalletDbError>;
+    ) -> Result<Vec<Account>, WalletDbError>;
 
     /// Update an account.
     /// The only updatable field is the name. Any other desired update requires
@@ -276,7 +276,7 @@ impl AccountModel for Account {
     fn get_by_txo_id(
         txo_id_hex: &str,
         conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
-    ) -> Result<Account, WalletDbError> {
+    ) -> Result<Vec<Account>, WalletDbError> {
         use crate::db::schema::account_txo_statuses::dsl::account_txo_statuses;
 
         match account_txo_statuses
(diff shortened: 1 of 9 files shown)


full-service/src/db/account.rs
```diff
@@ -276,23 +276,18 @@ impl AccountModel for Account {
...
-            Ok(a) => {
-                if a.len() > 1 {
-                    return Err(WalletDbError::MultipleStatusesForTxo);
-                }
-                Ok(Account::get(
-                    &AccountID(a[0].account_id_hex.to_string()),
-                    conn,
-                )?)
-            }
+            Ok(accounts) => Ok(accounts
```
sugargoat, line 287: When we have multiple accounts in the same wallet, then when getting a Txo, we need to return all accounts which may be associated with that Txo.

full-service/src/db/txo.rs
```diff
@@ -268,6 +269,13 @@ impl TxoModel for Txo {
                                     TXO_RECEIVED,
                                     conn,
                                 )?;
+                                txo_details.txo.update_received(
```
sugargoat, line 272: Make sure the subaddress gets added on received.

full-service/src/db/txo.rs
```diff
@@ -519,14 +527,30 @@ impl TxoModel for Txo {
...
-            let account = Account::get_by_txo_id(&txo_id.to_string(), conn)?;
+            let accounts = Account::get_by_txo_id(&txo_id.to_string(), conn)?;
 
             // Update the status to pending.
-            diesel::update(
-                account_txo_statuses.find((&account.account_id_hex, &txo_id.to_string())),
-            )
-            .set(crate::db::schema::account_txo_statuses::txo_status.eq(TXO_PENDING.to_string()))
-            .execute(conn)?;
+            if accounts.len() > 2 {
```
sugargoat, line 533: Handle the case where a Txo "belongs" to two accounts in the wallet (it was minted by one account and received by another).

full-service/src/db/txo.rs
```diff
@@ -638,26 +662,25 @@ impl TxoModel for Txo {
                 }
                 TXO_RECEIVED => {
                     txo_details.received_to_account = Some(account_txo_status.clone());
+                    // Get the subaddress details if assigned
```
sugargoat, line 665: Moved back under TXO_RECEIVED because the subaddress should only be added from the perspective of a Txo getting received.

christian-oudard (approved):
Looks good, tested it out, and the crash no longer happens.

PR #48, opened 2021-02-25
Adds doc comments for database models and functions.

mfaulk:
### Motivation

The database layer is complex and would benefit from documentation.

### In this PR
* Adds doc comments to database entities.
* Adds doc comments to database access functions.

This is incremental progress, but since it's only comments I think we should merge early-and-often.

### DIFF
--- a/full-service/src/db/models.rs
+++ b/full-service/src/db/models.rs
@@ -45,7 +45,7 @@ pub const TXO_CHANGE: &str = "change";
 pub struct Account {
     /// Primary key
     pub id: i32,
-    /// ???
+    /// An additional ID, derived from the account data.
     pub account_id_hex: String,
     /// Private keys for viewing and spending the MobileCoin belonging to an
     /// account.
@@ -59,7 +59,7 @@ pub struct Account {
     pub next_subaddress_index: i64,
     /// Index of the first block where this account may have held funds.
     pub first_block: i64,
-    /// TODO: ???
+    /// Index of the next block to inspect for transactions related to this account.
     pub next_block: i64,
     /// If the account was imported, account history prior to this block index
     /// is derived from the public ledger, and does not reflect client-side
@@ -92,7 +92,7 @@ pub struct NewAccount<'a> {
 pub struct Txo {
     /// Primary key
     pub id: i32,
-    /// ???
+    /// An additional ID derived from the contents of the ledger TxOut.
     pub txo_id_hex: String,
     /// The value of this transaction output, in picoMob.
     pub value: i64,
@@ -106,8 +106,7 @@ pub struct Txo {
     pub txo: Vec<u8>,
     /// The receiving subaddress, if known.
     pub subaddress_index: Option<i64>,
-    /// TODO: why is this optional? Denoting if the Txo has subsequently been
-    /// spent?
+    /// Pre-computed key image for this Txo, or None if the Txo is orphaned.
     pub key_image: Option<Vec<u8>>,
     /// Block index containing this Txo.
     pub received_block_count: Option<i64>,
--- a/full-service/src/db/models.rs
+++ b/full-service/src/db/models.rs
@@ -37,23 +37,40 @@ pub const TXO_INPUT: &str = "input";
 pub const TXO_OUTPUT: &str = "output";
 pub const TXO_CHANGE: &str = "change";
 
+/// An Account entity.
+///
+/// Contains the account private keys, subaddress configuration, and ...
 #[derive(Clone, Serialize, Identifiable, Queryable, PartialEq, Debug)]
 #[primary_key(id)]
 pub struct Account {
+    /// Primary key
     pub id: i32,
+    /// ???
     pub account_id_hex: String,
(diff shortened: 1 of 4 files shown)


full-service/src/db/models.rs
```diff
@@ -37,23 +37,40 @@ pub const TXO_INPUT: &str = "input";
...
 
+/// An Account entity.
+///
+/// Contains the account private keys, subaddress configuration, and ...
 #[derive(Clone, Serialize, Identifiable, Queryable, PartialEq, Debug)]
 #[primary_key(id)]
 pub struct Account {
+    /// Primary key
     pub id: i32,
+    /// ???
```
mfaulk, line 48: Is there a short explanation about why an account has both an 'id' (the DB table primary key) and an 'account_id_hex' (which appears to be the hex-encoded digest of the Account object)?

full-service/src/db/models.rs
```diff
@@ -37,23 +37,40 @@ pub const TXO_INPUT: &str = "input";
...
+    /// Default subadress that is given out to refer to this account.
     pub main_subaddress_index: i64,
+    /// Subaddress used to return transaction "change" to self.
     pub change_subaddress_index: i64,
+    /// The next unused subaddress index. (Assumes indices are used sequentially
+    /// from 0).
     pub next_subaddress_index: i64,
+    /// Index of the first block where this account may have held funds.
     pub first_block: i64,
+    /// TODO: ???
```
mfaulk, line 62: Not sure what this is...

full-service/src/db/models.rs
```diff
@@ -68,24 +85,38 @@ pub struct NewAccount<'a> {
...
 }
 
+/// A received transaction output entity that belongs to a an Account in this
+/// wallet. Also maybe a transaction output that hasn't been sent yet?
 #[derive(Clone, Serialize, Identifiable, Queryable, PartialEq, Debug)]
 #[primary_key(id)]
 pub struct Txo {
+    /// Primary key
     pub id: i32,
+    /// ???
```
mfaulk, line 95: Again, what's the story with this and the primary key 'id'?

full-service/src/db/models.rs
sugargoat: The short explanation is wanting to do operations based on a recoverable identifier from the account_key (the account_id_hex is hex-encoded hash of the account's public key), as opposed to doing lookups based on the account_key bytes blob (this would be in a context where we didn't have the primary key). It may be worth reviewing whether we have many contexts where you need to do that lookup...

full-service/src/db/models.rs
sugargoat, line 62:
```suggestion
    /// The next block to be scanned for syncing the account.
```

sugargoat: This is used by the account sync to pick up scanning at the next un-scanned block.

full-service/src/db/models.rs
sugargoat: This one might have more merit as a separate field than the account_id_hex - the reason for this is that you may get a Txo in different contexts, and you want to look up if you already have it in your DB, so you use the reproducible txo_id_hex, derived from the TxOut itself.

full-service/src/db/txo.rs
```diff
@@ -66,19 +67,24 @@ pub struct ProcessedTxProposalOutput {
 }
 
 pub trait TxoModel {
-    /// Create a received Txo.
+    /// Create (upsert?) a received Txo.
```
sugargoat, line 70: Upsert is correct - feel free to remove the ?

full-service/src/db/models.rs
mfaulk: Cool. My suspicion is that we can use uniqueness constraints somewhere (e.g. on the private key bytes) to make sure each account is never added to the table twice. In that case, an auto-incrementing primary key will probably work fine.

full-service/src/db/models.rs
```diff
@@ -68,24 +85,38 @@ pub struct NewAccount<'a> {
...
     pub target_key: Vec<u8>,
+    /// The serialized public_key of the TxOut.
     pub public_key: Vec<u8>,
+    /// The serialized e_fog_hint of the TxOut.
     pub e_fog_hint: Vec<u8>,
+    /// The serialized TxOut.
     pub txo: Vec<u8>,
+    /// The receiving subaddress, if known.
     pub subaddress_index: Option<i64>,
+    /// TODO: why is this optional? Denoting if the Txo has subsequently been
```
sugargoat, line 109: We actually precompute the KeyImage currently, so an unspent Txo could still have a `Some(key_image)`. I forget the reason for this...I recall I wanted to eliminate it, since at first I thought it was a security risk to have the precomputed key_images, but I think there's some reason it's necessary to store....which we should definitely document...

The reason it's optional is the same as the subaddress - you can only have the key_image if you also had the subaddress. So, if this field is None, then it's an "orphaned" txo.

sugargoat (approved):
Looks great! One unanswered question remains, which is why do we need to precompute and store the key_images? (Though that is not blocking this PR - just remains for us to evaluate in the larger context of the DB)

PR #49, opened 2021-02-25
TXO status constants

mfaulk:
Txos in the database have both a "status" and a "type", and both status and type have associated constants. However, the constants don't contain the word "status" or "type" in their names or values, which causes ambiguity and invites hard-to-see bugs.

### In this PR
* Gives the DB-level TXO status constants more descriptive names and unambiguous values.
* Fixes unit tests that referred to the DB-level constants, instead of the API-level JSON response fields.

This is a breaking database change, which I'm assuming is OK at this point?

### DIFF
--- a/full-service/src/db/models.rs
+++ b/full-service/src/db/models.rs
@@ -9,14 +9,21 @@ use super::schema::{
 
 use serde::Serialize;
 
-// FIXME: WS-13 - Would be great to get enums to work. Run into several issues
-// when attempting        to use https://github.com/adwhit/diesel-derive-enum for sqlite
-// TxoStatus
-pub const TXO_UNSPENT: &str = "unspent";
-pub const TXO_PENDING: &str = "pending";
-pub const TXO_SPENT: &str = "spent";
-pub const TXO_SECRETED: &str = "secreted";
-pub const TXO_ORPHANED: &str = "orphaned";
+/// A TXO owned by this wallet that has not yet been spent.
+pub const TXO_STATUS_UNSPENT: &str = "txo_status_unspent";
+
+/// A TXO owned by this wallet that is used by a pending transaction.
+pub const TXO_STATUS_PENDING: &str = "txo_status_pending";
+
+/// A TXO owned by this wallet that has been spent.
+pub const TXO_STATUS_SPENT: &str = "txo_status_spent";
+
+/// TODO: ???
+pub const TXO_STATUS_SECRETED: &str = "txo_status_secreted";
+
+/// The TXO is owned by this wallet, but not yet spendable (i.e., receiving
+/// subaddress is unknown).
+pub const TXO_STATUS_ORPHANED: &str = "txo_status_orphaned";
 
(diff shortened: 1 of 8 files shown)


full-service/src/db/models.rs
```diff
@@ -9,14 +9,21 @@ use super::schema::{
...
+/// A TXO owned by this wallet that has not yet been spent.
+pub const TXO_STATUS_UNSPENT: &str = "txo_status_unspent";
+
+/// A TXO owned by this wallet that is used by a pending transaction.
+pub const TXO_STATUS_PENDING: &str = "txo_status_pending";
+
+/// A TXO owned by this wallet that has been spent.
+pub const TXO_STATUS_SPENT: &str = "txo_status_spent";
+
+/// TODO: ???
```
mfaulk, line 21: @sugargoat Can you give a short definition for this?

full-service/src/service/wallet.rs
```diff
@@ -941,7 +939,7 @@ mod tests {
         });
         let result = dispatch(&client, body, &logger);
         let balance_status = result.get("status").unwrap();
-        let unspent = balance_status.get(TXO_UNSPENT).unwrap().as_str().unwrap();
+        let unspent = balance_status.get("unspent").unwrap().as_str().unwrap();
```
mfaulk, line 942: Here and in other places, the intent was to get a field from the JSON response object, but the code incorrectly used the database constants.

mfaulk: (In general, it would probably be easier to avoid get("foo") altogether, and instead use serde to convert the response JSON into the response struct, and assert that the deserialized struct equals the expected response struct.)

full-service/src/db/models.rs
sugargoat:
```suggestion
/// A TXO created by an account in this wallet to send as an output in an outgoing transaction.
```

full-service/src/db/models.rs
```diff
@@ -9,14 +9,21 @@ use super::schema::{
...
 
-// FIXME: WS-13 - Would be great to get enums to work. Run into several issues
-// when attempting        to use https://github.com/adwhit/diesel-derive-enum for sqlite
-// TxoStatus
-pub const TXO_UNSPENT: &str = "unspent";
-pub const TXO_PENDING: &str = "pending";
-pub const TXO_SPENT: &str = "spent";
-pub const TXO_SECRETED: &str = "secreted";
-pub const TXO_ORPHANED: &str = "orphaned";
+/// A TXO owned by this wallet that has not yet been spent.
```
sugargoat, line 12: Would maybe specify "by an account in this wallet" to be super clear that these statuses are relationships to a specific account, not the whole wallet.

full-service/src/service/wallet.rs
sugargoat: Sounds great, ty!

sugargoat (approved):
LGTM - a minor doc change request

PR #50, opened 2021-02-26
Txo type constants

mfaulk:
### In this PR
* Changes DB-level constants for Txo type to include the word "type" in their name and value.


sugargoat (approved):
LGTM!

PR #51, opened 2021-02-26
Updates TX-related constants

mfaulk:
### In this PR
* Continues to update DB constants to have doc comments, more descriptive names, and unique string values.

### DIFF
--- a/full-service/src/db/models.rs
+++ b/full-service/src/db/models.rs
@@ -22,20 +22,33 @@ pub const TXO_ORPHANED: &str = "orphaned";
 pub const TXO_MINTED: &str = "minted";
 pub const TXO_RECEIVED: &str = "received";
 
-// TransactionStatus
-pub const TX_BUILT: &str = "built";
-pub const TX_PENDING: &str = "pending";
-pub const TX_SUCCEEDED: &str = "succeeded";
-pub const TX_FAILED: &str = "failed";
+/// A transaction output used as an input to a new transaction.
+pub const TXO_USED_AS_INPUT: &str = "txo_used_as_input";
 
-// Transaction Direction
-pub const TX_DIR_SENT: &str = "sent";
-pub const TX_DIR_RECEIVED: &str = "received";
+/// A transaction output used as an output of a new transaction.
+pub const TXO_USED_AS_OUTPUT: &str = "txo_used_as_output";
 
-// Transaction Txo Type
-pub const TXO_INPUT: &str = "input";
-pub const TXO_OUTPUT: &str = "output";
-pub const TXO_CHANGE: &str = "change";
+/// A transaction output used as a change output of a new transaction.
+pub const TXO_USED_AS_CHANGE: &str = "txo_used_as_change";
+
+/// A transaction that has been built locally.
+pub const TX_STATUS_BUILT: &str = "tx_status_built";
+
+/// A transaction that has been submitted to the MobileCoin network.
+pub const TX_STATUS_PENDING: &str = "tx_status_pending";
+
+/// A transaction that appears to have been processed by the MobileCoin network.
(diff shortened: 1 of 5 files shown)


full-service/src/db/models.rs
```diff
@@ -22,20 +22,33 @@ pub const TXO_ORPHANED: &str = "orphaned";
...
+pub const TX_STATUS_SUCCEEDED: &str = "tx_status_succeeded";
+
+/// A transaction that was rejected by the MobileCoin network, or that expired
+/// before it could be processed.
+pub const TX_STATUS_FAILED: &str = "tx_status_failed";
+
+/// A transaction created by this wallet.
+pub const TX_DIRECTION_SENT: &str = "tx_direction_sent";
+
+/// ???
```
mfaulk, line 50: What's the story for a received Tx? That seems at odds with the MobileCoin ledger, which does not record transactions directly.

sugargoat: Yeah - this one is tricky. The current behavior is that every received `TxOut` generates a `TransactionLog` entry with `TX_DIRECTION_RECEIVED`, even if they were part of the same transaction - this is because the MobileCoin ledger cannot actually determine which `TxOut`s were grouped together in a "transaction."

The desired API is that you could query a block and get all the transactions in that block that you know something about, with their `direction` specified.

full-service/src/db/models.rs
```diff
@@ -22,20 +22,33 @@ pub const TXO_ORPHANED: &str = "orphaned";
...
+pub const TX_STATUS_PENDING: &str = "tx_status_pending";
+
+/// A transaction that appears to have been processed by the MobileCoin network.
+pub const TX_STATUS_SUCCEEDED: &str = "tx_status_succeeded";
+
+/// A transaction that was rejected by the MobileCoin network, or that expired
+/// before it could be processed.
+pub const TX_STATUS_FAILED: &str = "tx_status_failed";
+
+/// A transaction created by this wallet.
```
sugargoat, line 47:
```suggestion
/// A transaction created by an account in this wallet.
```

full-service/src/db/models.rs
```diff
@@ -22,20 +22,33 @@ pub const TXO_ORPHANED: &str = "orphaned";
...
+/// A transaction output used as a change output of a new transaction.
+pub const TXO_USED_AS_CHANGE: &str = "txo_used_as_change";
+
+/// A transaction that has been built locally.
+pub const TX_STATUS_BUILT: &str = "tx_status_built";
+
+/// A transaction that has been submitted to the MobileCoin network.
+pub const TX_STATUS_PENDING: &str = "tx_status_pending";
+
+/// A transaction that appears to have been processed by the MobileCoin network.
```
sugargoat, line 40: Would maybe drop "appears to have been," unless there is some specific uncertainty you're capturing? In which case, maybe we could specify how it could seem to be processed if it wasn't actually.

sugargoat (approved):
LGTM pending whatever docstring you think is appropriate for `TX_DIRECTION_RECEIVED`

full-service/src/db/models.rs
mfaulk: Maybe I'm splitting hairs here, but transactions do not map one-to-one onto the ledger data. For example, you could create two transactions that consume the same inputs and create the same outputs, but use different mixins. These two transactions would have the same effect on the ledger, and a client would be unable to tell which of the two transactions was processed.

Not sure what to do here. Seems like we should either add some kind of "processed transaction ID" to the ledger so that it is clear which transactions were processed in a block, or remove the client-side concept of "this transaction was processed".

sugargoat: Got it - I like the idea of a processed_transaction_id.

PR #54, opened 2021-02-26
[WIP] Adding documentation for account syncing

mfaulk:
### Motivation

The functions used to scan ledger blocks against the wallet accounts should have unit tests and better documentation.

### In this PR
* Adds doc comments to `process_txos`

### DIFF
--- a/full-service/src/error.rs
+++ b/full-service/src/error.rs
@@ -122,66 +122,6 @@ impl From<diesel::result::Error> for WalletServiceError {
 }
 
 #[derive(Display, Debug)]
-pub enum SyncError {
-    /// Could not find account
-    AccountNotFound,
-
-    /// Error with WalletDb: {0}
-    Database(WalletDbError),
-
-    /// Error with LedgerDB: {0}
-    LedgerDB(mc_ledger_db::Error),
-
-    /// Error with Keys: {0}
-    CryptoKey(mc_crypto_keys::KeyError),
-
-    /// Error decoding prost: {0}
-    ProstDecode(prost::DecodeError),
-
-    /// Error with the Amount: {0}
-    Amount(mc_transaction_core::AmountError),
-
-    /// Error executing diesel transaction: {0}
-    Diesel(diesel::result::Error),
-}
-
-impl From<WalletDbError> for SyncError {
-    fn from(src: WalletDbError) -> Self {
-        Self::Database(src)
-    }
-}
-
-impl From<mc_ledger_db::Error> for SyncError {
-    fn from(src: mc_ledger_db::Error) -> Self {
-        Self::LedgerDB(src)
-    }
-}
-
-impl From<mc_crypto_keys::KeyError> for SyncError {
-    fn from(src: mc_crypto_keys::KeyError) -> Self {
-        Self::CryptoKey(src)
-    }
-}
-
-impl From<prost::DecodeError> for SyncError {
-    fn from(src: prost::DecodeError) -> Self {
-        Self::ProstDecode(src)
-    }
-}
-
-impl From<mc_transaction_core::AmountError> for SyncError {
-    fn from(src: mc_transaction_core::AmountError) -> Self {
-        Self::Amount(src)
-    }
-}
-
-impl From<diesel::result::Error> for SyncError {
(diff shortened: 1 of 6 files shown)


full-service/src/error.rs
```diff
@@ -121,66 +121,6 @@ impl From<diesel::result::Error> for WalletServiceError {
     }
 }
 
-#[derive(Display, Debug)]
-pub enum SyncError {
```
mfaulk, line 125: I moved this to a new file in the service layer.

full-service/src/service/sync.rs
```diff
@@ -63,24 +55,27 @@ use std::{
...
+/// account.
+const BLOCKS_PER_ACCOUNT_UPDATE: u64 = 5;
 
 /// The AccountId corresponds to the Account's primary key: account_key_hex.
 pub type AccountId = String;
 
-/// Message type our crossbeam channel uses to communicate with the worker
-/// thread pull.
+/// Commands for the Sync worker.
 enum SyncMsg {
```
mfaulk, line 66: I'd like to move these towards more idempotent units of work. For example "Sync account X against blocks 100..200" is much easier to reason about than "Sync account X against whatever blocks happen to be in the ledger when you get this message".

full-service/src/service/sync.rs
```diff
@@ -91,44 +86,51 @@ pub struct SyncThread {
...
     ) -> Self {
-        // Queue for sending jobs to our worker threads.
+        // Worker task queue.
         let (sender, receiver) = crossbeam_channel::unbounded::<SyncMsg>();
 
-        // A hashset to keep track of which AccountIds were already sent to the queue,
-        // preventing them from being sent again until they are processed.
+        // AccountIds that have been added to the queue, but have not yet been
+        // processed.
         let queued_account_ids = Arc::new(Mutex::new(HashSet::<AccountId>::default()));
```
mfaulk, line 112: Gut feel: this stuff would be easier if the "task master" periodically reads the accounts and the ledger state, and then dispatches finite tasks like "Apply account X to blocks Y..Z", instead of letting the worker thread figure out what work needs to be done.

full-service/src/service/sync.rs
```diff
@@ -311,100 +305,101 @@ fn sync_thread_entry_point(
...
+            return Ok(AccountSyncStatus::Synced);
         }
     }
-    Ok(SyncAccountOk::MoreBlocksPotentiallyAvailable)
+    Ok(AccountSyncStatus::MoreBlocksPotentiallyAvailable)
 }
 
-/// Helper function for matching a list of TxOuts to a given account.
+
+/// TODO: What does this do?
```
mfaulk, line 366: Is there a short description of what the intention here is? I'm having a hard time describing this function accurately in one sentence.

full-service/src/service/sync.rs
```diff
@@ -311,100 +311,96 @@ fn sync_thread_entry_point(
...
-    for _ in 0..MAX_BLOCKS_PROCESSING_CHUNK_SIZE {
-        let conn = wallet_db.get_conn()?;
-        let sync_status = conn.transaction::<SyncAccountOk, SyncError, _>(|| {
-            // Get the account data. If it is no longer available, the account has been
-            // removed and we can simply return.
+) -> Result<AccountSyncStatus, SyncError> {
+    // Index of the last block currently in the ledger.
+    let final_block_index = ledger_db.num_blocks()? - 1;
+
+    let conn = wallet_db.get_conn()?;
```
mfaulk, line 324: Is there a reason that `conn` was opened inside the for loop? I **think** it's safe to get the connection once, and then create a separate DB transaction on each iteration of the loop...

eranrund: That depends on the pool size - if the size is small (e.g. 1 - https://github.com/mobilecoinofficial/full-service/blob/main/full-service/src/db/wallet_db.rs#L22) that means that until the loop finished executing, nothing else gets to get a connection.

mfaulk: Thanks. I will leave it in the loop

BrianCorbin:
Closing because stale

PR #58, opened 2021-02-28
FS-64: Export Account Secrets

sugargoat:
### Motivation

Users need a way to be able to export their account secrets. There should also be a way to import an account from an account key.

### In this PR
* adds account key secret export
* adds import by account key
* moves entropy to be a nullable field for the Account object (in case you import an account key)
* adds tests
* updates documentation
* removes tests and endpoints from api_v1 that are now deprecated.
* I think this also fixes the flaky test

[FS-64](https://mobilecoin.atlassian.net/browse/FS-64)
[FS-57](https://mobilecoin.atlassian.net/browse/FS-57)
[FS-125](https://mobilecoin.atlassian.net/browse/FS-125)

### DIFF
--- a/full-service/src/service/balance.rs
+++ b/full-service/src/service/balance.rs
@@ -25,13 +25,6 @@ use diesel::{
     Connection,
 };
 
-/*
-use displaydoc::Display;
-/// Errors for the Balance Service.
-#[derive(Display, Debug)]
-pub enum BalanceServiceError {}
-*/
-
 /// The balance object returned by balance services.
 ///
 /// This must be a service object because there is no "Balance" table in our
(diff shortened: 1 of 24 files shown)


sugargoat:
After discussing with @christian-oudard - will remove the entropy and account_key from the Account API object

full-service/src/service/balance.rs
```diff
@@ -175,6 +156,7 @@ where
                     min_synced_block_index,
                     (account.next_block as u64).saturating_sub(1),
                 );
+                println!("\x1b[1;33m For account {:?} got network block index = {:?}, local_ledger_block_count {:?}, and account.next_block {:?}", account_id, network_block_index, self.ledger_db.num_blocks().unwrap(), account.next_block);
```
christian-oudard, line 159:
```suggestion
                
```

christian-oudard (approved):
Discussed with Sara, approving.

PR #60, opened 2021-03-02
Moves account get_decorated to the API layer

mfaulk:
### Motivation

Continues the theme of encapsulating JSON-related code in the API layer.

### In this PR
* Moves Account::get_decorated out of account.rs and into wallet_api.rs,
* Adds Account::get_by_id (Seemed like a good idea to support querying by primary key),

### DIFF
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -116,14 +116,6 @@ pub trait AccountModel {
         conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
     ) -> Result<Account, WalletDbError>;
 
-    // /// Get the API-decorated Account object.
-    // fn get_decorated(
-    //     account_id_hex: &AccountID,
-    //     local_height: u64,
-    //     network_height: u64,
-    //     conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
-    // ) -> Result<JsonAccount, WalletDbError>;
-
     /// Get all accounts somehow related to a particular TXO?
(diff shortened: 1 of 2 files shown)


full-service/src/db/account.rs
```diff
@@ -231,6 +265,22 @@ impl AccountModel for Account {
...
     }
 
+    /// Get an account.
+    ///
+    /// # Arguments
+    /// * `id` - The account's ID.
+    ///
+    /// # Returns
+    /// * Account
+    fn get_by_id(
```
mfaulk, line 270: I tossed this in for good measure.

full-service/src/db/account.rs
```diff
@@ -79,23 +92,31 @@ pub trait AccountModel {
...
+    ///
+    /// # Returns
     /// * Account
     fn get(
         account_id_hex: &AccountID,
         conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
     ) -> Result<Account, WalletDbError>;
 
-    /// Get the API-decorated Account object
-    fn get_decorated(
```
mfaulk, line 92: Removing this is the main goal of the PR.

full-service/src/json_rpc/api_v1/wallet_api.rs
```diff
@@ -585,6 +593,43 @@ where
     Ok(Json(result))
 }
 
+fn get_json_account(
```
mfaulk, line 596: This isn't new, it just moved.

PR #64, opened 2021-03-04
Make Account entropy non nullable.

christian-oudard:
Also standardize some names, internally and on the API.
by_account -> for_account


eranrund:
Is this taking the stance that full-service only supports adding accounts when starting from entropy bytes?

sugargoat (approved):
Looks great!

sugargoat:
@eranrund - Yeah, we discussed and decided it's a bug in the MobileCoin ecosystem overall if there is a way to export an account key without entropy bytes.

full-service/src/db/account.rs
```diff
@@ -523,114 +497,10 @@ mod tests {
             account_id_hex
         };
         let account = Account::get(&account_id, &wallet_db.get_conn().unwrap()).unwrap();
-        let decoded_entropy = RootEntropy::try_from(account.entropy.unwrap().as_slice()).unwrap();
+        // let decoded_entropy = RootEntropy::try_from(&account.entropy).unwrap();
```
mfaulk, line 500: Leftover?

mfaulk (approved):

PR #66, opened 2021-03-04
Standardize naming to block_index, not _height or _count.

christian-oudard:
- Rename block_count and block_height to block_index.


full-service/src/scratch.txt
```diff
@@ -1,28 +0,0 @@
- GOT RESULTS [
```
sugargoat, line 1: Thanks for cleaning this up!

sugargoat (approved):
LGTM!

PR #67, opened 2021-03-04
Remove account secrets from API Account object.

christian-oudard:
Remove account secrets from API Account object.
Add `first_block_index` to the `get_account` endpoint.

### Motivation

We should only have one place to export account secrets, which is the export_account_secrets endpoint.


sugargoat (approved):

sugargoat (approved):
LGTM with rolled-up changes

PR #72, opened 2021-03-05
FS-159: Refactor proofs

sugargoat:
Soundtrack of this PR: [Baby Shark](https://www.youtube.com/watch?v=j8z7UjET1Is)

### Motivation

Continuing the refactor of services and API.

### In this PR
* Refactors the proof service

[FS-159](https://mobilecoin.atlassian.net/browse/FS-159)

### DIFF
--- a/full-service/src/error.rs
+++ b/full-service/src/error.rs
@@ -5,8 +5,9 @@
 use crate::{
     db::WalletDbError,
     service::{
-        proof::ProofServiceError, transaction::TransactionServiceError,
-        transaction_log::TransactionLogServiceError, txo::TxoServiceError,
+        balance::BalanceServiceError, ledger::LedgerServiceError, proof::ProofServiceError,
+        transaction::TransactionServiceError, transaction_log::TransactionLogServiceError,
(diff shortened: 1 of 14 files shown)


full-service/src/error.rs
```diff
@@ -67,6 +76,18 @@ impl From<TxoServiceError> for WalletServiceError {
     }
 }
 
+impl From<ProofServiceError> for WalletServiceError {
+    fn from(src: ProofServiceError) -> Self {
+        Self::ProofService(src)
+    }
+}
+
+impl From<TransactionLogServiceError> for WalletServiceError {
```
christian-oudard, line 85: Is there a way to reduce all of this duplication of `impl From`? It seems like every error class needs this line. Macros?

christian-oudard (approved):
Nice to learn all about the shared secret and proof system. Changes look good, glad so much stuff is moving to the new API system.

full-service/src/error.rs
sugargoat: Great question - I would love to figure out how to reduce the error conversion boilerplate

PR #74, opened 2021-03-08
FS-95: Receipts

sugargoat:
### Motivation

A primitive in the MobileCoin system is the "Receiver Receipt," where a recipient can be handed a "Receiver Receipt" from the sender, and they can use the data in this receipt to poll for the Txos in the transaction, and the verify the proof that the Txo was associated with the Receipt, implying that whoever is providing the receipt constructed the transaction.

### In this PR
* Adds a ReceiptService
* Implements `check_receiver_receipt_status` and `create_receiver_receipt`
* Implements `Txo::select_by_public_key` to select the Txos from the receipt
* Adds tests at each layer

[FS-95](https://mobilecoin.atlassian.net/browse/FS-95)

### Future Work
* Explore other convenience endpoints based on feedback from merchant services prototypes

### DIFF
--- a/full-service/migrations/2021-03-07-192850_receipts/up.sql
+++ b/full-service/migrations/2021-03-07-192850_receipts/up.sql
@@ -1,0 +1,1 @@
+CREATE UNIQUE INDEX idx_txos__txo_public_key ON txos (public_key);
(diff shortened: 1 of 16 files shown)


full-service/migrations/2021-03-07-192850_receipts/up.sql
```diff
@@ -0,0 +1 @@
+CREATE UNIQUE INDEX idx_txos__txo_public_key ON txos (public_key);
```
christian-oudard, line 1: Yes, this is important.

christian-oudard (approved):

PR #84, opened 2021-03-18
Fix FS-163, db crash after removing an account when it still has outlays in the system.

christian-oudard:
https://mobilecoin.atlassian.net/browse/FS-163


sugargoat (approved):
Nice!

PR #87, opened 2021-03-20
FS-174: Improve Receipt API

sugargoat:
### Motivation

The `check_receipt_status` endpoint was not quite right to meet the needs of receipt verification.

### In this PR
* Uses `mc_api::external.Receipt` as opposed to `mobilecoind_api::ReceiverReceipt`
* Returns Txo along with status
* Verifies on address as opposed to account
* Updates API docs

[FS-174](https://mobilecoin.atlassian.net/browse/FS-174)


christian-oudard (approved):
I think this is a good improvement to the API, and the implementation looks solid.

PR #88, opened 2021-03-20
Handle database-is-locked more gracefully

sugargoat:
Soundtrack of this PR: [Yo-Yo Ma: Hush Little Baby](https://www.youtube.com/watch?v=UTstU-st-7o)

### Motivation

Currently, the max db connections are set at 1, to avoid `database is locked` errors from SQLite. This PR makes some configuration changes and app-level changes to gracefully retry when database is locked.

### In this PR
* Enable WAL
* Gracefully requeue account if syncing fails due to database locked (this also fixes when the sync thread appeared to die and not recover)
* Remove sleeping from tests (except when in the controlled `wait_for_sync` method

[FS-162](https://mobilecoin.atlassian.net/browse/FS-162)
[FS-121](https://mobilecoin.atlassian.net/browse/FS-121)
[FS-156](https://mobilecoin.atlassian.net/browse/FS-156)

### DIFF
--- a/full-service/src/bin/main.rs
+++ b/full-service/src/bin/main.rs
@@ -47,7 +47,6 @@ fn main() {
                 config.wallet_db, err
             )
         });
-
     embedded_migrations::run(&conn).expect("failed running migrations");
 
     let wallet_db = WalletDb::new_from_url(
@@ -55,6 +54,7 @@ fn main() {
             .wallet_db
             .to_str()
             .expect("Could not get wallet_db path"),
+        10,
         logger.clone(),
     )
     .expect("Could not access wallet db");
--- a/full-service/src/config.rs
+++ b/full-service/src/config.rs
@@ -183,11 +183,7 @@ impl APIConfig {
                 let report_responses = conn
                     .fetch_fog_reports(fog_uris.iter().cloned())
                     .map_err(|err| format!("Failed fetching fog reports: {}", err))?;
-                log::debug!(
-                    logger,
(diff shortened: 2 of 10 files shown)


full-service/src/bin/main.rs
```diff
@@ -47,14 +47,14 @@ fn main() {
...
         });
-
     embedded_migrations::run(&conn).expect("failed running migrations");
 
     let wallet_db = WalletDb::new_from_url(
         config
             .wallet_db
             .to_str()
             .expect("Could not get wallet_db path"),
+        10,
```
BrianCorbin, line 57: I mean, 10 sounds good to me (though I doubt we'll ever hit more than 2), is this just a feel good number or is it a standard you came across for pooled connections?

BrianCorbin (approved):
Cool, I think this is probably the direction we want to head in. I can't think of any reason why we'd want to hard limit the DB connections to 1, as that seems to be something we _should_ be able to do no problem.

I'm still a little fresh off the boat in regards to full service and rust, so it may be good to have a second set of eyes double check to make sure I'm not overlooking something. Otherwise, it LGTM 👍

full-service/src/bin/main.rs
sugargoat: It's the default for pooled connections as shipped by diesel - but I don't have any other reason to set 10, perhaps when we're doing performance testing we'll figure out what's most optimal here

christian-oudard (approved):

PR #108, opened 2021-03-29
Add IP Check

sugargoat:
### Motivation

According to the [TERMS-OF-USE](https://github.com/mobilecoinofficial/full-service/blob/main/TERMS-OF-USE.md):

> Purchasers of MobileCoins may not transact, transfers, or trade MobileCoins in the United States or with U.S. Persons or entities, or persons or entities present in the United States.

This PR adds a check to disable full-service in IPs originating from the US.

### In this PR
* Adds IP check in full-service config

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2152,6 +2152,7 @@ dependencies = [
  "num_cpus",
  "prost",
  "rand 0.7.3",
+ "reqwest",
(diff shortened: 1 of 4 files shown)


christian-oudard (approved):
Seems to work correctly.

PR #112, opened 2021-03-31
🚑  allowing export of legacy account secrets

BrianCorbin:
EOM

### DIFF
--- a/full-service/src/json_rpc/account_secrets.rs
+++ b/full-service/src/json_rpc/account_secrets.rs
@@ -2,10 +2,7 @@
 
 //! API definition for the Account Secrets object.
 
-use crate::{
-    db::{account::ROOT_ENTROPY_KEY_DERIVATION_VERSION, models::Account},
-    json_rpc::account_key::AccountKey,
-};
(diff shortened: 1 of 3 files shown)


full-service/src/json_rpc/account_secrets.rs
```diff
@@ -22,9 +19,10 @@ pub struct AccountSecrets {
     /// The account ID for this account key in the wallet database.
     pub account_id: String,
 
-    /// The mnemonic from which this account key was derived, as a list of BIP39
-    /// words.
-    pub mnemonic: String,
+    /// The entropy from which this account key was derived, as a String.
+    /// Depending on the version, this will either be a hex encoded string
+    /// for version 1, or a 24 word mnemonic for version 2.
+    pub entropy: String,
```
sugargoat, line 25: I would suggest leaving `mnemonic` and adding `entropy`, and making each one `Option<String>` - then only populate based on version. 

I would also suggest adding `version` to the AccountSecrets struct

sugargoat (approved):

PR #114, opened 2021-04-02
Enable foreign key constraints

christian-oudard:
-

### DIFF
--- a/full-service/migrations/2021-03-30-021521_slip10_account_key/up.sql
+++ b/full-service/migrations/2021-03-30-021521_slip10_account_key/up.sql
@@ -1,2 +1,34 @@
-ALTER TABLE accounts
-ADD COLUMN key_derivation_version INTEGER NOT NULL DEFAULT 1;
+-- ALTER TABLE accounts ADD COLUMN key_derivation_version INTEGER NOT NULL DEFAULT 1;
+PRAGMA foreign_keys=OFF;
+CREATE TABLE NEW_accounts (
+    id INTEGER NOT NULL PRIMARY KEY,
+    account_id_hex VARCHAR NOT NULL UNIQUE,
+    account_key BLOB NOT NULL,
+    entropy BLOB NOT NULL,
+    key_derivation_version INTEGER NOT NULL DEFAULT 1,
+    main_subaddress_index UNSIGNED BIG INT NOT NULL,
+    change_subaddress_index UNSIGNED BIG INT NOT NULL,
+    next_subaddress_index UNSIGNED BIG INT NOT NULL,
+    first_block_index UNSIGNED BIG INT NOT NULL,
+    next_block_index UNSIGNED BIG INT NOT NULL,
+    import_block_index UNSIGNED BIG INT,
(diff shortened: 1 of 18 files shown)


full-service/migrations/2021-03-30-021521_slip10_account_key/up.sql
```diff
@@ -1,2 +1,34 @@
-ALTER TABLE accounts
-ADD COLUMN key_derivation_version INTEGER NOT NULL DEFAULT 1;
+-- ALTER TABLE accounts ADD COLUMN key_derivation_version INTEGER NOT NULL DEFAULT 1;
```
sugargoat, line 1: Chatted with @christian-oudard - this is rewriting the migration to be compatible with different SQLite3 versions

full-service/migrations/2021-03-31-220723_nullable_transaction_log_account_id/up.sql
```diff
@@ -0,0 +1,26 @@
+-- ALTER TABLE transaction_logs ALTER COLUMN account_id_hex DROP NOT NULL;
```
sugargoat, line 1: Note: for foreign key constraints, we need to have nullable fields. The downside is that there are now "two ways" for a string to be "null" (either null or empty string), but the benefit of having proper FK constraints is worth it.

christian-oudard: The empty string will never pass foreign key constraints, so this is not a problem.

full-service/src/db/models.rs
```diff
@@ -227,14 +227,14 @@ pub struct NewAssignedSubaddress<'a> {
 pub struct TransactionLog {
     pub id: i32,
     pub transaction_id_hex: String,
-    pub account_id_hex: String,
+    pub account_id_hex: Option<String>,
     pub recipient_public_address_b58: String, // empty string for nullable
-    pub assigned_subaddress_b58: String,      // empty string for nullable
+    pub assigned_subaddress_b58: Option<String>, // empty string for nullable
```
sugargoat, line 232:
```suggestion
    pub assigned_subaddress_b58: Option<String>,
```

full-service/src/db/transaction_log.rs
```diff
@@ -509,14 +517,13 @@ impl TransactionLogModel for TransactionLog {
...
                 let transaction_id = TransactionID::from(&tx_proposal.tx);
                 let tx = mc_util_serial::encode(&tx_proposal.tx);
+
                 // Create a TransactionLogs entry
                 let new_transaction_log = NewTransactionLog {
                     transaction_id_hex: &transaction_id.to_string(),
-                    account_id_hex: &account_id_hex.unwrap_or(""), /* Can be empty str if
-                                                                    * submitting an "unowned"
-                                                                    * proposal */
+                    account_id_hex, // Can be null if submitting an "unowned" proposal.
```
sugargoat, line 524: We've previously introduced the concept that you must provide an account_id to log submitted, so this can actually be non-null

sugargoat (approved):

PR #116, opened 2021-04-02
Display u64 in Origin Block

sugargoat:
### Motivation

When investigating the origin block for an upcoming blog post, it became apparent that the jsonrpc layer was not displaying the value correct for a large u64. This PR fixes that.

### In this PR
* casts the DB's i64 to u64 before displaying
* Adds tests for sync and jsonrpc display of txo


BrianCorbin (approved):
This looks good to me!

christian-oudard (approved):

PR #121, opened 2021-04-06
Add pagination.

christian-oudard:
Add limit and offset parameters to these RPC calls:
* get_all_txos_for_account
* get_all_transaction_logs_for_account
* get_all_addresses_for_account


sugargoat (approved):
Looks great! We could maybe drop the "all" in the names, and leave the "all" endpoints as wrappers that provide "None" to the pagination params.

BrianCorbin:
> Looks great! We could maybe drop the "all" in the names, and leave the "all" endpoints as wrappers that provide "None" to the pagination params.

Yeah, I like this suggestion I think it will definitely help avoid confusion :)

PR #123, opened 2021-04-10
Updated Assigned Subaddress B58 of transaction log when orphaned TXO is picked up by a subaddress

BrianCorbin:
Quick fix for bug with transaction logs assigned subaddress index not updating when an orphaned txo was claimed by a subaddress


christian-oudard:
Is this worth a unittest or is it too simple?

christian-oudard (approved):

BrianCorbin:
I think we can just wait until we refactor for the unit test, as it's fairly trivial anyway and will be changing with a few other things re: transaction logs really soon.

PR #124, opened 2021-04-12
Allow for multiple recipients.

christian-oudard:
* Allow building transactions with multiple recipients, each receiving
  different amounts

* Change transaction logging to record multiple recipient addresses and amounts.


full-service/src/json_rpc/json_rpc_request.rs
```diff
@@ -113,6 +113,14 @@ pub enum JsonCommandRequest {
         tombstone_block: Option<String>,
         max_spendable_value: Option<String>,
     },
+    build_multi_transaction {
```
BrianCorbin, line 116: We need to chat with @sugargoat and @davidyhall about making multi-lay transactions the only build transaction API. Also need to discuss if we should change build_and_submit to this style as well.

christian-oudard: This is done. There is now just one endpoint supporting both APIs.

BrianCorbin (approved):
lgtm

PR #126, opened 2021-04-20
Change default tombstone blocks to 10.

christian-oudard:


BrianCorbin (commented):
This looks good to me. Would like to quickly chat about if we indeed do want to default it at maximum tombstone block, though.

eranrund:
I think defaulting to the highest possible value is suboptimal. The reasoning for the tombstone block is to limit how much time the client has to wait before learning their transaction did not go through. The bigger the value used here is, the longer clients would be forced to wait. In most cases (everything but offline transactions, in my opinion), a small value is desirable.
Transactions can fail to go through for all sorts of reasons - consensus being overloaded, the node crashing before getting to processing their queued transaction, future fee stuff that causes the tx to be constantly de-prioritized, etc. The only way clients can move on after submitting a transaction is either when it gets into the ledger or when the block is exceeded. A high value, unless the network is creating blocks at a high and constant pace, could result in the client losing access to the pending TxOuts for a long duration of time.

BrianCorbin:
@christian-oudard let's set the default to 10 and adjust from there, without being overly scientific about it. I'd like to check out mainnet block timestamps (if that's possible) to see the avg time between blocks, but until then this should suffice.

christian-oudard:
Rebasing somehow got messed up and I couldn't push the PR branch. Made a new one here: https://github.com/mobilecoinofficial/full-service/pull/138

PR #129, opened 2021-04-27
Statically link the correct version of sqlite3

eranrund:
### Motivation

Statically linking sqlite3 means we will have an easier time distributing binaries. Ensuring that all users are running with the same version would make debugging easier if any issues arise. 

### In this PR
* Switch to a fork of `diesel` that allows using a newer version of the `libsqlite3-sys` crate. That version has the required sqlite3 3.35.4 bundled.

### DIFF
--- a/full-service/src/db/wallet_db.rs
+++ b/full-service/src/db/wallet_db.rs
@@ -29,6 +29,8 @@ impl diesel::r2d2::CustomizeConnection<SqliteConnection, diesel::r2d2::Error>
             }
             if self.enable_foreign_keys {
                 conn.batch_execute("PRAGMA foreign_keys = ON;")?;
+            } else {
+                conn.batch_execute("PRAGMA foreign_keys = OFF;")?;
             }
             if let Some(d) = self.busy_timeout {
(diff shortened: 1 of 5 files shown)


sugargoat (approved):

full-service/src/db/wallet_db.rs
```diff
@@ -29,6 +29,8 @@ impl diesel::r2d2::CustomizeConnection<SqliteConnection, diesel::r2d2::Error>
             }
             if self.enable_foreign_keys {
                 conn.batch_execute("PRAGMA foreign_keys = ON;")?;
+            } else {
+                conn.batch_execute("PRAGMA foreign_keys = OFF;")?;
```
eranrund, line 33: The default bundled sqlite3 version has foreign keys enabled by default. I think it is significant tech debt that the tests don't run with this turned on and I suggest fixing it gets prioritized.

eranrund:
I also noticed this is not using the https://github.com/mobilecoinofficial/AEADs fork, so I am going to fix that as well in this PR.

BrianCorbin (approved):

PR #132, opened 2021-04-28
feature/split_txo

BrianCorbin:
Soundtrack of this PR: [Split Dance](https://www.youtube.com/watch?v=FnFMS11g4fM)

### Motivation

Making it easier for users to split their TXO's

### In this PR
* Added a helper API call called "split_txo" that allows users to select a txo to split into multiple values.

### Future Work
None

### To Dos

- [x] Update tests

### DIFF
--- a/full-service/src/db/assigned_subaddress.rs
+++ b/full-service/src/db/assigned_subaddress.rs
@@ -262,22 +262,13 @@ impl AssignedSubaddressModel for AssignedSubaddress {
         index: i64,
         conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
     ) -> Result<AssignedSubaddress, WalletDbError> {
-        use crate::db::schema::assigned_subaddresses::dsl::{
-            account_id_hex as dsl_account_id_hex, assigned_subaddresses, subaddress_index,
-        };
+        let account = Account::get(&AccountID(account_id_hex.to_string()), conn)?;
 
-        let assigned_subaddress: AssignedSubaddress = match assigned_subaddresses
-            .filter(dsl_account_id_hex.eq(account_id_hex))
-            .filter(subaddress_index.eq(index))
-            .get_result::<AssignedSubaddress>(conn)
-        {
-            Ok(t) => t,
-            Err(e) => {
-                return Err(e.into());
-            }
-        };
+        let account_key: AccountKey = mc_util_serial::decode(&account.account_key)?;
+        let subaddress = account_key.subaddress(index as u64);
 
-        Ok(assigned_subaddress)
+        let subaddress_b58 = b58_encode(&subaddress)?;
+        Ok(Self::get(&subaddress_b58, &conn)?)
     }
 
     fn find_by_subaddress_spend_public_key(
--- a/full-service/src/db/assigned_subaddress.rs
+++ b/full-service/src/db/assigned_subaddress.rs
@@ -66,6 +66,14 @@ pub trait AssignedSubaddressModel {
         conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
     ) -> Result<AssignedSubaddress, WalletDbError>;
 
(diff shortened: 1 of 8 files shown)


full-service/src/db/assigned_subaddress.rs
```diff
@@ -249,6 +257,29 @@ impl AssignedSubaddressModel for AssignedSubaddress {
...
+    fn get_for_account_by_index(
+        account_id_hex: &str,
+        index: i64,
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<AssignedSubaddress, WalletDbError> {
+        use crate::db::schema::assigned_subaddresses::dsl::{
+            account_id_hex as dsl_account_id_hex, assigned_subaddresses, subaddress_index,
+        };
+
+        let assigned_subaddress: AssignedSubaddress = match assigned_subaddresses
```
sugargoat, line 269: The other way to do this is to derive the `public_address_b58` from the account_key - then you can call `AssignedAddress::get`, which looks up by the indexed field.

full-service/src/json_rpc/wallet.rs
```diff
@@ -590,6 +590,32 @@ where
                 txo: Txo::from(&result),
             }
         }
+        JsonCommandRequest::split_txo {
+            txo_id,
+            output_values,
+            subaddress_index,
```
sugargoat, line 596: Part of me thinks we should maybe pass a public address rather than a subaddress_index, so that you could potentially `split_txo` to an abitrary address, but that's not actually required by the spec. Alternatively, would maybe rename this `destination_subaddress_index` to make it clear that it's not the source of the funds.

full-service/src/service/txo.rs
```diff
@@ -86,6 +114,57 @@ where
...
+        txo_id: &TxoID,
+        output_values: &[String],
+        subaddress_index: Option<i64>,
+        fee: Option<String>,
+        tombstone_block: Option<String>,
+    ) -> Result<(TransactionLog, AssociatedTxos), TxoServiceError> {
+        use crate::service::txo::TxoServiceError::{TxoNotSpendable, TxoNotSpendableByAnyAccount};
+
+        let conn = self.wallet_db.get_conn()?;
+        let txo_details = Txo::get(&txo_id.to_string(), &conn)?;
```
sugargoat, line 128: I think this should all go into a db transaction

full-service/src/json_rpc/wallet.rs
BrianCorbin: I was thinking that as well, but my gut is saying that this should restrict splitting a txo to remain on the account that owns it. I guess one could still perform a check to make sure the subaddress belongs to the account that owns the txo, but this seemed more straightforward. Do you have a strong preference either way, or maybe even allowing either an index or a subaddress_b58?

full-service/src/service/txo.rs
BrianCorbin: 👍

full-service/src/json_rpc/wallet.rs
BrianCorbin: To follow that up, if a user wants to _move_ the txo to a different subaddress that _may or may not_ be a part of the account, then the normal build_transaction or build_and_submit_transaction endpoints would be appropriate

sugargoat: That works for me :) Let's restrict `split_txo` to remain in the same account, and just provide the index

BrianCorbin:
@sugargoat I've made the split_txo method return a tx_proposal instead of doing everything all-in-one. This is in part for testing via the e2e tests, but I also feel this is the more correct way to do this, giving the user a chance to review the tx_proposal if they wish, and then submitting it like a normal transaction via the submit_transaction method.

christian-oudard (approved):

PR #136, opened 2021-05-04
Uprev mobilecoin and deal with new Rust version fallbacks

eranrund:
The `mobilecoin` repository has moved to a newer Rust toolchain, and as a result a bunch of small stuff had to be fixed. This PR uprevs `mobilecoin` and fixes the small issues.


christian-oudard (approved):
These all look like good updates. Approved.

PR #139, opened 2021-05-05
Clean up usage of database transactions.

christian-oudard:
* Move all usage of transactions out of the db layer and into the service
layer.
* Every time we create a connection with get_conn(), start a db transaction.


mfaulk (approved):
Thanks! This feels like a good pattern.

PR #146, opened 2021-05-17
Use mobilecoinofficial fork of diesel in order to statically link sqlite.

christian-oudard:


Cargo.toml
```diff
@@ -36,8 +36,8 @@ grpcio = { git = "https://github.com/jcape/grpc-rs", rev = "2ad042e9e65ecb664a60
 grpcio-compiler = { git = "https://github.com/jcape/grpc-rs", rev = "2ad042e9e65ecb664a60e034d0a899a8760d81d3" }
 
 # Override diesel dependency with our fork, in order to use a version of libsqlite3-sys that has bundled-sqlcipher. This allows us to
-# statically link SQLite, and encrypt the database with SQLCipher
-diesel = { git = "https://github.com/mobilecoinofficial/diesel", branch = "sqlite-static-attempt" }
+# statically link SQLite.
+diesel = { git = "https://github.com/mobilecoinofficial/diesel", branch = "1.4.x-bundled-sqlcipher", features = ["sqlite", "bundled"] }
```
eranrund, line 40: Please pin it to a specific commit. This has been our policy everywhere else and it protects from unexpected changes. The same goes for the changes in the diesel fork that reference https://github.com/mobilecoinofficial/rusqlite

eranrund (changes requested):
Does this moves full-service to use sqlcipher? If so, are existing database compatible with this change?

eranrund (approved):
LGTM once CI is green

PR #148, opened 2021-05-18
Block IP addresses in OFAC sanctioned countries.

christian-oudard:
-

### DIFF
--- a/full-service/src/bin/main.rs
+++ b/full-service/src/bin/main.rs
@@ -31,8 +31,10 @@ fn main() {
 
     let config = APIConfig::from_args();
 
-    if !cfg!(debug_assertions) && !config.offline {
-        config.validate_host().expect("Could not validate host");
+    // Exit if the user is not in an authorized country.
+    if !cfg!(debug_assertions) && !config.offline && config.validate_host().is_err() {
+        println!("Could not validate host");
+        return;
     }
 
     let (logger, _global_logger_guard) = create_app_logger(o!());
--- a/full-service/src/config.rs
+++ b/full-service/src/config.rs
@@ -364,11 +364,7 @@ impl APIConfig {
 
         let err = Err(ConfigError::InvalidCountry);
         match country {
-            "US" => err,
-            "IR" => err,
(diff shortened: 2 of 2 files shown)


full-service/src/config.rs
```diff
@@ -351,17 +351,29 @@ impl APIConfig {
...
-            }
-        } else {
-            Err(ConfigError::DataMissing(data_json.to_string()))
+
+        let data_missing_err = Err(ConfigError::DataMissing(data_json.to_string()));
+        let country: &str = match data_json["country"].as_str() {
+            Some(c) => c,
+            None => return data_missing_err,
+        };
+        let region: &str = match data_json["region"].as_str() {
```
eranrund, line 360: Would there always be a region for all countries? Maybe its better to only look at it when necessary (`"UA"` only, it seems)

full-service/src/config.rs
```diff
@@ -351,17 +351,29 @@ impl APIConfig {
...
+            Some(r) => r,
+            None => return data_missing_err,
+        };
+
+        let err = Err(ConfigError::InvalidCountry);
+        match country {
+            "US" => err,
```
eranrund, line 367: You should be able to do `"US" | "IR" | "SY" | ...` which I think is more pleasant to look at but if clippy doesn't prefer one way or the other I don't feel strongly about it.

eranrund (changes requested):
Thank you!

full-service/src/bin/main.rs
```diff
@@ -31,8 +31,10 @@ fn main() {
 
     let config = APIConfig::from_args();
 
-    if !cfg!(debug_assertions) && !config.offline {
-        config.validate_host().expect("Could not validate host");
+    // Exit if the user is not in an authorized country.
+    if !cfg!(debug_assertions) && !config.offline && config.validate_host().is_err() {
```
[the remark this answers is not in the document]
eranrund: Looks like this is already the case, I missed it.

eranrund (approved):

PR #149, opened 2021-05-18
Scrub unreferenced Txos from the database when an account is deleted.

christian-oudard:
-


sugargoat (changes requested):
Awesome - would maybe add an e2e test adding two accounts, sending a transaction from one to the other, deleting one, and confirming only the txos from the deleted one were scrubbed, and the other still has all their Txos.

sugargoat (approved):

PR #151, opened 2021-05-19
Wallet database encryption via SQLCipher.

christian-oudard:
* Embed SQLCipher library by static linking.
* Use MC_PASSWORD environment variable to set a password and activate database
  encryption.
* Use MC_CHANGE_PASSWORD environment variable to change database password.

### DIFF
--- a/DATABASE.md
+++ b/DATABASE.md
@@ -1,0 +1,28 @@
+# Database Usage
+
+The MobileCoin wallet service stores its data in two places:
+- The first database is the ledger database, which is an append-only data store, recording the state of the public ledger. This contains no readable secret information, because all transaction outputs are stored on the ledger in an encrypted format. Its location is specified with the command line option `--ledger_db`.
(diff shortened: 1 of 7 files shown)


christian-oudard:
Commenting after a discussion with @eranrund. The reason we had to use manual quoting of the password is that the PRAGMA call in sqlite does not accept bind variables.

DATABASE.md
```diff
@@ -0,0 +1,28 @@
+# Database Usage
+
+The MobileCoin wallet service stores its data in two places:
+- The first database is the ledger database, which is an append-only data store, recording the state of the public ledger. This contains no readable secret information, because all transaction outputs are stored on the ledger in an encrypted format. Its location is specified with the command line option `--ledger_db`.
```
eranrund, line 4: Nit: `--ledger-db`, similarly to `--wallet-db`

eranrund (approved):
I think this looks good but I have not played with it locally. Would appreciate if a more active user of full-service gives this a shot.

DATABASE.md
christian-oudard: Good catch.

PR #153, opened 2021-05-21
Use a fallback fee of 10mMOB, instead of 400μMOB.

christian-oudard:
-

### DIFF
--- a/full-service/src/service/ledger.rs
+++ b/full-service/src/service/ledger.rs
@@ -15,7 +15,7 @@ use mc_fog_report_validation::FogPubkeyResolver;
 use mc_ledger_db::Ledger;
 use mc_ledger_sync::NetworkState;
 use mc_transaction_core::{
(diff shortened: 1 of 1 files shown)


jcape (approved):

full-service/src/service/ledger.rs
```diff
@@ -139,7 +144,7 @@ where
                 .filter_map(|conn| conn.fetch_block_info(empty()).ok())
                 .map(|block_info| block_info.minimum_fee)
```
jcape, line 145: Eran pointed out the fix is incomplete, because protobuf will default the minimum_fee member to zero, not `None`:

```suggestion
                .filter_map(|block_info| {
                    // Cleanup the protobuf default fee
                    if block_info.minimum_fee == 0 {
                        None
                    } else {
                        Some(block_info.minimum_fee)
                    }
                })
```

jcape (changes requested):

jcape (approved):

PR #157, opened 2021-05-26
Update error reporting format to match JSON-RPC 2.0 spec.

christian-oudard:


BrianCorbin (approved):
@christian-oudard could you review the few things I've added? Everything else looks good to me.

PR #160, opened 2021-05-30
Feature/payment request

BrianCorbin:
Soundtrack of this PR: [Down The (Rabbit) Hole](https://www.youtube.com/watch?v=tfwf-e4rJr4)

### Motivation

We want to add the ability to generate a payment request with Full Service, which will allow a user to send a b58 encoded message containing a public address, value in pmob, and memo.

### In this PR
* Added create_payment_request
* Added check_b58
* Refactored b58 utils into a new utils crate
* Added transfer_payload and payment_request support to b58 mod
* Added tests to b58 utils

### Future Work
* Continued refactoring of all files to removed tests and errors from the actual implementation of the code, similar to the b58 collection.


docs/transactions/payment-request/create_payment_request.md
```diff
@@ -0,0 +1,52 @@
...
+## Parameters
+
+| Required Param | Purpose | Requirements |
+| :--- | :--- | :--- |
+| `account_id` | The account on which to perform this action. | Account must exist in the wallet. |
+| `amount_pmob` | The amount of pMOB to send in this transaction. | `u64` |
+
+| Optional Param | Purpose | Requirements |
+| :--- | :--- | :--- |
+| `subaddress_index` | The subaddress index on the account to send the transaction | `i64` |
```
eranrund, line 16: I don't think `send` makes sense here. This is the address at which you want the recipient to pay you.

docs/transactions/payment-request/create_payment_request.md
```diff
@@ -0,0 +1,52 @@
...
+## Example
+
+{% tabs %}
+{% tab title="Request Body" %}
+```text
+{
+  "method": "create_payment_request",
+  "params": {
+    "account_id": "a8c9c7acb96cf4ad9154eec9384c09f2c75a340b441924847fe5f60a41805bde",
+    "amount_pmob": 42000000000000,
```
eranrund, line 28: Int or string?

docs/transactions/payment-request/create_payment_request.md
```diff
@@ -0,0 +1,52 @@
...
+    "subaddress_index": 4,
+    "memo": "Happy Birthday!"
```
eranrund, line 30: Odd nit, but this memo might be confusing. You're asking someone to pay you and saying them happy birthday?

full-service/src/service/payment_request.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+impl From<LedgerServiceError> for PaymentRequestServiceError {
+    fn from(src: LedgerServiceError) -> Self {
+        Self::LedgerService(src)
+    }
+}
+
+/// Trait defining the ways in which the wallet can interact with and manage
+/// accounts.
```
christian-oudard, line 80: Was this comment an accident?

BrianCorbin: Yes, thanks for the catch!

full-service/src/util/b58/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+        wrapper.get_public_address()
+    } else {
+        return Err(B58Error::NotPublicAddress);
+    };
+
+    let public_address = PublicAddress::try_from(pubaddr_proto)?;
+    Ok(public_address)
+}
+
+pub fn decode_public_address_b58(public_address_b58_code: &str) -> Result<PublicAddress, B58Error> {
```
eranrund, line 73: Whats the difference between this and `b58_decode`?
Why would you ever want to use this one?

eranrund: (Is it actually used at all?)

eranrund (approved):
Some nits and small questions but other than that looks good as far as I can tell.

docs/transactions/payment-request/create_payment_request.md
BrianCorbin: Int. Unless there is a reason to not be using it, I was curious why every field we have are strings.

christian-oudard (approved):

PR #162, opened 2021-06-03
Allow fees less than the transaction crate's default `MINIMUM_FEE`

jcape:
### Motivation

We discovered that the WalletTransactionBuilder is actually unable to take fees lower than the default MINIMUM_FEE, even though this is technically legal per [RFC 1](https://github.com/mobilecoinfoundation/rfcs/#1).

### In this PR
* Updates the setter to allow fees as low as 1pMOB.


BrianCorbin (approved):

BrianCorbin:
LGTM, just need to fix some linting errors, otherwise it's good :)

BrianCorbin:
redone in https://github.com/mobilecoinofficial/full-service/pull/169

PR #178, opened 2021-07-29
Add some utilities for handling b64-encoded addresses/receipts

eranrund:
### Motivation

To ease integration with `signald`, following the merge of a recent commit that added MobileCoin payments support, we want to add some utility methods for converting between the B58-encodings of Public Addresses and Receipts that full-service uses, and the B64-encodings that `signald` uses.

### In this PR
* The utility methods mentioned above

### Future Work
* Our Python packaging/libraries are a mess. We are increasing the role Python plays in our organization and this is due for a cleanup. This includes enforcing standards (black, flake8), CI, proper packaging + pip release, etc.
* This PR simply hardcodes the protobuf-generated Python code, but that is janky. There should be a way to easily build these files from the .proto files, and CI should alert if the .proto file has changed.

Thank you @BrianCorbin for doing a good chunk of the work on this.


eranrund:
Not sure why @BrianCorbin grabbed the CI commits into this branch. If it passes, I'll let them stay. If not, they're going out.

cli/mobilecoin/external_pb2.py
```diff
@@ -0,0 +1,1419 @@
+# -*- coding: utf-8 -*-
+# Generated by the protocol buffer compiler.  DO NOT EDIT!
```
sugargoat, line 2: Do we want to commit the generated code? Maybe that's the most expedient?

sugargoat (approved):
LGTM

cli/mobilecoin/external_pb2.py
eranrund: For now it is, but as I mentioned in the PR description this is janky

PR #179, opened 2021-08-10
Py module fixes

eranrund:
Hack around Python imports and add missing dependencies so that the Python lib is pip-installable


queenbdubs (approved):
nice

PR #193, opened 2021-09-02
Add exit codes to indicate full-service process exit reason.

christian-oudard:


BrianCorbin (approved):

BrianCorbin (changes requested):
Some lint issues to fix :)

PR #194, opened 2021-09-02
added option to log a transaction when building

BrianCorbin:


eranrund:
What is this trying to achieve? 
Is this related to exposing the tx_proposal in build_and_submit_transaction or unrelated?

BrianCorbin:
@eranrund it's unrelated to that. This is to allow logging of transactions for offline wallets, which are unable to use submit or build_and_submit functionality. Adding this flag will let offline wallets optionally log their tx_proposals, which will be updated when the ledger gets scanned the next time they import it.

full-service/src/service/transaction.rs
```diff
@@ -213,8 +226,20 @@ where
 
         let tx_proposal = builder.build()?;
 
-        // FIXME: WS-32 - Might be nice to have a tx_proposal table so that you don't
-        // have to write these out to local files.
+        if let Some(should_log) = log_tx_proposal {
```
eranrund, line 229: Instead of nesting you should be able to do something like `if log_tx_proposal.unwrap_or_default(false) {`

eranrund (approved):

PR #198, opened 2021-09-16
Optionally return tx_proposal with build_and_submit_transaction

monkeymantra:
### Motivation

Currently, we use the python CLI to build and submit in two separate actions because we need a tx_proposal. For throughput and to allow multithreading without accidentally sending the same uTXO, we would like to pull the tx_proposal from build_and_submit_transaction.

### In this PR
* added to `build_and_submit_transaction_with_proposal` in the python client

[Python client: get tx_proposal from build_and_submit_transactions](https://app.asana.com/0/1200175610893108/1200918570741698) 

### Future Work
* [MOBot: Migrate to new python client for full-service API](https://app.asana.com/0/1200175610893108/1201002142005233)

### DIFF
--- a/cli/mobilecoin/client.py
+++ b/cli/mobilecoin/client.py
@@ -228,7 +228,7 @@ class Client:
 
     def build_and_submit_transaction_with_proposal(self, account_id, amount, to_address):
         r = self._build_and_submit_transaction(account_id, amount, to_address)
-        return r['transaction_log'], r['proposal']
+        return r['transaction_log'], r['tx_proposal']
 
     def build_transaction(self, account_id, amount, to_address, tombstone_block=None):
         amount = str(mob2pmob(amount))
--- a/cli/mobilecoin/client.py
+++ b/cli/mobilecoin/client.py
@@ -211,7 +211,7 @@ class Client:
         })
         return r['address_map']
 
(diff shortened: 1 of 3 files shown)


cli/mobilecoin/client.py
```diff
@@ -220,8 +220,16 @@ def build_and_submit_transaction(self, account_id, amount, to_address):
                 "addresses_and_values": [(to_address, amount)],
             }
         })
+        return r
+
+    def build_and_submit_transaction(self, account_id, amount, to_address):
+        r = self._build_and_submit_transaction(account_id, amount, to_address)
         return r['transaction_log']
 
+    def build_and_submit_transaction_with_proposal(self, account_id, amount, to_address):
```
eranrund, line 229: I think we should just return the entire response. This approach doesn't scale as more data is returned. Suppose we start returning information about the node we submitted the transaction to. Will we now have `build_and_submit_transaction_with_proposal_and_node`?
With that said, I don't feel strongly about it so I am going to approve and defer to @christian-oudard since he wrote most of this.

While we're at the topic of breaking changes, I think if we are starting to treat this client code as "the official full-service python client", we should not be exposing an API that limits the functionality - this API limits to a single outlay, but full-service actually supports multiple outlays.

eranrund (approved):

PR #202, opened 2021-11-05
Put export_account_secrets behind a feature flag.

NotGyro:
### Motivation

The export_account_secrets jsonrpc method was discussed as a possible security risk, if full-service's jsonrpc interface is ever publicly-exposed over http by mistake.

### In this PR
* Adds an "export_account_secrets" feature flag in cargo.toml
* The export_account_secrets jsonrpc method returns an error instead of evaluating if this flag is disabled. 
* Wrote a test-case to ensure that the jsonrpc method returns an error when the feature is disabled, using dispatch_expect_error()

Fixes Asana ticket "Ability to build full-service without exporting account key" 

### Future Work
* Determine if the "export_account_secrets" Cargo feature shouldn't be default - right now, it's enabled by default, but that may not be ideal.
* Evaluate options for more thoroughly-securing the whole API - should full-service jsonrpc commands require some kind of authorization / key exchange?


eranrund:
Hi! Congrats on your first PR!

My personal take on this is that this is a misfeature, since while it does protect the account keys, money can still be stolen and as such might provide a false sense of security.

I think a better solution would be to have an optional command line configuration flag/environment flag that specifies an API key. It provided, full-service should decline requests that do not pass it.
Since communication with full-service is currently plaintext (unless someone fronts it with something that handles SSL), it is still not amazing but I think is an improvement over only removing this single endpoint. I think this shouldn't be too complicated to implement.

Open to hearing any other thoughts and opinions on this!

NotGyro:
It looks like `cargo test` is failing even though it worked on my machine - it seems like I mistakenly forked this from `main` and then tried to create a PR on it into `develop`. I can pull changes from `develop` and see if there are any merge conflicts to resolve.

christian-oudard (approved):

BrianCorbin:
> Hi! Congrats on your first PR!
> 
> My personal take on this is that this is a misfeature, since while it does protect the account keys, money can still be stolen and as such might provide a false sense of security.
> 
> I think a better solution would be to have an optional command line configuration flag/environment flag that specifies an API key. It provided, full-service should decline requests that do not pass it. Since communication with full-service is currently plaintext (unless someone fronts it with something that handles SSL), it is still not amazing but I think is an improvement over only removing this single endpoint. I think this shouldn't be too complicated to implement.
> 
> Open to hearing any other thoughts and opinions on this!

I agree, this should be handled with an API key, which is currently in the works! :) Link to PR https://github.com/mobilecoinofficial/full-service/pull/205

PR #206, opened 2021-11-15
bugfix for tx selection: 

eranrund:
Prior to this change full-service would select the smallest TXOs, even though all previous checks look at the large ones.
This results in incorrect txo selection (total sum of selected txos would be less than desired amount), even when the account has sufficient txos.

This should've been caught by unit tests, so a unit test still needs to be written to cover this.


full-service/src/db/txo.rs
```diff
@@ -1847,11 +1848,156 @@ mod tests {
...
+            None,
+            &wallet_db.get_conn().unwrap(),
+        );
+        assert!(result.is_err());
+
+        // testing over the max spendable but under the total in account, should fail
+        let result = Txo::select_unspent_txos_for_value(
+            &account_id.to_string(),
+            201 as u64,
+            None,
```
eranrund, line 1896: This doesn't match the comment above (max spendable is None here)

full-service/src/db/txo.rs
```diff
@@ -1847,11 +1848,156 @@ mod tests {
...
+            &wallet_db.get_conn().unwrap(),
+        );
+        assert!(result.is_err());
+
+        // testing at the max spendable limit, should pass and have 16 txouts selected
+        let result = Txo::select_unspent_txos_for_value(
+            &account_id.to_string(),
+            200 as u64,
+            None,
```
eranrund, line 1905: Max spendable not being set

full-service/src/db/txo.rs
```diff
@@ -1847,11 +1848,156 @@ mod tests {
...
+            &wallet_db.get_conn().unwrap(),
+        )
+        .unwrap();
+        assert_eq!(result.len(), 16);
+        let sum = result
+            .into_iter()
+            .map(|x| x.value)
+            .reduce(|a, b| a + b)
```
eranrund, line 1913: Isnt this `.sum()`?

full-service/src/db/txo.rs
```diff
@@ -1847,11 +1848,156 @@ mod tests {
...
+            .unwrap();
+        assert_eq!(3 as i64, sum);
+    }
+
+    #[test_with_logger]
+    fn test_eran_issue(logger: Logger) {
```
eranrund, line 1935: Interesting name... :p

full-service/src/db/txo.rs
```diff
@@ -1847,11 +1848,156 @@ mod tests {
...
+            src_txos.push(txo);
+        }
+
+        for i in 1..=500 {
+            let (_txo_id, txo, _key_image) =
+                create_test_received_txo(&account_key, i, 100000000, i, &mut rng, &wallet_db);
+            src_txos.push(txo);
+        }
+
+        // testing over the total in account, should fail
```
eranrund, line 1980: This doesn't match what the test is testing for

eranrund (commented):
Thank you for dealing with all the flaky tests!
I cant select "Request changes" because its my own PR, but there are a few places where the test description doesn't match whats actually being tested.

full-service/src/db/txo.rs
BrianCorbin: O_O

full-service/src/db/txo.rs
BrianCorbin: Oh! hah. will update that :)

full-service/src/db/txo.rs
eranrund: Resolved over slack, this test checks for max spendable by the account and not by filtering for specific utxos that are under a certain value.

BrianCorbin: update should have clarified this, but max spendable can have 2 meanings. In the context of the select_unspent_txos_for_value function, max_spendable_value is referring to the largest value an individual txo can have to be considered valid for the sake of selecting it for a transaction. In the context of the test, max_spendable_in_account is referring to "what is the maximum spendable amount of MOB that I can have be part of a single transaction"

eranrund:
LGTM, thanks @BrianCorbin 
CI failure seems unrelated.

PR #210, opened 2021-12-06
Get Subaddress by Index

BrianCorbin:
added functionality for getting a subaddress by its index. The address must have already been assigned to prevent users for accidentally getting an address at some random integer and effectively losing the txo's that end up there.

### DIFF
--- a/docs/accounts/address/get_address_for_account.md
+++ b/docs/accounts/address/get_address_for_account.md
@@ -1,0 +1,52 @@
+---
+description: Get an assigned address by index for an account.
+---
+
+# Get Address For Account
+
+## Parameters
+
+| Required Param | Purpose | Requirements |
+| :--- | :--- | :--- |
+| `account_id` | The account on which to perform this action. | The account must exist in the wallet. |
+| `index` | The subaddress index to lookup | The address must have already been assigned. |
+
+## Example
+
+{% tabs %}
+{% tab title="Request Body" %}
+```text
+{
+  "method": "get_address_for_account",
+  "params": {
+    "account_id": "a8c9c7acb96cf4ad9154eec9384c09f2c75a340b441924847fe5f60a41805bde",
+    "index": 1
+  },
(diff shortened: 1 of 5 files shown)


full-service/src/json_rpc/json_rpc_request.rs
```diff
@@ -133,6 +133,10 @@ pub enum JsonCommandRequest {
     get_account_status {
         account_id: String,
     },
+    get_address_for_account {
```
eranrund, line 136: Is it returning the public address? b58 encoded?

full-service/src/json_rpc/json_rpc_response.rs
```diff
@@ -180,6 +180,9 @@ pub enum JsonCommandResponse {
         account: Account,
         balance: Balance,
     },
+    get_address_for_account {
+        address: serde_json::Value,
```
eranrund, line 184: Can you show an example of whats being returned here?

eranrund (commented):
I'm a bit confused at what data is being returned, and what the purpose of this thing is - please see the inline comments.
Can you please update the API docs as well?

full-service/src/json_rpc/json_rpc_request.rs
BrianCorbin: The full address object. Updated to specify the type instead of generic json value

docs/accounts/address/get_address_for_account.md
```diff
@@ -0,0 +1,52 @@
...
+        "account_id": "3407fbbc250799f5ce9089658380c5fe152403643a525f581f359917d8d59d52",
+        "metadata": "Main",
+        "subaddress_index": "0"
+    }
+  },
+  "error": null,
+  "jsonrpc": "2.0",
+  "id": 1,
+}
+verify_a
```
eranrund, line 48: verify_a?

eranrund (approved):
LGTM once CI is green

PR #212, opened 2021-12-13
can now claim gift codes to fog enabled addresses

BrianCorbin:


full-service/src/service/gift_code.rs
```diff
@@ -131,6 +132,9 @@ pub enum GiftCodeServiceError {
     /// Error with Transaction Builder
     TxBuilder(mc_transaction_std::TxBuilderError),
 
+    /// Error parsing URI {0}
```
eranrund, line 135:
```suggestion
    /// Error parsing URI: {0}
```

full-service/src/service/gift_code.rs
```diff
@@ -556,8 +566,14 @@ where
...
-        let fog_resolver =
-            (self.fog_resolver_factory)(&[]).map_err(GiftCodeServiceError::UnexpectedTxStatus)?;
+        let fog_resolver = {
+            let fog_uris = core::slice::from_ref(&recipient_public_address)
+                .iter()
+                .filter_map(|x| extract_fog_uri(x).transpose())
+                .collect::<Result<Vec<_>, _>>()?;
+            (self.fog_resolver_factory)(&fog_uris.as_slice())
+                .map_err(GiftCodeServiceError::UnexpectedTxStatus)?
+        };
```
eranrund, line 576: I think this can be simplified to something like this since there is only a single address here:

```
let fog_resolver = {
    // This should convert the Option<String> into Option<Result<FogUri, Err>> and then transpose will convert to Result<Option<FogUri>, Err>. The ? operator will then return the error if there is one, and if not fog_uri will be Option<FogUri>
    let fog_uri = recipient_public_address.fog_report_url().map(FogUri::from_str).transpose().map_err(GiftCodeServiceError::InvaligFogUri)?;

    let uri_slice = if let Some(uri) = fog_uri {
        &[uri]
    } else {
        &[]
    };

    (self.fog_resolver_factory)(uri_slice).map_err(GiftCodeServiceError::FogPubkeyResolver)?;
}
```

You'll need to add error types to GiftCodeServiceError. UnexpectedTxStatus is not suitable for this.

eranrund (changes requested):

full-service/src/service/gift_code.rs
eranrund: Actually maybe this isn't any better, but the comment on the error variant is still relevant.
I believe `core::slice::from_ref(&x)` can be replaced with `&[&x]`.

itdaniher (approved):
YAY BRIAN!

full-service/src/service/gift_code.rs
BrianCorbin: ah, thanks for catching that, meant to change that after getting it working!

eranrund (approved):

PR #215, opened 2021-12-15
Feature/refactor txo status

BrianCorbin:
Soundtrack of this PR: [Ripping Me Apart](https://www.youtube.com/watch?v=omvfJOio2Uo&ab_channel=BetterNoiseMusic)

### Motivation

This was needed to simplify how txo's are associated with accounts, removing a significant amount of unnecessary code and simplifying the code base while slightly improving performance surrounding scanning txos for accounts and looking up txos by accounts/addresses. This also will enable us to both improve the account sync speed in a future PR and more easily support view only accounts.

### In this PR
* Removed AccountTxoStatus from the database and replaced it with 2 fields on the Txo table: minted_account_id_hex and received_account_id_hex, as these were the only 2 possible types of AccountTxoStatuses that could exist.
* We now derive the TXO_STATUS directly from the Txo itself instead of relying on the AccountTxoStatus field, as this information is all self contained within the Txo. 

### Future Work
* Changing the Api to reflect the move away from AccountTxoStatus and just return the Txo object. Since all of the information is self contained, the AccountTxoStatus field is redundant and only decreases performance and increases the amount of data per Txo being retrieved from the database.


eranrund:
Can you please edit the PR description to explain what this about?

brucek:
Bug:  link to appropriate soundtrack broken.  Could not review.

BrianCorbin:
Yeah, didn't mean to request review quite yet still in a draft. Gonna write up the description today! :)

full-service/src/db/txo.rs
```diff
@@ -537,146 +436,163 @@ impl TxoModel for Txo {
...
+    }
+
+    fn list_secreted(
+        account_id_hex: &str,
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<Vec<Txo>, WalletDbError> {
+        use crate::db::schema::txos;
+
+        let txos: Vec<Txo> = txos::table
+            .filter(txos::received_account_id_hex.ne(account_id_hex))
```
eranrund, line 542: From someone that understands very little of the full-service internals: Wouldn't this return all txos that were received by other accounts? what is the purpose of this? what does `secreted` means?

full-service/src/json_rpc/txo.rs
```diff
@@ -93,67 +102,64 @@ pub struct Txo {
...
-                .received_block_index
-                .map(|x| (x as u64).to_string()),
-            spent_block_index: txo_details
-                .txo
-                .spent_block_index
-                .map(|x| (x as u64).to_string()),
+            txo_id_hex: txo.txo_id_hex.clone(),
+            value_pmob: (txo.value as u64).to_string(),
+            recipient_address_id: None,
+            received_block_index: txo.received_block_index.map(|x| (x as u64).to_string()),
```
eranrund, line 150: Why do you need the cast?

full-service/src/service/balance.rs
```diff
@@ -175,34 +171,28 @@ where
...
-                    TXO_STATUS_SECRETED => secreted += txo.txo.value as u128,
-                    TXO_STATUS_ORPHANED => orphaned += txo.txo.value as u128,
-                    _ => {
-                        return Err(BalanceServiceError::UnexpectedAccountTxoStatus(
-                            status.txo_status,
-                        ))
-                    }
-                }
-            }
+            let orphaned: u128 = 0;
```
eranrund, line 176: So what is actually orphaned? Why is it hardcoded to zero?

full-service/src/service/receipt.rs
```diff
@@ -185,29 +185,34 @@ where
...
-            // us. (For to-self transactions). If the Txo was minted by us, this
-            // transaction is pending.
-            if status.txo_type == TXO_TYPE_MINTED && status.txo_status == TXO_STATUS_SECRETED {
-                return Ok((ReceiptTransactionStatus::TransactionPending, None));
+            // Return if the Txo from the receipt has a pending tombstone block index
+            if txo.pending_tombstone_block_index.is_some() {
+                return Ok((ReceiptTransactionStatus::TransactionPending, Some(txo)));
+            }
+
+            // We are reproducing a bug with this logic and it is not how the
```
eranrund, line 208: Can you please update the comment to explain the bug, current behavior and desired behavior?

full-service/src/service/txo.rs
```diff
@@ -123,19 +123,15 @@ where
...
             let txo_details = Txo::get(&txo_id.to_string(), &conn)?;
-            let received_to_account_txo_status = match txo_details.received_to_account {
-                Some(account_status) => Ok(account_status),
-                None => Err(TxoNotSpendableByAnyAccount(txo_details.txo.txo_id_hex)),
-            }?;
 
-            let account_id_hex = match received_to_account_txo_status.txo_status.as_str() {
-                TXO_STATUS_UNSPENT => Ok(received_to_account_txo_status.account_id_hex),
-                _ => Err(TxoNotSpendable(received_to_account_txo_status.txo_status)),
+            let account_id_hex = match txo_details.received_account_id_hex {
```
eranrund, line 132: I believe this can be written as `let account_id_hex = txo_details.received_account_id_hex.ok_or_else(|| TxoNotSpendableByAnyAccount(txo_details.txo_id_hex))?;`

eranrund (commented):
I like simplifying things but my lack of understanding of full-service's internals makes it hard for me to effectively review this. I do have some questions that might make it easier for me or others to go over this.
Thank you!

full-service/src/db/txo.rs
christian-oudard: Secreted means minted by this account, but not received by this account. This query was in fact wrong, good catch!

full-service/src/json_rpc/txo.rs
christian-oudard: The database stores these values as i64. This was not actually new code, just moved.

full-service/src/service/balance.rs
christian-oudard: Orphaned txos have no subaddress assigned, so none of these txos can be orphaned. Added a comment explaining this.

full-service/src/service/receipt.rs
christian-oudard: We are about to fix this in the next PR. We'll put a full writeup in that PR.

full-service/src/service/txo.rs
christian-oudard: Good catch.

christian-oudard (approved):

PR #217, opened 2022-01-05
Stub for validator service

wjuan-mob:
Soundtrack of this PR: [https://www.youtube.com/watch?v=l-epqIHe4w0]()

### Motivation

Adding initial stub for validator service in full service for visibility and to make PRs easier to review by breaking into smaller chunks. This is the first in a series of PRs for the Ledger Validator project intended to work with FullService. The full featured ledger validator will be in charge of maintaining the ledger and act as the portal for full service to fog and consensus.

### In this PR

Adding stubs for validator service and config. James noted that this initial PR should basically be a binary that parses the config that is to be used in the next set of PRs.

### Future Work
* Syncing and maintaining the ledger
* Update LedgerValidator to include api to allow services to get ledger from it.
* Update FullService to include a method of using the ledger from the LedgerValidator
* Update LedgerValidator to include api to allow services to send transactions down it.
* Update FullService to include a method of sending transactions to LedgerValidator
* Unit Tests


validator-service/api/build.rs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2018-2021 The MobileCoin Foundation
```
eranrund, line 1: Nit: Copyrights are now 2022

validator-service/service/Cargo.toml
```diff
@@ -0,0 +1,78 @@
...
+[lib]
+name = "validator_server"
+path = "src/lib.rs"
+
+[[bin]]
+name = "validator_main"
+path = "src/bin/main.rs"
+
+[dependencies]
+mc-full-service = { path = "../../full-service"}
```
eranrund, line 17: Nit: Sort by alphabetical order or separate this dependency from the rest with an empty line.

validator-service/service/Cargo.toml
```diff
@@ -0,0 +1,78 @@
...
+
+[lib]
+name = "validator_server"
+path = "src/lib.rs"
+
+[[bin]]
+name = "validator_main"
+path = "src/bin/main.rs"
+
+[dependencies]
```
eranrund, line 16: There are a lot of unused dependencies here. Can you please clean that up?

validator-service/service/src/config.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[structopt(long)]
+    pub ledger_db_bootstrap: Option<String>,
+
+    /// The location for the network.toml/json configuration file.
+    #[structopt(long = "network", parse(from_os_str))]
+    pub network_path: PathBuf,
+    
+    /// Offline mode.
+    #[structopt(long)]
+    pub offline: bool,
```
eranrund, line 29: Is there an offline mode in this service?

eranrund (changes requested):
This is a good first step!
Overall LGTM, but I would like to see the dependencies cleaned up.
Thank you!

eranrund:
It also appears CI is unhappy, so that needs to be addressed as well.

validator-service/service/src/config.rs
wjuan-mob: I was using this flag for testing. But in retrospect there's probably a better way to do that. Thank you.

eranrund:
Why was this closed?

wjuan-mob:
> Why was this closed?
The config/service etc ended up being the wrong approach. After discussion with Brian we are just going to be using full service directly to avoid code redundancy and just add the validator/proxy aspect as an option for full service, so I closed this.

PR #220, opened 2022-01-11
Excluding mobilecoin submodule from linter

wjuan-mob:
Soundtrack of this PR: [link to song that really fits the mood of this PR](https://www.youtube.com/watch?v=2Pzq5f6-PYo)

### Motivation

Running the linter in Fullservice fails because it tries to lint the mobilecoin submodule which needs to be run in an enclave. 

### In this PR
Excluding mobilecoin submodule from the linter

### DIFF
--- a/tools/lint.sh
+++ b/tools/lint.sh
@@ -8,7 +8,7 @@ if [[ ! -z "$1" ]]; then
     cd "$1"
 fi
 
-for toml in $(grep --exclude-dir cargo --exclude-dir rust-mbedtls --include=Cargo.toml -r . -e '\[workspace\]' | cut -d: -f1); do
+for toml in $(grep --exclude-dir cargo --exclude-dir rust-mbedtls --exclude-dir mobilecoin --include=Cargo.toml -r . -e '\[workspace\]' | cut -d: -f1); do
   pushd $(dirname $toml) >/dev/null
   echo "Linting in $PWD"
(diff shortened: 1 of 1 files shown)


eranrund:
Can you share the error you are seeing without this change?
Enclave code should be lintable unrelated to whether the host has an enclave or not.

wjuan-mob:
> Can you share the error you are seeing without this change? Enclave code should be lintable unrelated to whether the host has an enclave or not.

Yes. Both Brian and I tried running this linter and got an error of this form: 
image

jcape:
If you want to lint everything, you'll just need to provide a `CONSENSUS_ENCLAVE_CSS` file via environment, but I'm also OK to ignore `mobilecoin.git`, under the assumption that it is happening already in the mobilecoin repo.

wjuan-mob:
> If you want to lint everything, you'll just need to provide a `CONSENSUS_ENCLAVE_CSS` file via environment, but I'm also OK to ignore `mobilecoin.git`, under the assumption that it is happening already in the mobilecoin repo.

As per our chat, opened a new issue for this. I think this should still be helpful for people using full service in the meantime, especially as it might take a while for the submodule to catch the change after the issue is resolved.

jcape:
https://github.com/mobilecoinfoundation/mobilecoin/issues/1304

eranrund (approved):

PR #222, opened 2022-01-12
Basic Validator GRPC API

eranrund:
### Motivation

The validator service exposes its functionality over GRPC. As previously discussed, this includes endpoints for:
* Fetching blocks and their contents
* Getting last block info (for fee purposes)
* Proxying Tx proposal
* Proxying fog report resolution

### In this PR
* Stubbed out GRPC implemenations + code to start the GRPC server.

### Future Work
* Add ledger syncing
* Implement GRPC endpoints
* Probably more...


validator/api/proto/validator_api.proto
```diff
@@ -0,0 +1,28 @@
...
+syntax = "proto3";
+import "blockchain.proto";
+import "consensus_common.proto";
+import "external.proto";
+import "google/protobuf/empty.proto";
+import "report.proto";
+
+package validator_api;
+
+service ValidatorAPI {
```
eranrund, line 16: I believe these are the new APIs we will need to implement, in addition to BlockchainAPI.

validator/service/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "mc-validator-service"
+version = "1.0.0"
+authors = ["MobileCoin"]
+edition = "2018"
+license = "GPL-3.0"
+
+[[bin]]
```
zixu-zhao, line 8: btw is this intentional?

eranrund: What are you referring to? This follows a pattern we've done in other services where we place most code in a reusable library and separately from that provide a binary that puts it all together.

wjuan-mob (approved):

PR #224, opened 2022-01-13
Initial take on API endpoint implementation

eranrund:
This PR contains my initial implementation of the various RPC API endpoints. It is not tested, but I believe the basic functionality likely works. To test it I will need to start modifying full-service (or write unit tests) and I wanted to keep this PR as small as possible.

Unit tests are part of the plan, but for the sake of moving quickly and wanting to de-risk the approach we're taking here I am opting to postpone writing them until confirming this actually plays nicely with full-service.


validator/service/src/validator_api.rs
```diff
@@ -2,71 +2,263 @@
...
+
+        match tx_propose_result {
+            Ok(block_count) => {
+                result.set_block_count(block_count);
+                Ok(())
+            }
+
+            Err(RetryError::Operation { error, .. }) => {
+                match error {
+                    ConnectionError::TransactionValidation(err) => {
```
wjuan-mob, line 160: Is there a specific reason that the Transaction validation error has a different format and behaviour than the other 2?

wjuan-mob (approved):

validator/service/src/validator_api.rs
eranrund: Yes. This is because I am trying to preserve the semantics of the ClientTxPropose call. This API call can fail either by returning a `TransactionValidation` error which is a valid GRPC response encapsulated in https://github.com/mobilecoinfoundation/mobilecoin/blob/bdd5ded7aff9b8a86bd10c568a1f2bcf1ee20d27/consensus/api/proto/consensus_common.proto#L80, or the GRPC call itself can fail for various reasons (attestation issues, network issues, etc).
The plan is to try and implement a "ThinClient" like you started doing, and I want to be able to return proper error results from it as much as possible. This special treatment of `TransactionValidation`, as well as the `Cipher` and `Attestation` variants will hopefully do that. Relaying the correct error value and not coercing everything into a generic "internal error" or something of that nature is important for correct client behavior.

PR #233, opened 2022-01-18
Fix a bug in receipt status endpoint.

christian-oudard:
.


BrianCorbin (approved):

BrianCorbin:
I would like to get a secondary review from @davidyhall to make sure this is okay from a product perspective, or if you're at least indifferent.

PR #235, opened 2022-01-19
Ledger Validator Service

eranrund:
This PR introduces the Ledger Validator Service.
The LVN addresses some specific use case where a client needs to run `full-service` on a machine that can only access a single internal endpoint, and have that endpoint perform ledger validation and transaction relying on its behalf.

This is composed of a series of PRs that were each individually reviewed.


BrianCorbin (approved):
This looks fantastic! Played around with it a bunch, it all looks good to me :)

PR #236, opened 2022-01-21
LVN IP Check

eranrund:
### Motivation

When using the `--validator` option and LVN, the LVN should perform the IP checking and not `full-service`.

### In this PR
* Add IP checking to LVN
* Skip IP checking in `full-service` when running with `--validator`


eranrund:
Closing in favor of a PR @wjuan-mob will submit

PR #239, opened 2022-01-25
Make logging of validator-related failures more explicit

eranrund:
### Motivation

I ran into this log message today:
```
 "details": "Error building transaction: Error generating FogPubkeyResolver Error fetching fog reports for [Uri { url: Url { scheme: \"fog\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"fog.prod.mobilecoinww.com\")), port: None, path: \"/\", query: None, fragment: None }, host: \"fog.prod.mobilecoinww.com\", port: 443, use_tls: true, username: \"\", password: \"\", _scheme: PhantomData }]: GRPC: RpcFailure: 14-UNAVAILABLE failed to connect to all addresses"
```

Initially it appeared to me as if `fog.prod.mobilecoinww.com` was not responding, but in reality it was the validator node that was not responding. This confusion will make future debugging more difficult, so I am proposing to make some logging more verbose.

### In this PR
* Making some failure cases log more verbosely

### DIFF
--- a/validator/connection/src/lib.rs
+++ b/validator/connection/src/lib.rs
@@ -5,7 +5,7 @@
 mod error;
 
 use grpcio::{ChannelBuilder, EnvBuilder};
-use mc_common::logger::Logger;
+use mc_common::logger::{log, Logger};
 use mc_connection::{
     BlockInfo, BlockchainConnection, Connection, Error as ConnectionError,
     Result as ConnectionResult, UserTxConnection,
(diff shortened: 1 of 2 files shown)


sugargoat (approved):

validator/connection/src/lib.rs
```diff
@@ -57,6 +58,7 @@ impl ValidatorConnection {
             uri: uri.clone(),
             validator_api_client,
             blockchain_api_client,
+            logger,
```
wjuan-mob, line 61: Would it make sense to add context to this logger so that it says that it's for the validator rather than adding "validator" to all the logged warnings/errors?

eranrund: In my experience the context is not great for immediately communicating information to users, since its usually the message that gets looked at and not the custom fields.

wjuan-mob (approved):

PR #256, opened 2022-02-17
Initial action to build containers for tag pushes

MCrank:
Soundtrack of this PR: [Fantasy Ambient](https://www.youtube.com/watch?v=AohPsuy65hQ)

### Motivation

Currently there is no automated build of Docker images for Full-Service so updates lag behind the code.  This adds a github action to build and publish docker images when a tag is pushed

### In this PR
* Adds a github action workflow to build docker images for testnet and mainnet.
* Currently only builds images when a tag is pushed.
* Added the `-testnet` and `-mainnet` suffixes to the images for better clarification.
* Commented out a `tests` job placeholder for Mikey who is currently working on some stress tests.
  * Defined the `develop` and `feature/**` filters so when the tests are put in they should run for any push to those branches.  This is also just a placeholder and should be adjusted when the test action is created if need be for any additional branches (main?) 

### Future Work
* Would like to actually perform the build of the binaries as well in the future to apply more automation around CI/CD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
 
       - name: Build and Publish to DockerHub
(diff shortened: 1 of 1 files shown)


BrianCorbin (approved):

MCrank:
@BrianCorbin Just double checking some things and realized I put the wrong secret name.  Just pushed an update to fix that error.

PR #257, opened 2022-02-22
FOG Creds default to empty string if not provided from API

BrianCorbin:
Since we were defaulting to empty string when creating the account key, it makes it easier to deal with this earlier in the chain and simply turn null values for fog creds into empty strings at the RPC layer


Shramp (approved):
nice

PR #260, opened 2022-02-25
Feature/view only account

Shramp:
TODO: update gitbook

Add view-only-accounts. These accounts are created by importing a private-view-key. They can view transactions for that view key and that's pretty much it. They'll be useful in hot + cold wallet flows and for viewing proof of burn.

Add view-only account functionality:
- add new tables for `view_only_accounts` and `view_only_txos`
- add `ViewOnlyAccount`, and `ViewOnlyTxo` models
- add `ViewOnlyAccount`, and `ViewOnlyTxo` services
- add `ViewOnlyBalance` to balance service
- add `ViewOnlyAccount` , `ViewOnlySecrets`, and `ViewOnlyTxo` json conversions
- update `sync.rs` to also sync view-only-accounts in a separate loop
- add endpoints:
  - export_view_only_account_secrets
  - get_view_only_account
  - get_all_view_only_accounts
  - get_balance_for_view_only_account
  - get_txos_for_view_only_account
  - import_view_only_account
  - remove_view_only_account
  - update_view_only_account_name

### Motivation

< The motivation for the changes in this PR. "Currently we...", "This is needed because..." >

### In this PR
* < Additions, removals, fixes, features >

< Ticket status, e.g. "fixes #issue number" > 

### Future Work
* < Out of scope non-goals for this PR >
* < These should be links to tickets. If the tickets do not exist, make them. >


full-service/src/db/view_only_account.rs
```diff
@@ -0,0 +1,288 @@
...
+                "account not found".to_string(),
+            )),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    fn list_all(
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<Vec<ViewOnlyAccount>, WalletDbError> {
+        use crate::db::schema::view_only_accounts;
```
remoun, line 146: Optional/Equivalent: Add a top-level `use crate::db::schema` then these in-method declarations can be `use schema::...`

full-service/src/json_rpc/e2e.rs
```diff
@@ -3611,4 +3612,355 @@ mod e2e {
...
+        dispatch_expect_error(
+            &client,
+            body,
+            &logger,
+            json!({
+                "method": "get_view_only_account",
+                "error": json!({
+                    "code": -32603,
+                    "message": "InternalError",
+                    "data": json!({
```
remoun, line 3735: Nit: This nested `json!` is unnecessary, judging by other nested usages in this file

full-service/src/json_rpc/wallet.rs
```diff
@@ -428,6 +432,17 @@ where
                 account_secrets: AccountSecrets::try_from(&account).map_err(format_error)?,
             }
         }
+        JsonCommandRequest::export_view_only_account_secrets { account_id } => {
+            let account = service
+                .get_view_only_account(&account_id)
+                .map_err(format_error)?;
+            JsonCommandResponse::export_view_only_account_secrets {
+                view_only_account_secrets:
+                    json_rpc::view_only_account::ViewOnlyAccountSecrets::try_from(&account)
```
remoun, line 441: Optional/Equivalent: `(&account).try_into()?` or maybe even `account.try_into().map_err(format_error)?`

full-service/src/json_rpc/wallet.rs
```diff
@@ -806,12 +862,70 @@ where
...
             }
         }
+        JsonCommandRequest::get_txos_for_view_only_account {
+            account_id,
+            offset,
+            limit,
+        } => {
+            // TODO(cc) share logic between this and regular account get txos
+            let o = offset.parse::<i64>().map_err(format_error)?;
+            let l = limit.parse::<i64>().map_err(format_error)?;
```
remoun, line 872: IIUC these should use `format_invalid_request_error`

full-service/src/service/account.rs
```diff
@@ -43,6 +46,12 @@ pub enum AccountServiceError {
 
     /// Invalid BIP39 english mnemonic: {0}
     InvalidMnemonic(String),
+
+    /// Error decoding base64 {0}
```
remoun, line 50: Nit: Colon before error message
```suggestion
    /// Error decoding base64: {0}
```

full-service/src/service/account.rs
```diff
@@ -43,6 +46,12 @@ pub enum AccountServiceError {
...
+    Base64DecodeError(String),
+
+    /// Error decoding private view key {0}
```
remoun, line 53: Ditto
```suggestion
    /// Error decoding private view key: {0}
```

full-service/src/service/sync.rs
```diff
@@ -477,4 +598,66 @@ mod tests {
...
+    #[test_with_logger]
+    fn test_sync_view_only_account(logger: Logger) {
+        let mut rng: StdRng = SeedableRng::from_seed([20u8; 32]);
+
+        let view_private_key = RistrettoPrivate::from_random(&mut rng);
+        // let key =
+        //     hex::decode("
+        // 0a20928c29f916586c0fae22de17784b2b9ac573a1b1d75c2ba531838650ca0a5302")
+        //         .unwrap();
+        // let view_private_key = mc_util_serial::decode(&key).unwrap();
```
remoun, line 611: Nit: Remove dead code

full-service/src/service/sync.rs
```diff
@@ -477,4 +598,66 @@ mod tests {
...
+                o.clone(),
+                o.clone(),
+                o.clone(),
+                o.clone(),
+                o.clone(),
+                o.clone(),
+                o.clone(),
+                o.clone(),
+                o.clone(),
+            ],
```
remoun, line 641: Optional/Equivalent: Use [`Range.map`](https://doc.rust-lang.org/std/ops/struct.Range.html#method.map) to DRY these calls.
```suggestion
            &(0..15).map(|_| o.clone()).collect::<Vec<_>>(),
```

full-service/src/service/view_only_account.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        let view_private_key = RistrettoPrivate::from_random(&mut rng);
+        // let view_private_key = mc_util_serial::encode(&expected_key);
+
+        let account_id_hex = ViewOnlyAccountID::from(&view_private_key).to_string();
+        let name = "coins for cats";
+        let first_block_index = 25;
+
+        // test import
+
```
remoun, line 177: Nit: Remove empty line
```suggestion
```

full-service/src/db/view_only_account.rs
```diff
@@ -0,0 +1,288 @@
+// Copyright (c) 2020-2021 MobileCoin Inc.
```
remoun, line 1: Nit: Use `-2022` in new files

full-service/src/test_utils.rs
```diff
@@ -361,6 +367,51 @@ pub fn manually_sync_account(
...
+                            info.details(),
+                            info.column_name(),
+                            info.table_name(),
+                            info.hint(),
+                        );
+                        panic!("Could not manually sync account.");
+                    }
+                };
+                std::thread::sleep(Duration::from_millis(500));
+            }
```
remoun, line 403: Optional: Restrict this match further with a guard.
```suggestion
            Err(SyncError::Database(WalletDbError::Diesel(
                diesel::result::Error::DatabaseError(kind, info),
            ))) if info.message() == "database is locked" => {
                log::trace!(logger, "Database locked. Will retry"),
                std::thread::sleep(Duration::from_millis(500));
            },
```

remoun (commented):
LGTM overall

full-service/migrations/2022-03-09-221715_view_only_txos/down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+DROP TABLE IF EXISTS view_only_txos;
```
christian-oudard, line 2: Let's have both of these tables in a single migration.

full-service/src/json_rpc/wallet.rs
```diff
@@ -514,6 +539,28 @@ where
                 account_map,
             }
         }
+        JsonCommandRequest::get_all_view_only_accounts => {
+            // TODO(cc) how to refactor this so that view_only and regular accounts can
```
christian-oudard, line 543: I think that since we're only doing this twice, that we don't need to worry so much about refactoring. If we were doing this multiple times then we should have a helper function. Let's remove the TODO comment.

full-service/src/json_rpc/wallet.rs
```diff
@@ -806,12 +862,70 @@ where
                 txo_map,
             }
         }
+        JsonCommandRequest::get_txos_for_view_only_account {
+            account_id,
+            offset,
+            limit,
+        } => {
+            // TODO(cc) share logic between this and regular account get txos
```
christian-oudard, line 870: Let's not leave too many TODO comments.

christian-oudard (approved):
Approved pending addressing comments.

full-service/src/json_rpc/wallet.rs
Shramp: my plan was to address these before I merged this PR. They were TODOs for this PR and not TODOs to be left in the code

full-service/src/db/view_only_account.rs
Shramp: this makes sense, but do you know why Diesel suggests inlining this import at all? I don't see it mentioned in the documentation and it seems to work when not inlined.

full-service/src/json_rpc/wallet.rs
Shramp: Tried this and it works, but I'm curious why you prefer it. To me `try_from` seems to make more sense since it's explicitly stating that we're going to create this json_rpc `ViewOnlyAccountSecrets` from our `account` entity. Whereas `try_into` looks more ambiguous.

full-service/src/json_rpc/wallet.rs
Shramp: yes!

full-service/src/db/view_only_account.rs
```diff
@@ -0,0 +1,288 @@
...
+    r2d2::{ConnectionManager, PooledConnection},
+    RunQueryDsl,
+};
+use mc_crypto_digestible::{Digestible, MerlinTranscript};
+use mc_crypto_keys::{RistrettoPrivate, RistrettoPublic};
+use std::{fmt, str};
+
+pub const DEFAULT_FIRST_BLOCK_INDEX: u64 = 0;
+pub const ROOT_ENTROPY_KEY_DERIVATION_VERSION: u8 = 1;
+pub const MNEMONIC_KEY_DERIVATION_VERSION: u8 = 2;
```
david1alvarez, line 20: Unsure about the `DEFAULT_FIRST_BLOCK_INDEX` value, but the other two look like they should live in mobilecoin/transaction/core/src/constants.rs

full-service/src/db/view_only_account.rs
```diff
@@ -0,0 +1,288 @@
...
+    }
+}
+
+pub trait ViewOnlyAccountModel {
+    // insert new view-only-account in the db
+    fn create(
+        account_id_hex: &str,
+        view_private_key: &RistrettoPrivate,
+        first_block_index: i64,
+        import_block_index: i64,
```
david1alvarez, line 45:
```suggestion
        first_block_index: u64,
        import_block_index: u64,
```

full-service/src/db/view_only_account.rs
```diff
@@ -0,0 +1,288 @@
...
+        conn: &PooledConnection<ConnectionManager<SqliteConnection>>,
+    ) -> Result<(), WalletDbError>;
+}
+
+impl ViewOnlyAccountModel for ViewOnlyAccount {
+    fn create(
+        account_id_hex: &str,
+        view_private_key: &RistrettoPrivate,
+        first_block_index: i64,
+        import_block_index: i64,
```
david1alvarez, line 94:
```suggestion
        first_block_index: u64,
        import_block_index: u64,
```

full-service/src/db/models.rs
```diff
@@ -104,6 +105,44 @@ pub struct Account {
...
+#[primary_key(id)]
+pub struct ViewOnlyAccount {
+    /// Primary key
+    pub id: i32,
+    /// An additional ID, derived from the account data.
+    pub account_id_hex: String,
+    /// private key for viewing MobileCoin belonging to an account.
+    pub view_private_key: Vec<u8>,
+    /// Index of the first block where this account may have held funds.
+    pub first_block_index: i64,
```
david1alvarez, line 121: each of the `i64` references should be `u64` where negative values don't have meaning

full-service/src/db/models.rs
```diff
@@ -104,6 +105,44 @@ pub struct Account {
...
+}
+
+/// A structure that can be inserted to create a new entity in the
+/// `view_only_accounts` table.
+#[derive(Insertable)]
+#[table_name = "view_only_accounts"]
+pub struct NewViewOnlyAccount<'a> {
+    pub account_id_hex: &'a str,
+    pub view_private_key: &'a [u8],
+    pub first_block_index: i64,
```
david1alvarez, line 140: `i64`->`u64`

full-service/src/db/view_only_account.rs
```diff
@@ -0,0 +1,288 @@
...
+            view_only_accounts,
+        };
+        diesel::update(view_only_accounts.filter(dsl_account_id.eq(&self.account_id_hex)))
+            .set(dsl_next_block.eq(next_block_index))
+            .execute(conn)?;
+        Ok(())
+    }
+
+    fn delete(
+        self,
```
david1alvarez, line 184: maybe newb question, but why pass `self` here rather than `&self`?

full-service/src/db/view_only_txo.rs
```diff
@@ -0,0 +1,249 @@
+// Copyright (c) 2020-2022 MobileCoin Inc.
+
+//! DB impl for the Txo model.
```
david1alvarez, line 3:
```suggestion
//! DB impl for the view-only Txo model.
```

david1alvarez (commented):
Generally LGTM, though as I can't comment on the core functionality I'll hold off on officially approving/requesting on my end

full-service/src/db/view_only_account.rs
Shramp: I don't want to add these to the mobilecoin repo, I want to keep them in full-service but I agree they should be in a constants file somewhere. Will add that

full-service/src/db/models.rs
Shramp: I wasn't sure about this. I followed the example of all the other models in our schema, which all use i64. I agree that it would make more sense to use u64 but I think there is something with diesel that makes us use i64

full-service/src/db/view_only_account.rs
Shramp: using `self` here instead of `&self` because this is a delete method. I could be wrong, but I think it's unlikely that the entity will be used again after being deleted so it's okay for the method to take ownership

full-service/src/json_rpc/balance.rs
```diff
@@ -70,3 +70,47 @@ impl From<&service::balance::Balance> for Balance {
         }
     }
 }
+
+/// The "balance" for a view-only-account, as well as some information about
+/// syncing status needed to interpret the balance correctly. In order for the
+/// balance to be accurate, you must mark view_only_txos as spent
+#[derive(Deserialize, Serialize, Default, Debug, Clone)]
+pub struct ViewOnlyBalance {
```
BrianCorbin, line 78: So, this will eventually have a field to represent the calculated balance, yeah? Maybe that can wait until we get into implementing tracking spent txos.

full-service/src/json_rpc/wallet.rs
```diff
@@ -514,6 +541,26 @@ where
                 account_map,
             }
         }
+        JsonCommandRequest::get_all_view_only_accounts => {
```
BrianCorbin, line 544: We should order these alphabetically

BrianCorbin (changes requested):

full-service/src/json_rpc/balance.rs
Shramp: yeah, that's my plan

BrianCorbin (commented):
LGTM! Just gonna run this on my end a little bit before approving for merge :)

BrianCorbin (changes requested):
the `get_wallet_status` endpoint should probably be updated to include a list of `view_only_accounts` as well. Thoughts on this?

BrianCorbin (approved):

PR #265, opened 2022-03-10
Fix the handling of txos in the range of 2^63 to 2^64 pmob. Refactor usage of i64 and u64 values.

christian-oudard:
.


NotGyro (approved):

christian-oudard:
Ignoring flaky test.

PR #266, opened 2022-03-14
Deprecate get_all_* Pattern

david1alvarez:
### Motivation

Currently we have implemented a get_all_* pattern for a number of use cases, and we're looking to limit vendor reliance on this pattern.

### In this PR
- Fixes ticket https://app.asana.com/0/1200175610892874/1201824758436244/f
- Marks get_all_* pattern api endpoints as deprecated in the documentation, and passes an extra `"deprecated": true` field to the api response body in the results field.

### Future Work
Monitor the endpoints to see if they're still seeing high use by integrators.


christian-oudard (changes requested):
`get_all_accounts` should not be removed. Rejecting this change, until we figure out what to do here.

BrianCorbin:
> `get_all_accounts` should not be removed. Rejecting this change, until we figure out what to do here.
I think all "get_all" endpoints need to have an alternative paging version, at least. I agree with "get_all_accounts" being an exception for now, as this is most likely never going to be an issue.

Shramp:
thanks for taking this on! I don't think we should mark these as deprecated until we provide an alternative for fetching multiple entities (limit/offset).

christian-oudard:
Removing this, and handling it a different way in a new PR.

PR #270, opened 2022-03-16
Feature/lock txos during selection

BrianCorbin:
These changes cause TXOs to be set to pending as soon as they're selected in the build step, as opposed to when submitted. This is an optional flag in `build_transaction`, but not optional in `build_and_submit_transaction`. This will prevent concurrent requests to build transactions from selecting the same TXOs, resulting in one of the two (whichever was second to submission) failing.


full-service/src/db/txo.rs
```diff
@@ -733,114 +735,133 @@ impl TxoModel for Txo {
 
     fn select_by_id(
         txo_ids: &[String],
+        pending_tombstone_block_index: Option<u64>,
```
Shramp, line 738: I think it feels a bit weird to have a `select` method perform an update.

Shramp (approved):

PR #278, opened 2022-04-12
Bugfixes for offline account creation.

christian-oudard:
Bugs fixed:
* When creating an account, it sets the next block index equal to the
  current block, forcing the current block to re-sync unnecessarily.
* When creating an account offline, it continues to endlessly sync the
  account past the zero blocks that are known in the ledger.
* When creating an account offline with no ledger, the first block is
  set to negative one, which overflows to <u64>::MAX - 1

### DIFF
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -252,7 +252,11 @@ impl AccountModel for Account {
         use crate::db::schema::accounts;
 
         let account_id = AccountID::from(account_key);
+
+        // The next block for scanning should should start one after the current block, unless we
+        // are starting at the beginning of the ledger.
         let fb = first_block_index.unwrap_or(DEFAULT_FIRST_BLOCK_INDEX);
+        let nb = if fb == DEFAULT_FIRST_BLOCK_INDEX { 0 } else { fb + 1 };
 
         let change_subaddress_index = if fog_enabled {
             DEFAULT_SUBADDRESS_INDEX as i64
@@ -276,7 +280,7 @@ impl AccountModel for Account {
             change_subaddress_index,
             next_subaddress_index,
             first_block_index: fb as i64,
-            next_block_index: fb as i64,
+            next_block_index: nb as i64,
             import_block_index: import_block_index.map(|i| i as i64),
             name,
             fog_enabled,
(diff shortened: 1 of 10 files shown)


full-service/src/db/account.rs
```diff
@@ -252,7 +252,15 @@ impl AccountModel for Account {
         use crate::db::schema::accounts;
 
         let account_id = AccountID::from(account_key);
+
+        // The next block for scanning should should start one after the current block,
+        // unless we are starting at the beginning of the ledger.
```
Shramp, line 257: do you also need to make this change for `import_block_index`?

full-service/src/db/account.rs
```diff
@@ -252,7 +252,15 @@ impl AccountModel for Account {
...
         let fb = first_block_index.unwrap_or(DEFAULT_FIRST_BLOCK_INDEX);
+        let nb = if fb == DEFAULT_FIRST_BLOCK_INDEX {
+            0
```
Shramp, line 260: should this condition return `DEFAULT_FIRST_BLOCK_INDEX` instead of `0`?

full-service/src/test_utils.rs
```diff
@@ -658,21 +669,36 @@ pub fn get_resolver_factory(
...
+        setup_peer_manager_and_network_state(ledger_db.clone(), logger.clone(), offline);
 
     WalletService::new(
         wallet_db,
         ledger_db,
         peer_manager,
         network_state,
         get_resolver_factory(&mut rng).unwrap(),
-        false,
+        offline,
```
Shramp, line 701: oh, interesting. There was an 'offline' option already, but it didn't actually change anything?

full-service/src/db/account.rs
```diff
@@ -252,7 +252,9 @@ impl AccountModel for Account {
         use crate::db::schema::accounts;
 
         let account_id = AccountID::from(account_key);
-        let fb = first_block_index.unwrap_or(DEFAULT_FIRST_BLOCK_INDEX);
+
+        let first_block_index = first_block_index.unwrap_or(DEFAULT_FIRST_BLOCK_INDEX);
+        let next_block_index = first_block_index;
```
BrianCorbin, line 257: Just noting that this is not a functional change. We explicitly do this to show intent that we mean to set the variables this way.

BrianCorbin (approved):

PR #280, opened 2022-04-22
bugfix/db transactions

christian-oudard:
- Fix DB locking in assign_address_for_account.
- Fix database locking during account creation and deletion.
- Remove unnecessary transactions from account service functions.
- Redo usage of transactions for view-only accounts.
- Use a type alias for Diesel connections.
- Move transaction for select-unspent-txos into service layer
- Use new transaction fn in gift code service
- Move db transactions from db layer to service layer.
- Remove unnexessary transactions
- Remove or replace more instances of conn.transaction
- Finish converting transactions to new style.

### DIFF
--- a/cli/test/stress.py
+++ b/cli/test/stress.py
@@ -1,34 +1,59 @@
 import aiohttp
 import asyncio
-import time
+from time import perf_counter
 
 
 async def main():
-        c = StressClient()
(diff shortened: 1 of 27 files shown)


Shramp (approved):
LGTM!

PR #283, opened 2022-04-29
feature/limit offset optional

christian-oudard:
Remove endpoints duplicated due to pagination.
* Three endpoints had duplicate versions, one for pagination, and one for
no pagination. Enabled optional pagination for these endpoints.
* Use u64 to specify limit and offset, converting to i64 only in the DB
  layer.
* The new default page size (limit value) is 100.
* Added aliases to redirect deprecated method names.
* Updated tests and CLI for usage of these endpoints.
* Updated and cleaned up relevant documentation.


christian-oudard:
Unittest failure is due to a flaky test, so it's ignorable.

BrianCorbin (approved):

Issue #310, opened 2022-05-16
Transactions should always include Change TXO even if not needed


BrianCorbin:
To better support RTH, a change TXO should always be included in a transaction, even when the total change would be 0 MOB.

PR #323, opened 2022-05-23
Add change output for zero change

Shramp:
### Motivation

To better support RTH, a change TXO should always be included in a transaction, even when the total change would be 0 MOB.

### In this PR
* Add change output, regardless of change value

https://github.com/mobilecoinofficial/full-service/issues/310

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -795,7 +795,7 @@ mod tests {
     }
 
     #[test_with_logger]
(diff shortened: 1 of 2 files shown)


BrianCorbin (changes requested):
So, there is still the use case of having 16 output txos in the transaction, in which case there is no room for a change txo to exist. I think this is still a valid situation.

BrianCorbin (approved):

PR #328, opened 2022-05-24
Feature/vo account balance improvements

BrianCorbin:
Before the large view only account update, VO accounts had been using a simplified balance that simply gave the known spent vs unspent txos. This update brings the new format up to date with the VO account expansion to include all states of txos, as well as balances for subaddresses.


Shramp (approved):
Added tests and it all lGTM!

PR #330, opened 2022-05-25
Docs/fix new updates

Shramp:


BrianCorbin (changes requested):
Could you truncate the example response in build unsigned transaction? No need for all 12k lines to be there, can just use "..." to represent the main objects instead of their full structure.

BrianCorbin (approved):
perfect, thank you!

PR #362, opened 2022-06-08
Feature/tx logs since block

BrianCorbin:
This adds the ability for a user to scan for new transactions over a range of blocks


david1alvarez (approved):
LGTM! Thanks for adding tests around this too

PR #374, opened 2022-06-13
Feature/merge view only and normal db models

BrianCorbin:
### Tests
- [x] View Only Account Import
- [x] View Only Account Sync Process
- [x] VO Assigned Subaddress Creation


full-service/src/service/account.rs
```diff
@@ -53,6 +63,15 @@ pub enum AccountServiceError {
 
     /// Error decoding private view key: {0}
     DecodePrivateKeyError(String),
+
+    /// Error decoding keys
+    KeyError(mc_crypto_keys::KeyError),
+
+    /// Account is a view only account and shouldn't be
+    AccountIsViewOnly(AccountID),
```
Shramp, line 71: I think these two errors still need to be implimented?

full-service/src/util/encoding_helpers.rs
```diff
@@ -1,9 +1,13 @@
-use mc_crypto_keys::RistrettoPrivate;
+use mc_crypto_keys::{RistrettoPrivate, RistrettoPublic};
 
 pub fn ristretto_to_vec(key: &RistrettoPrivate) -> Vec<u8> {
     mc_util_serial::encode(key)
 }
 
+pub fn ristretto_public_to_vec(key: &RistrettoPublic) -> Vec<u8> {
```
Shramp, line 7: stoked to see that these helpers are being used!

Shramp (changes requested):
Diff is a bit large for deep understanding of everything happening here, but most of it makes sense to me and looks like. Just a few minor questions/requests

christian-oudard (approved):

full-service/migrations/2022-06-13-204000_api_v3/up.sql
```diff
@@ -2,14 +2,17 @@ CREATE TABLE accounts (
   id INTEGER NOT NULL PRIMARY KEY,
```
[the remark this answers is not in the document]
christian-oudard: This is starting the new 2.0 release series, and it seems prohibitively difficult at this time to do the data migrations necessary to untangle the separate view-only accounts. So, we decided to wipe the current migration table, and start afresh, requiring people to re-import.

Shramp (approved):
LGTM once merge conflicts are resolved

PR #383, opened 2022-06-28
Clean up e2e tests file

Shramp:
Split up e2e tests into modules that follow similar delineation as our docs


Shramp:
> 

yeah, good call. Just boxed the tests up into smaller boxes

Shramp:
> This is a good start, I think it could still use some more breaking up though, so that any 1 file isn't too gigantic to make it easier to navigate and find tests that already exist. Not sure what your thoughts are on this.

yeah, good call. Just separated the tests into smaller files

BrianCorbin (approved):
LGTM! Thank you for organizing this :)

PR #392, opened 2022-06-30
Refactor Txos

BrianCorbin:

### Motivation

The goal is to decrease the complexity of and normalize the database. This tackles the txos table, and how we determine the status of a given TXO.

### In this PR
* Refactored txos table
* Refactored transaction_txos table into transaction_input_txos and transaction_output_txos
* Refactored how to determine status of txos

### Future Work
* continued refactoring of tables and code in preparation of support multiple token ids

### DIFF
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -487,20 +487,21 @@ impl AccountModel for Account {
         }
     }
 
+    // TODO - UPDATE THIS
     fn get_by_txo_id(txo_id_hex: &str, conn: &Conn) -> Result<Vec<Account>, WalletDbError> {
         let txo = Txo::get(txo_id_hex, conn)?;
 
         let mut accounts: Vec<Account> = Vec::<Account>::new();
 
-        if let Some(received_account_id_hex) = txo.received_account_id_hex {
+        if let Some(received_account_id_hex) = txo.account_id_hex {
             let account = Account::get(&AccountID(received_account_id_hex), conn)?;
             accounts.push(account);
         }
 
-        if let Some(minted_account_id_hex) = txo.minted_account_id_hex {
-            let account = Account::get(&AccountID(minted_account_id_hex), conn)?;
-            accounts.push(account);
-        }
+        // if let Some(minted_account_id_hex) = txo.minted_account_id_hex {
+        //     let account = Account::get(&AccountID(minted_account_id_hex), conn)?;
+        //     accounts.push(account);
+        // }
 
         Ok(accounts)
     }
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -492,8 +492,8 @@ impl AccountModel for Account {
 
         let mut accounts: Vec<Account> = Vec::<Account>::new();
 
-        if let Some(received_account_id_hex) = txo.account_id_hex {
-            let account = Account::get(&AccountID(received_account_id_hex), conn)?;
+        if let Some(account_id_hex) = txo.account_id_hex {
+            let account = Account::get(&AccountID(account_id_hex), conn)?;
             accounts.push(account);
         }
 
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -487,7 +487,6 @@ impl AccountModel for Account {
         }
     }
 
-    // TODO - UPDATE THIS
     fn get_by_txo_id(txo_id_hex: &str, conn: &Conn) -> Result<Vec<Account>, WalletDbError> {
         let txo = Txo::get(txo_id_hex, conn)?;
 
@@ -497,11 +496,6 @@ impl AccountModel for Account {
             let account = Account::get(&AccountID(received_account_id_hex), conn)?;
             accounts.push(account);
         }
-
-        // if let Some(minted_account_id_hex) = txo.minted_account_id_hex {
-        //     let account = Account::get(&AccountID(minted_account_id_hex), conn)?;
-        //     accounts.push(account);
-        // }
 
         Ok(accounts)
     }
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -85,8 +85,8 @@ pub struct ValueMap(pub HashMap<TokenId, u64>);
 #[derive(Debug)]
 pub struct AssociatedTxos {
     pub inputs: Vec<Txo>,
-    pub outputs: Vec<Txo>,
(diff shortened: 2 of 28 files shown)


full-service/src/db/account.rs
```diff
@@ -492,16 +492,11 @@ impl AccountModel for Account {
 
         let mut accounts: Vec<Account> = Vec::<Account>::new();
 
-        if let Some(received_account_id_hex) = txo.received_account_id_hex {
+        if let Some(received_account_id_hex) = txo.account_id_hex {
```
Shramp, line 495: small nit: rename `received_account_id_hex` to just `account_id_hex`. If we're not differentiating between received and minted, I think that not using the word `received` will avoid confusion for future devs

full-service/src/db/txo.rs
```diff
@@ -639,23 +635,52 @@ impl TxoModel for Txo {
...
         token_id: Option<u64>,
         offset: Option<u64>,
         limit: Option<u64>,
         conn: &Conn,
     ) -> Result<Vec<Txo>, WalletDbError> {
-        use crate::db::schema::txos;
+        use crate::db::schema::{transaction_input_txos, transaction_logs, txos};
+
+        /*
+            SELECT * FROM txos
```
Shramp, line 651: Do we need to save this raw SQL comment?

full-service/src/db/transaction_log.rs
```diff
@@ -202,30 +238,33 @@ impl TransactionLogModel for TransactionLog {
...
-            .load(conn)?;
+        use crate::db::schema::{transaction_input_txos, transaction_output_txos, txos};
 
-        let inputs = txos::table
-            .inner_join(
-                transaction_inputs::table.on(txos::txo_id_hex.eq(transaction_inputs::txo_id_hex)),
-            )
-            .filter(transaction_inputs::transaction_log_id.eq(&self.id))
+        let inputs: Vec<Txo> = txos::table
+            .inner_join(transaction_input_txos::table)
```
Shramp, line 244: oh neat! you don't have to specify which columns to join on

full-service/src/db/txo.rs
```diff
@@ -639,23 +635,52 @@ impl TxoModel for Txo {
...
+            transaction_logs::id
+                .is_null()
+                .or(transaction_logs::failed.eq(true))
+                .or(transaction_logs::id
+                    .is_not_null()
+                    .and(transaction_logs::submitted_block_index.is_null())),
+        );
+
+        query = query.filter(txos::received_block_index.is_not_null());
+        query = query.filter(txos::key_image.is_not_null());
```
Shramp, line 685: I think you need to add a filter for `txos::subaddress_index.is_not_null()` here

full-service/src/db/txo.rs
```diff
@@ -15,22 +15,55 @@ use mc_transaction_core::{
...
+    Unspent,
+    Pending,
+    Spent,
+    Orphaned,
+}
+
+impl fmt::Display for TxoStatus {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            TxoStatus::Unverified => write!(f, "unverified"),
```
Shramp, line 46: maybe not necessary for this PR, but I think a commented description for each txo status would be super helpful. Things to include: high-level description of what this status means, DB conditions for this status

full-service/src/service/txo.rs
```diff
@@ -304,48 +297,48 @@ mod tests {
 
         // We should now have 3 txos - one pending, two minted (one of which will be
         // change)
-        let txos = service
-            .list_txos(&AccountID(alice.account_id_hex.clone()), None, None, None)
+        // let txos = service
```
Shramp, line 300: delete this comment?

full-service/src/service/txo.rs
```diff
@@ -304,48 +297,48 @@ mod tests {
...
-            .cloned()
-            .filter(|txo| txo.minted_account_id_hex.is_some())
-            .collect();
-        assert_eq!(minted.len(), 2);
-        let minted_value_set = HashSet::from_iter(minted.iter().map(|m| m.value as u64));
-        assert!(minted_value_set.contains(&(58 * MOB - Mob::MINIMUM_FEE)));
-        assert!(minted_value_set.contains(&(42 * MOB)));
+        assert_eq!(pending[0].0.value, 100000000000000);
+
+        // let minted: Vec<Txo> = txos
```
Shramp, line 324: delete this comment?

Shramp (changes requested):

full-service/src/db/txo.rs
BrianCorbin: that's guaranteed by the key image not being null.

full-service/src/db/txo.rs
BrianCorbin: will add! definitely useful and should be included :)

full-service/src/db/txo.rs
BrianCorbin: I think it's nice to have the raw SQL query for reference, as reading the diesel stuff can be a little confusing if you're not familiar with it. Especially with filter chaining, the rules for that are very particular and diesel does some weird stuff sometimes that's unexpected.

full-service/src/db/account.rs
BrianCorbin: Hmm, I think that's the case currently. The txo table has a single account reference called "account_id_hex"

full-service/src/db/txo.rs
Shramp: wait, then what was the point of this PR?

https://github.com/mobilecoinofficial/full-service/commit/f4f807ba7429521c8a3d021bf37a08eeb3a9a4b7

full-service/src/db/account.rs
BrianCorbin: ah I see! thanks, updated :)

full-service/src/db/txo.rs
BrianCorbin: It was checking the wrong thing, it needed to check for key_image, not subaddress_index.

BrianCorbin: checking key_image or both key_image and subaddress index results in the same thing, but checking just subaddress_index will result in both unspent AND unverified txos

Shramp: ahhhh gotchya. Thanks

Shramp (approved):

PR #405, opened 2022-07-14
Balances Report All Token Ids In Wallet

BrianCorbin:


full-service/src/service/balance.rs
```diff
@@ -67,21 +71,27 @@ impl From<LedgerServiceError> for BalanceServiceError {
...
     pub pending: u128,
     pub spent: u128,
     pub secreted: u128,
     pub orphaned: u128,
-    pub unverified: u128,
-    pub network_block_height: u64,
-    pub local_block_height: u64,
-    pub synced_blocks: u64,
-    pub max_spendable: u128,
+    // pub network_block_height: u64,
```
Shramp, line 91: looks like it's safe to delete these comments now

full-service/src/service/balance.rs
```diff
@@ -266,21 +262,13 @@ where
...
-            conn,
-        )?
-        .max_spendable_in_wallet;
-
+    ) -> Result<Balance, BalanceServiceError> {
         let unspent = sum_query_result(Txo::list_unspent(
             account_id_hex,
             assigned_subaddress_b58,
-            Some(0),
+            Some(*token_id),
```
Shramp, line 271: what is the `*` doing here? according to SO it de-references the value? Not sure what that means or why it's being used here

Shramp, line 271: why the de-referencing with `*` here?

Shramp (commented):
LGTM once commented code is deleted and tests pass! Stoked to get stablecoin into FS!

full-service/src/service/balance.rs
BrianCorbin: it breaks the wrapper object TokenID and turns it into a u64, which is what the consuming function needs

PR #408, opened 2022-07-18
storing and using shared secret instead of confirmation

briancorbin:

### Motivation

Swap to storing the shared_secret instead of the confirmation number in the database, as it's more useful and the confirmation number can be derived from it.

### In this PR
* Storing hex encoded bytes, not protobuf data


Shramp (approved):

iamalwaysuncomfortable:
I'd also second the notion that shared secret shouldn't be stored directly - it's a key member of a lot of the algebra used to derive  several crucial secrets and [some of those equations are reversible.](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/core/src/amount/mod.rs)

iamalwaysuncomfortable (commented):
I'd second this probably shouldn't be stored directly. The shared secret is used in a lot of the algbera to derive sensitive secrets and some of that algebra is reversible. [Reference](https://github.com/mobilecoinfoundation/mobilecoin/blob/master/transaction/core/src/amount/mod.rs).

PR #410, opened 2022-07-22
assigns new change subaddress on startup if missing

BrianCorbin:
fixes accounts that are migrating from 1.8 -> 1.9+ that are using the legacy change address by auto assigning the new change subaddress, which will collect any orphaned txos sent to that address by transactions made with 1.9+

### DIFF
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -25,6 +25,7 @@ use mc_account_keys::{
 };
 use mc_account_keys_slip10::Slip10Key;
 use mc_crypto_digestible::{Digestible, MerlinTranscript};
+use mc_ledger_db::LedgerDB;
 use std::fmt;
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
@@ -70,6 +71,7 @@ pub trait AccountModel {
         fog_report_url: String,
         fog_report_id: String,
         fog_authority_spki: String,
+        ledger_db: &LedgerDB,
         conn: &Conn,
     ) -> Result<(AccountID, String), WalletDbError>;
 
@@ -87,6 +89,7 @@ pub trait AccountModel {
         fog_report_url: String,
         fog_report_id: String,
         fog_authority_spki: String,
+        ledger_db: &LedgerDB,
         conn: &Conn,
     ) -> Result<(AccountID, String), WalletDbError>;
 
@@ -104,6 +107,7 @@ pub trait AccountModel {
         next_subaddress_index: Option<u64>,
         name: &str,
         fog_enabled: bool,
+        ledger_db: &LedgerDB,
         conn: &Conn,
     ) -> Result<(AccountID, String), WalletDbError>;
 
@@ -118,6 +122,7 @@ pub trait AccountModel {
         fog_report_url: String,
         fog_report_id: String,
         fog_authority_spki: String,
+        ledger_db: &LedgerDB,
         conn: &Conn,
     ) -> Result<Account, WalletDbError>;
 
@@ -132,6 +137,7 @@ pub trait AccountModel {
         fog_report_url: String,
         fog_report_id: String,
         fog_authority_spki: String,
+        ledger_db: &LedgerDB,
         conn: &Conn,
     ) -> Result<Account, WalletDbError>;
 
@@ -176,6 +182,7 @@ impl AccountModel for Account {
         fog_report_url: String,
         fog_report_id: String,
         fog_authority_spki: String,
+        ledger_db: &LedgerDB,
         conn: &Conn,
     ) -> Result<(AccountID, String), WalletDbError> {
         let fog_enabled = !fog_report_url.is_empty();
@@ -195,6 +202,7 @@ impl AccountModel for Account {
             next_subaddress_index,
             name,
             fog_enabled,
+            ledger_db,
             conn,
         )
     }
@@ -208,6 +216,7 @@ impl AccountModel for Account {
         fog_report_url: String,
         fog_report_id: String,
         fog_authority_spki: String,
+        ledger_db: &LedgerDB,
         conn: &Conn,
     ) -> Result<(AccountID, String), WalletDbError> {
         let fog_enabled = !fog_report_url.is_empty();
@@ -229,6 +238,7 @@ impl AccountModel for Account {
             next_subaddress_index,
             name,
             fog_enabled,
+            ledger_db,
             conn,
         )
     }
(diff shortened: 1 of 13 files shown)


full-service/src/db/account.rs
```diff
@@ -683,13 +715,13 @@ mod tests {
...
             account_id_hex: account_id_hex.to_string(),
             account_key: [
-                10, 34, 10, 32, 129, 223, 141, 215, 200, 104, 120, 117, 123, 154, 151, 210, 253,
-                23, 148, 151, 2, 18, 182, 100, 83, 138, 144, 99, 225, 74, 214, 14, 175, 68, 167, 4,
-                18, 34, 10, 32, 24, 98, 18, 92, 9, 50, 142, 184, 114, 99, 34, 125, 211, 54, 146,
-                33, 98, 71, 179, 56, 136, 67, 98, 97, 230, 228, 31, 194, 119, 169, 189, 8, 26, 17,
-                102, 111, 103, 47, 47, 115, 111, 109, 101, 46, 102, 111, 103, 46, 117, 114, 108,
-                42, 23, 13, 231, 226, 158, 43, 94, 151, 32, 17, 121, 169, 69, 56, 96, 46, 182, 26,
-                43, 138, 220, 146, 60, 162,
+                10, 34, 10, 32, 135, 37, 92, 17, 29, 203, 205, 96, 6, 72, 229, 193, 11, 153, 50, 3,
```
sugargoat, line 718: Curious the reason why this changed, given the rootID is from the same seed?

full-service/src/service/address.rs
```diff
@@ -189,6 +192,33 @@ where
...
+                CHANGE_SUBADDRESS_INDEX as i64,
+                &conn,
+            )
+            .is_err()
+            {
+                let account_key: AccountKey = mc_util_serial::decode(&account.account_key).unwrap();
+                AssignedSubaddress::create(
+                    &account_key,
+                    None,
+                    CHANGE_SUBADDRESS_INDEX,
```
sugargoat, line 212: Does this need to have the same `fog_enabled` check?

https://github.com/mobilecoinofficial/full-service/blob/develop/full-service/src/db/account.rs#L283

sugargoat: Side note - this is fragile, referencing magic number `1`:

We probably want to have this be `DEFAULT_SUBADDRESS_INDEX + 1`

full-service/src/db/assigned_subaddress.rs
```diff
@@ -116,6 +124,64 @@ impl AssignedSubaddressModel for AssignedSubaddress {
         diesel::insert_into(assigned_subaddresses::table)
             .values(&subaddress_entry)
             .execute(conn)?;
+
+        // Find and repair orphaned txos at this subaddress.
+        let orphaned_txos = Txo::list_orphaned(&account_id.to_string(), None, None, None, conn)?;
```
sugargoat, line 129: This should only be exercised if the account is fully synced - is that enforced on AssignedAddress::create?

full-service/src/db/assigned_subaddress.rs
```diff
@@ -116,6 +124,64 @@ impl AssignedSubaddressModel for AssignedSubaddress {
...
+
+        for orphaned_txo in orphaned_txos.iter() {
```
sugargoat, line 131: Given that this logic is also implemented at `create_next_for_account`, can we leverage that implementation, especially given that it handles `view_only` vs `view_and_spend`? I see that above we're entering the new subaddress into the DB directly. Maybe we should be calling `AssignedSubaddress::create_next_for_account` here instead?

full-service/src/db/account.rs
BrianCorbin: my understanding is that it's due to the use of rng to generate the mock ledger db that was added earlier in the test, effectively changing the resulting root id

full-service/src/service/address.rs
BrianCorbin: Fog enabled accounts use the reserved subaddresses as well

full-service/src/db/assigned_subaddress.rs
BrianCorbin: it doesn't need to be only on fully synced accounts. when this is called it will check any orphaned txos, and the following sync process will now be checking for the newly assigned subaddress

full-service/src/db/assigned_subaddress.rs
BrianCorbin: this logic will not be making it into 1.10, as there will be a different requirement for migrating existing accounts from <1.9 -> 1.10+.

full-service/src/service/address.rs
BrianCorbin: doesn't need the fog enabled check, since the mc library handles change txos. These subaddress still needs to be assigned on fog accounts to pick up the txos

full-service/src/db/account.rs
sugargoat: Got it, ty!

full-service/src/db/assigned_subaddress.rs
sugargoat: Ah got it - nice

full-service/src/db/assigned_subaddress.rs
sugargoat: Ok understood. Hot fix status :)

full-service/src/service/address.rs
sugargoat: Ok, I think I understand - the reason we don't need the fog enabled check is because on fog all change goes to subaddress 0 anyway?

sugargoat (approved):

full-service/src/service/address.rs
BrianCorbin: no they go to the change subaddress defined in the mc library, but the PublicAddress is built in such a way that the fog hint uses subaddress 0's info so that it appears in a query to Fog for txos, but lives on the correct change subaddress index

BrianCorbin: https://github.com/mobilecoinfoundation/mobilecoin/blob/302bf9f0c238c86ec517d1e75308c68f8e5135d8/transaction/std/src/transaction_builder.rs#L312

BrianCorbin: https://github.com/mobilecoinfoundation/mobilecoin/blob/302bf9f0c238c86ec517d1e75308c68f8e5135d8/transaction/std/src/transaction_builder.rs#L328

sugargoat: Great, ty!

PR #412, opened 2022-07-27
Feature/update account db schema

BrianCorbin:

### Motivation

Updating the account db schema because it's storing duplicate data.

### In this PR
* Removed next_subaddress_index from the account schema
* Removed change_subaddress_index from the account schema
* Removed main_subaddress_index from the account schema
* WalletStatus in v2 no long reports an account map, as that can be obtained via the "get_accounts" call and seems unnecessary here.


full-service/src/db/account.rs
```diff
@@ -585,6 +579,19 @@ impl AccountModel for Account {
...
         Ok(distinct_token_ids)
     }
+
+    fn next_subaddress_index(self, conn: &Conn) -> Result<u64, WalletDbError> {
+        use crate::db::schema::assigned_subaddresses;
+
+        let highest_subaddress_index: i64 = assigned_subaddresses::table
+            .filter(assigned_subaddresses::account_id.eq(&self.id))
+            .order_by(assigned_subaddresses::subaddress_index.desc())
+            .select(diesel::dsl::max(assigned_subaddresses::subaddress_index))
```
Shramp, line 589: do you need both of these select statements here?

Shramp (approved):
One question about a diesel query, otherwise lgtm

full-service/src/db/account.rs
BrianCorbin: yeah, otherwise it returns the whole object which I don't want. It's not supposed to work that way, but alas, diesel.

PR #425, opened 2022-08-15
Feature/burn token

BrianCorbin:

### Motivation

This was added to support burning tokens

### In this PR
* Added build_burn_transaction endpoint
* Added build_unsigned_burn_transaction endpoint

### To Do
* Further refactoring of transaction builder to separate it into build, sign, and submit so that there is one source of truth for the logic that builds all transactions, instead of the logic being duplicated and only slightly altered for each type of transaction.


full-service/src/db/txo.rs
```diff
@@ -1396,6 +1396,7 @@ mod tests {
     use mc_fog_report_validation::MockFogPubkeyResolver;
     use mc_ledger_db::Ledger;
     use mc_transaction_core::{tokens::Mob, Amount, Token, TokenId};
+    use mc_transaction_std::EmptyMemoBuilder;
```
eranrund, line 1399: Where is this used?

full-service/src/json_rpc/v2/api/request.rs
```diff
@@ -53,6 +53,16 @@ pub enum JsonCommandRequest {
         max_spendable_value: Option<String>,
         comment: Option<String>,
     },
+    build_burn_transaction {
+        account_id: String,
+        amount: Amount,
+        redemption_memo: Option<String>,
```
eranrund, line 59: Suggest to rename to `redemption_memo_hex` to indicate its a hex string not a user-readable string.

full-service/src/json_rpc/v2/api/request.rs
```diff
@@ -53,6 +53,16 @@ pub enum JsonCommandRequest {
...
+        input_txo_ids: Option<Vec<String>>,
+        fee_value: Option<String>,
+        fee_token_id: Option<String>,
```
eranrund, line 62: Is this for a world where the fee is different than the token we are transacting in?

full-service/src/service/burn_transaction.rs
```diff
@@ -0,0 +1,337 @@
+// Copyright (c) 2020-2021 MobileCoin Inc.
+
+//! Service for managing gift codes.
+//!
+//! Gift codes are onetime accounts that contain a single Txo. They provide
```
eranrund, line 5: Gift codes?

full-service/src/service/burn_transaction.rs
```diff
@@ -0,0 +1,337 @@
+// Copyright (c) 2020-2021 MobileCoin Inc.
```
eranrund, line 1:
```suggestion
// Copyright (c) 2020-2022 MobileCoin Inc.
```

full-service/src/service/burn_transaction.rs
```diff
@@ -0,0 +1,337 @@
...
+                    None
+                };
+                builder.select_txos(&conn, max_spendable)?;
+            }
+
+            let mut memo_data = [0; BurnRedemptionMemo::MEMO_DATA_LEN];
+
+            if let Some(redemption_memo) = redemption_memo {
+                hex::decode_to_slice(&redemption_memo, &mut memo_data)
+                    .map_err(|_| BurnTransactionServiceError::InvalidNumberOfBytes)?;
```
eranrund, line 248: This error is confusing since you will also get it if the string is not a valid hex string.

full-service/src/service/burn_transaction.rs
```diff
@@ -0,0 +1,337 @@
...
+            }
+
+            let mut memo_builder = BurnRedemptionMemoBuilder::new(memo_data);
+            memo_builder.enable_destination_memo();
+            let tx_proposal: TxProposal =
+                builder.build(Some(Box::new(memo_builder)), &conn).unwrap();
```
eranrund, line 254: unwrap shouldn't be used here.

full-service/src/service/burn_transaction.rs
```diff
@@ -0,0 +1,337 @@
...
+        redemption_memo: Option<String>,
+        input_txo_ids: Option<&Vec<String>>,
+        fee_value: Option<String>,
+        fee_token_id: Option<String>,
+        tombstone_block: Option<String>,
+        max_spendable_value: Option<String>,
+    ) -> Result<(UnsignedTx, FullServiceFogResolver), BurnTransactionServiceError> {
+        let conn = self.wallet_db.get_conn()?;
+        transaction(&conn, || {
+            let mut builder = WalletTransactionBuilder::new(
```
eranrund, line 275: There seem to be a lot of duplication in the signed and unsigned flows. It is possible to unify them into a method that returns the `WalletTransactionBuilder` and then just the last part of either `.build()` or `.build_unsigned` could be different?

full-service/src/service/transaction_builder.rs
```diff
@@ -268,7 +268,10 @@ impl<FPR: FogPubkeyResolver + 'static> WalletTransactionBuilder<FPR> {
         Ok(FullServiceFogResolver(fully_validated_fog_pubkeys))
     }
 
-    pub fn build_unsigned(&self) -> Result<UnsignedTx, WalletTransactionBuilderError> {
+    pub fn build_unsigned(
+        &self,
+        redemption_memo: Option<[u8; BurnRedemptionMemo::MEMO_DATA_LEN]>,
```
eranrund, line 273: To future-proof this for other memos it might be worth making this an enum that currently only has one variant (the redemption memo). Otherwise for the next memo we want to use in an unsigned transaction another argument will need to be added, as well as another field to `UnsignedTx` - which then also makes it possible to provide two memos which is an error case.

eranrund (changes requested):
Nice, thank you!

full-service/src/json_rpc/v2/api/request.rs
BrianCorbin: In the eventual future of mixed token transactions, yes. At this exact moment, no.

full-service/src/service/burn_transaction.rs
BrianCorbin: Oops, copy pasta.

full-service/src/service/transaction_builder.rs
BrianCorbin: Ah, good point. Will implement something like that. Thanks!

full-service/src/service/burn_transaction.rs
BrianCorbin: Leftover from implementation, thanks!

full-service/src/service/burn_transaction.rs
BrianCorbin: Hmmm... Unfortunately the FromHexError doesn't give anything valuable, it seems to just be a catchall. I guess that'll be good enough, or maybe check the size first?

eranrund: Could check size first as it should be 128 characters

full-service/src/service/gift_code.rs
```diff
@@ -54,6 +54,8 @@ use rand::Rng;
 use serde::{Deserialize, Serialize};
 use std::{convert::TryFrom, fmt, iter::empty, str::FromStr, sync::atomic::Ordering};
 
+use super::transaction::TransactionMemo;
```
eranrund, line 57: Nit: Shouldn't be separated from other `use` statements

full-service/src/service/transaction.rs
```diff
@@ -147,6 +155,22 @@ impl From<mc_ledger_db::Error> for TransactionServiceError {
...
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub enum TransactionMemo {
+    /// Recoverable Transaction History memo.
+    RTH,
+
+    /// Burn Redemption memo, with an optional 64 byte redemption memo hex
+    /// string.
+    BurnRedemption(Option<String>),
```
eranrund, line 171: Why is the memo optional?
Separately, I think it should be `BurnRedemption([u8; BurnRedemptionMemo::MEMO_DATA_LEN])`, we should avoid passing invalid/unparseable memos as early as possible.

full-service/src/service/transaction.rs
```diff
@@ -351,7 +383,26 @@ where
...
             }
 
-            let tx_proposal: TxProposal = builder.build(&conn)?;
+            let tx_proposal = match memo {
+                TransactionMemo::RTH => {
+                    let mut memo_builder = RTHMemoBuilder::default();
+                    memo_builder
+                        .set_sender_credential(SenderMemoCredential::from(&from_account_key));
+                    memo_builder.enable_destination_memo();
+                    builder.build(Some(Box::new(memo_builder)), &conn)?
```
eranrund, line 392: This can be moved out of the match statement, the match statement can construct the boxed memo builder.

full-service/src/service/transaction_builder.rs
```diff
@@ -40,14 +40,15 @@ use mc_transaction_core::{
...
-    InputCredentials, RTHMemoBuilder, ReservedSubaddresses, SenderMemoCredential,
-    TransactionBuilder,
+    EmptyMemoBuilder, InputCredentials, MemoBuilder, ReservedSubaddresses, TransactionBuilder,
 };
 use mc_util_uri::FogUri;
 
 use rand::Rng;
 use std::{collections::BTreeMap, convert::TryFrom, str::FromStr, sync::Arc};
 
+use super::transaction::TransactionMemo;
```
eranrund, line 50: Nit: all imports should be grouped together

full-service/src/unsigned_tx.rs
```diff
@@ -53,13 +59,29 @@ impl UnsignedTx {
...
         // Create transaction builder.
-        let mut memo_builder = RTHMemoBuilder::default();
-        memo_builder.set_sender_credential(SenderMemoCredential::from(account_key));
-        memo_builder.enable_destination_memo();
         let fee = Amount::new(self.fee, TokenId::from(self.fee_token_id));
-        let mut transaction_builder =
-            TransactionBuilder::new(self.block_version, fee, fog_resolver, memo_builder)?;
+
+        let mut transaction_builder = match self.memo {
+            TransactionMemo::RTH => {
```
eranrund, line 67: Looks like this is duplicated from the signed tx flow.
Can this become a method on the `TransactionMemo` type? e.g. `fn get_boxed_memo_builder(&self, account_key: &AccountKey) -> Box<dyn ...>`?

eranrund (changes requested):

full-service/src/unsigned_tx.rs
BrianCorbin: Yeah, I have a separate PR coming to refactor the transaction building process into build and sign which everything will work off of

full-service/src/service/transaction.rs
```diff
@@ -91,6 +98,12 @@ pub enum TransactionServiceError {
 
     /// No default fee found for token id: {0}
     DefaultFeeNotFoundForToken(TokenId),
+
+    /// Error decoding hex string: {0}
+    FromHexError(hex::FromHexError),
+
+    /// Burn Redemption Memo must be exactly 128 characters (64 bytes) long.
```
eranrund, line 105: The String argument is not being rendered in the error

full-service/src/service/transaction.rs
```diff
@@ -290,12 +323,17 @@ where
...
+        memo: TransactionMemo,
     ) -> Result<TxProposal, TransactionServiceError> {
         validate_number_inputs(input_txo_ids.unwrap_or(&Vec::new()).len() as u64)?;
         validate_number_outputs(addresses_and_amounts.len() as u64)?;
 
         let conn = self.wallet_db.get_conn()?;
         transaction(&conn, || {
+            let account = Account::get(&AccountID(account_id_hex.to_string()), &conn)?;
+            let from_account_key: AccountKey =
+                mc_util_serial::decode(&account.account_key).unwrap();
```
eranrund, line 335: No unwraps in non-test code

full-service/src/service/transaction.rs
```diff
@@ -351,7 +389,22 @@ where
                 builder.select_txos(&conn, max_spendable)?;
             }
 
-            let tx_proposal: TxProposal = builder.build(&conn)?;
+            let memo_builder: Box<dyn MemoBuilder + Send + Sync> = match memo {
```
eranrund, line 392: This should be a method in `TransactionMemo`, is that doable?

eranrund (changes requested):

full-service/src/service/transaction.rs
BrianCorbin: hmm, yeah I don't see why not. I'll tinker around with that.

BrianCorbin: Rust is cool. Thanks for prompting me on that, I keep forgetting I'm not writing in Swift.

full-service/src/service/transaction.rs
```diff
@@ -91,6 +98,15 @@ pub enum TransactionServiceError {
...
     /// No default fee found for token id: {0}
     DefaultFeeNotFoundForToken(TokenId),
+
+    /// Error decoding hex string
+    FromHexError(hex::FromHexError),
+
+    /// Burn Redemption Memo must be exactly 128 characters (64 bytes) long.
+    InvalidBurnRedemptionMemo(String),
+
+    /// Error decoding with mc_util_serial
```
eranrund, line 108:
```suggestion
    /// mc_util_serial decode error: {0}
```

eranrund (changes requested):
Small nits but after that LGTM. Thank you!

eranrund (approved):

PR #426, opened 2022-08-16
Refactor Transaction Building

BrianCorbin:
In an effort to simplify and centralize the logic around building and signing transactions, this moves all duplicate/similar logic into a single source of truth

### DIFF
--- a/full-service/src/service/txo.rs
+++ b/full-service/src/service/txo.rs
@@ -245,7 +245,7 @@ where
             TransactionMemo::RTH,
         )?;
 
-        let account = Account::get(&AccountID(account_id_hex.clone()), &conn)?;
+        let account = Account::get(&AccountID(account_id_hex), &conn)?;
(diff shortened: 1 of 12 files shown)


full-service/src/service/txo.rs
```diff
@@ -213,17 +234,21 @@ where
...
             None,
-            None,
             TransactionMemo::RTH,
-        )?)
+        )?;
+
+        let account = Account::get(&AccountID(account_id_hex.clone()), &conn)?;
+        let account_key: AccountKey = mc_util_serial::decode(&account.account_key)?;
+
+        Ok(unsigned_tx.sign(&account_key, fog_resolver)?)
```
eranrund, line 251: I've seen these few lines of getting an ccount, decoding the key and then signing repeated a few times. Might be worth refactoring into a helper method.

eranrund (approved):

PR #433, opened 2022-09-02
Feature/import from fs

briancorbin:
This allows users to migrate data from 1.9.4 -> 1.10 by setting the --import-uri flag when starting full service.


tools/run-testnet.sh
```diff
@@ -19,4 +19,5 @@ ${WORK_DIR}/full-service \
     --peer mc://node2.test.mobilecoin.com/ \
     --tx-source-url https://s3-us-west-1.amazonaws.com/mobilecoin.chain/node1.test.mobilecoin.com/ \
     --tx-source-url https://s3-us-west-1.amazonaws.com/mobilecoin.chain/node2.test.mobilecoin.com/ \
-    --fog-ingest-enclave-css ${WORK_DIR}/ingest-enclave.css
+    --fog-ingest-enclave-css ${WORK_DIR}/ingest-enclave.css \
+    --import-uri http://localhost:9091
```
Shramp, line 23: I think we don't want the line for `import uri` committed to this script?

full-service/src/util/import_from_v1.rs
```diff
@@ -0,0 +1,227 @@
...
+        .unwrap();
+    } else {
+        panic!(
+            "No entropy or mnemonic found for account {}",
+            account.account_id
+        );
+    };
+
+    let body = json!({
+        "method": "get_sent_transaction_logs_for_account",
```
Shramp, line 118: I don't see `get_sent_transaction_logs_for_account` anywhere else in the code. In this version of FS or in V1.9.1

BrianCorbin: it's in v1.9.4+

Shramp (approved):
This lgtm, provided we don't have to also import gift codes. One thing that tripped me up running it was providing `localhost:9091` as the import_uri argument when I need to provide `localhost:9091/wallet`

Issue #339, opened 2022-05-31
Create cross version test suite


briancorbin:
The general idea is to have a github action that can

1. Select version A of Full Service
2. build and deploy a mc network with the version selected in FullServiceA
3. build and run FullServiceA targeting that network
4. perform a predefined set of setup commands to grow an account to a reasonable size (one that might be expected on an exchange) to include all forms of txos
5. Select version B of Full Service
6. build and deploy a mc network with the newly selected FullServiceB that uses the same ledger with the accounts created on the previous FullServiceA
7. build and run FullServiceB
8. check to make sure the state of the wallet_db is correct, indicating that migrations have worked across these 2 versions.

briancorbin:
Duplicate and broken up into smaller tasks

PR #435, opened 2022-09-13
Preliminary test infra for e2e tests

pouneh:
### Motivation
Currently it's difficult to build e2e tests for full-service to include in CI-CD. Here are some preliminary steps for building an e2e test suite. Specifically, this is an sample test that deploys a local network, and runs some basic full service stuff. 

## Next steps
- Ensure workflow_dispatcher works after the workflow file that enables it is in the default (develop) branch
- Make a test that follows the upgrade flow described in issue #339. 
- Refactor local_network.py to how we would like to use it, and then send that as a PR over to the mobilecoin repo
- Leverage the copy of local_network.py in the mobilecoin repo so that we a) don't duplicate all that code and b) can pick up any changes that might affect us


brucek (approved):
Approving based on needing to merge into main to get the manual workflow launch button to show up in GHA as per [this thread](https://github.com/orgs/community/discussions/25219)

Didn't do a deep parse but this looks like it is pretty independent / low risk at first look 🤞

cxloe (approved):
lgtm!

tools/local_network.py
```diff
@@ -0,0 +1,534 @@
+#!/usr/bin/python3
```
eranrund, line 1: I don't think it's desirable to copy this file here. It is available under https://github.com/mobilecoinfoundation/mobilecoin/tree/33af560693f1d1d6a10f77863cd617b50c10dcd5/tools/local-network 
Having a copy here would create unnecessary maintenance burden. Is it possible to use the one in the `mobilecoin` submodule?

eranrund (changes requested):

tools/build_full_service.py
```diff
@@ -0,0 +1,23 @@
...
+
+PROJECT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+MOBILECOIN_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'mobilecoin'))
+CARGO_FLAGS = '--release'
+LEDGER_BASE = os.path.join(PROJECT_DIR, 'target', "sample_data", "ledger")
+
+print('building full service...')
+cmd=' '.join([
+  'RUST_LOG=debug'
+  'MC_SEED=a4aa76e4a5ca70c8447dd544a63f180b5a6fe0aff96495802506354c10f2886e',
```
eranrund, line 15: MC_SEED does not affect anything `cargo build` is doing - I suspect it can be removed.

tools/build_full_service.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/python3
```
eranrund, line 1: Any reason this is a python script and not a shell script?

tools/build_mobilecoin.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/python3
```
eranrund, line 1: No reason for this to be a python script, it can be a shell script.

tools/mc_ledger_test.py
```diff
@@ -0,0 +1,76 @@
+import local_network
```
eranrund, line 1: Nit: missing copyright notice

tools/mc_ledger_test.py
```diff
@@ -0,0 +1,76 @@
...
+        local_network.cleanup(1)
+    elif len(fs.account_ids) < 2:
+        print(f'found {len(fs.account_ids)} account(s), minimum required is 2')
+        local_network.cleanup(1)
+    account_0 = fs.account_map[fs.account_ids[0]]
+    account_1 = fs.account_map[fs.account_ids[1]]
+    p_mob_amount = str(600_000_000)
+
+    # flakey tests due to accounts having a variable amount of pmob. This needs to be controlled for use.
+    log_0 = fs.send_transaction(account_0['account_id'], account_1['main_address'], p_mob_amount)
```
eranrund, line 22: Potential improvement: A balance check before the test could make it easier to understand why the test is failing, if it fails

tools/mc_ledger_test.py
```diff
@@ -0,0 +1,76 @@
...
+
+if __name__ == '__main__':
+    # TODO: This test can probably just always use the same network argument so we don't need it from command line
+
+    args = parse_network_type_cmd_line_args()
+
+    # start networks
+    print('________________________________________________________________________________')
+    print('Starting networks')
+    mobilecoin_network = None
```
eranrund, line 38: No need for this

tools/start_mobilecoin_ledger.sh
```diff
@@ -0,0 +1,87 @@
+# SGX_SPID=9E933DB667EDDE8454258EF3CDC6C2BC
```
eranrund, line 1: Shouldn't be left here

tools/start_mobilecoin_ledger.sh
```diff
@@ -0,0 +1,87 @@
...
+docker_command+="--expose 4444 "
+docker_command+="--expose 3200 "
+docker_command+="--expose 3201 "
+docker_command+="--expose 3202 "
+docker_command+="--publish 3200:3200 "
+docker_command+="--publish 3201:3201 "
+docker_command+="--publish 3202:3202 "
+docker_command+="--cap-add SYS_PTRACE "
+docker_command+="--name mobilecoin "
+docker_command+="-c 'cd tools/local-network && ls && ./bootstrap.sh && export MC_SEED=a4aa76e4a5ca70c8447dd544a63f180b5a6fe0aff96495802506354c10f2886e && export LEDGER_BASE=/tmp/mobilenode/target/sample_data/ledger && export IAS_API_KEY={{ secrets.SGX_PRIMARY_API_KEY }} && export IAS_SPID={{ secrets.SGX_SPID }} && python3 local_network.py --network-type a-b-c'"
```
eranrund, line 86: What are secrets.SGX_PRIMARY_API_KEY/secrets.SGX_SPID? What is replacing them with something? I don't recognize this syntax in a shell script so I am wondering how this gets used. Is this some GHA thing?

tools/test_utils/constants.py
```diff
@@ -0,0 +1,30 @@
...
+
+if not MOB_RELEASE or MOB_RELEASE == '0':
+    TARGET_DIR = 'target/debug'
+
+# Sane default log configuration
+if 'MC_LOG' not in os.environ:
+    os.environ[
+        'MC_LOG'] = 'debug,rustls=warn,hyper=warn,tokio_reactor=warn,mio=warn,want=warn,rusoto_core=error,h2=error,reqwest=error,rocket=error,<unknown>=error'
+if 'FS_LOG' not in os.environ:
+    os.environ['FS_LOG'] = 'info'
```
eranrund, line 30: what is FS_LOG? I haven't seen that before

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,215 @@
+# TODO: This should actually be more generic so that the python CLI
```
eranrund, line 1: Nit: copyright notice

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,215 @@
+# TODO: This should actually be more generic so that the python CLI 
+#   can also use it as a library (or maybe tests will use the CLI's library)
+from . import constants
```
eranrund, line 3: Nit: sort imports, group `import` and `from`.

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,215 @@
...
+                shutil.rmtree(tmpdir/'wallet-db')
+                shutil.rmtree(tmpdir/'ledger-db')
+            except Exception as e:
+                print(e) 
+        
+
+
+    def start(self):
+        cmd = ' '.join([
+            'mkdir -p /tmp/wallet-db/',
```
eranrund, line 42: This is hacky. Python has a way of creating a directory, no need to do it using a shell command.

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,215 @@
...
+            f'--tx-source-url file://{constants.MOBILECOIN_DIR}/target/release/mc-local-network/node-ledger-distribution-1',
+        ])
+        print('===================================================')
+        print('starting full service')
+        print(cmd)
+        self.full_service_process = subprocess.Popen(cmd, shell=True)
+
+    def stop(self):
+        try:
+            subprocess.check_output("pkill full-service", shell=True)
```
eranrund, line 58: `self.full_service_process.terminate()` should do the job

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,215 @@
...
+
+        print(f'request data: {request_data}')
+
+        try:
+            parsed_url = urlparse(url)
+            connection = http.client.HTTPConnection(parsed_url.netloc)
+            connection.request('POST', parsed_url.path, json.dumps(request_data), {'Content-Type': 'application/json'})
+            r = connection.getresponse()
+        except ConnectionError:
+            raise ConnectionError
```
eranrund, line 83: What purpose does this serve?

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,215 @@
...
+
+        raw_response = None
```
eranrund, line 85: Is this needed?

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,215 @@
...
+        try:
+            raw_response = r.read()
+            response_data = json.loads(raw_response)
+        except ValueError:
+            raise ValueError('API returned invalid JSON:', raw_response)
+        print(f'request returned {response_data}')
+
+        # TODO requests might be flakey due to timing issues... waiting 2 seconds to bypass most of these issues
+        time.sleep(2)
```
eranrund, line 94: I don't think it makes sense to always sleep 2 seconds after sending an HTTP request. What does that get you here?

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,215 @@
...
+        r = self._request({
+            "method": "import_account",
+            "params": params
+        })
+
+        if 'error' in r:
+            if 'Diesel Error: UNIQUE constraint failed' in r['error']['data']['details']:
+                return True
+            else:
+                return False
```
eranrund, line 115: This can be shortened and be documented.
```suggestion
            # If we failed due to a unique constraint, it means the account already exists
            return 'Diesel Error: UNIQUE constraint failed' in r['error']['data']['details']

File: tools/test_utils/fullservice.py [hunk 19]
```diff
@@ -0,0 +1,215 @@
...
+        keydata_0 = json.load(keyfile_0)
+        keydata_1 = json.load(keyfile_1)
+
+        keyfile_0.close()
+        keyfile_1.close()
+
+        return (keydata_0['mnemonic'], keydata_1['mnemonic'])
+
+    # check if full service is synced within margin
+    def sync_status(self, eps=5) -> bool:
```
eranrund, line 132: what is eps?

eranrund (changes requested):

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,121 @@
+name: mc-ledger-test
+
+env:
+  SGX_MODE: SW
+  IAS_MODE: DEV
+  RUST_BACKTRACE: full
+  MC_SEED: a4aa76e4a5ca70c8447dd544a63f180b5a6fe0aff96495802506354c10f2886e
```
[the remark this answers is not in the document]
david1alvarez: This should be removable. It was likely there as an environment variable for testing purposes

run_mobilecoin.sh
```diff
@@ -0,0 +1,61 @@
+SGX_SPID=9E933DB667EDDE8454258EF3CDC6C2BC
+SGX_PRIMARY_KEY=8e98028f29e6472ea4a9f1933c7fc093
+SGX_SECONDARY_KEY=acf143a705a249728a4665633a932038
+MC_SEED=a4aa76e4a5ca70c8447dd544a63f180b5a6fe0aff96495802506354c10f2886e
+SGX_MODE=HW
+IAS_MODE=PROD
+COMMIT_HASH=22d3894b6b556eb361167b121b934565d5d87c5d
```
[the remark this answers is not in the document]
david1alvarez: the SGX keys are sensitive, the commit hash is not. The commit hash is used to determine which commit is pointed to as part of building/running mobilecoin. I'll cycle the sgx keys now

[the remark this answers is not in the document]
david1alvarez: Good news, I've already rotated these keys, and that would've been done before the handoff to @pouneh. They should be removable

run_mobilecoin.sh
```diff
@@ -0,0 +1,61 @@
...
+
+
+    
+# run mob
+cd mobilecoin
+sudo ./mob prompt \
+mobilecoin \
+--expose 3200 3201 3202 \
+--publish 3200 3201 3202 \
+--command "cd tools/local-network && ls && ./bootstrap.sh && export MC_SEED=a4aa76e4a5ca70c8447dd544a63f180b5a6fe0aff96495802506354c10f2886e && export LEDGER_BASE=/tmp/mobilenode/target/sample_data/ledger && export IAS_API_KEY=8e98028f29e6472ea4a9f1933c7fc093 && export IAS_SPID=9E933DB667EDDE8454258EF3CDC6C2BC && python3 local_network.py --network-type a-b-c"
```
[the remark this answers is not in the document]
david1alvarez: IAS_API_KEY and IAS_SPID are keys/identifiers for SGX. Both of these keys are outdated / were rotated a long time ago, but its an error prone process (I've accidentally committed keys before and had to go cycle them several times in the past). Not 100% sure the concern of these particular keys getting leaked though, as they will just point to my personal SGX Developer Zone account

run_mobilecoin.sh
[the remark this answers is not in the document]
eranrund: Shouldn't we run the mobilecion version thats referenced in the submodule? If not, when would this get updated?
There should probably be some documentation around this.

tools/test_utils/fullservice.py
eranrund: How big is the ledger in these tests? Any reason to not wait for it to sync completely?

run_mobilecoin.sh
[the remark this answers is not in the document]
eranrund: It appears that this is the way to pass secrets to github actions: https://docs.github.com/en/actions/security-guides/encrypted-secrets

tools/test_utils/fullservice.py
cxloe: I think this is related to a FS bug actually. `http.client` expects connections to stay open per https://github.com/python/cpython/blob/3.10/Lib/http/client.py#L383 and FS doesn't seem to let this happen: 
image
@BrianCorbin any ideas?

eranrund: The fact it says the connection was left intact means it didn't get closed, as expected.

cxloe: persistent connections are the default for HTTP 1.1, for example curling google.com twice gives me this 
image

eranrund: Ah, I see. The connection does die for some reason.

cxloe: https://github.com/mobilecoinofficial/full-service/pull/453

Maybe this 🤔

cxloe:
Merged my PR in to try running the action.

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
+name: mc-ledger-test
```
eranrund, line 1: Why mc-ledger?

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+
+env:
+  SGX_MODE: SW
+  IAS_MODE: DEV
+  RUST_BACKTRACE: full
+  IAS_API_KEY: ${{ secrets.SGX_PRIMARY_API_KEY }}
```
eranrund, line 7: Why PRIMARY and not IAS?

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+      version_a_ref:
+        description: 'Tag or commit-hash for the base version of Full-Service'
+        type: string
+        required: true
+        default: 'v1.9.0' 
+      version_b_ref:
+        description: 'Tag or commit-hash to upgrade Full-Service to'
+        type: string
+        required: true
+        default: "v1.9.2"
```
eranrund, line 23: Nit: Other places using single quotes

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+  push:
+    branches:
+      - setup-mobilecoin
+      - pouneh/e2e_test_infra
```
eranrund, line 27: Seems like this doesn't belong here

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+    inputs:
```
eranrund, line 28: Why are there two `inputs` here? When is the first one used?

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+      - pouneh/e2e_test_infra
+    inputs: 
+      version_a_ref:
+        description: 'Tag or commit-hash for the base version of Full-Service'
+        required: true
+        default: 'v1.10.0-pre.5' #'v1.9.0' 
+      version_b_ref:
+        description: 'Tag or commit-hash to upgrade Full-Service to'
+        required: true
+        default: 'v1.9.2 '
```
eranrund, line 36: There's a space here that should be removed

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+        default: "v1.9.2"
+  push:
+    branches:
+      - setup-mobilecoin
+      - pouneh/e2e_test_infra
+    inputs: 
+      version_a_ref:
+        description: 'Tag or commit-hash for the base version of Full-Service'
+        required: true
+        default: 'v1.10.0-pre.5' #'v1.9.0'
```
eranrund, line 32: Why is this version newer than the one being upgraded to? Where are these versions coming from and when is someone expected to change them?

.gitignore
```diff
@@ -75,3 +75,7 @@ cli/build/**
 dbml-error.log
 
 *.profraw
+
+
+*__pycache__
+*bin
```
eranrund, line 81: What is this ignoring? Can it be narrowed down to something more specific?

.gitignore
```diff
@@ -75,3 +75,7 @@ cli/build/**
 dbml-error.log
 
 *.profraw
+
+
+*__pycache__
```
eranrund, line 80: Is the star here needed?

tools/build_full_service.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# Copyright (c) 2018-2022 The MobileCoin Foundation
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+PROJECT_DIR=$SCRIPT_DIR/../mobilecoin
+
+echo "building full service..."
+export RUST_LOG=debug
+export SGX_MODE=SW
```
eranrund, line 9: What about IAS_MODE? Is this export actually needed?

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+
+      - name: Build MobileCoin
+        shell: bash
+        run: |
+          tools/build_mobilecoin.sh
+      
+      - name: Build FullService
+        shell: bash
+        run: |
+          tools/build_full_service.sh
```
eranrund, line 87: Since this script does very little, maybe it can be inlined here?

tools/build_full_service.sh
```diff
@@ -0,0 +1,11 @@
...
+export CONSENSUS_ENCLAVE_CSS="$PROJECT_DIR/target/release/consensus-enclave.css"
```
eranrund, line 10: Where is this CSS file coming from?

tools/mc_ledger_test.py
```diff
@@ -0,0 +1,93 @@
...
+from test_utils.utilities import parse_network_type_cmd_line_args
+from test_utils import fullservice as fslib
+
+# Shoving the raw request into a python class would allow us to just access these things as members of the class
+# TODO: create a python dataclass to hold these
+balance = 'balance'
+pending = 'pending_pmob'
+unspent = 'unspent_pmob'
+spent = 'spent_pmob'
+fee = 'fee_pmob'
```
eranrund, line 14: Nit: I don't think this is making the code any more readable, but I think if you are going to use these global constants, it would be better if they are uppercase.

tools/mc_ledger_test.py
```diff
@@ -0,0 +1,93 @@
...
+        raise Exception(f'accounts not found in wallet')
+    elif len(fs.account_ids) < 2:
+        raise Exception(f'found {len(fs.account_ids)} account(s), minimum required is 2')
+
+    acc0_id = fs.account_ids[0]
+    acc1_id = fs.account_ids[1]    
+    
+    account_0 = fs.account_map[acc0_id]
+    account_1 = fs.account_map[acc1_id]
+    p_mob_amount = 600_000_000
```
eranrund, line 28: Where is this number coming from?

tools/test_utils/constants.py
eranrund: This question still stands, what is using `FS_LOG`?

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,213 @@
...
+        self.remove_wallet_and_ledger = remove_wallet_and_ledger
+        self.wallet_path = pathlib.Path('/tmp/wallet-db')
+        self.ledger_path = pathlib.Path('/tmp/ledger-db')
+        
+    def __enter__(self):
+        self.remove_wallet_and_ledger = True
+        self.start()
+        return self
+    
+    def __exit__(self,exc_type, exc_val, exc_tb):
```
eranrund, line 32:
```suggestion
    def __exit__(self, exc_type, exc_val, exc_tb):
```

tools/test_utils/fullservice.py
eranrund: This remains an issue. I don't see why we would need to sleep 2 seconds after every request we issue to FS, no matter what the response is.

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,213 @@
...
+        params = {
+            'mnemonic': mnemonic,
+            'key_derivation_version': '2',
+        }
+        r = self._request({
+            "method": "import_account",
+            "params": params
+        })
+
+        if 'error' in r:
```
eranrund, line 109: Isn't this being swallowed by the code in lines 93-94? I don't recall the response having nested `error` fields (e.g. `{"error": {"error": ...}}`)

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,213 @@
...
+            return response_data['result']
+
+    def import_account(self, mnemonic) -> bool:
+        print(f'importing full service account {mnemonic}')
+        params = {
+            'mnemonic': mnemonic,
+            'key_derivation_version': '2',
+        }
+        r = self._request({
+            "method": "import_account",
```
eranrund, line 105: Nit: either use single quotes or double quotes, but please choose one.

tools/test_utils/fullservice.py
eranrund: This question is still relevant, not sure why this was marked as resolved.

tools/test_utils/fullservice.py
```diff
@@ -0,0 +1,213 @@
...
+            "method": "get_wallet_status"
+        })
+        return r['wallet_status']
+
+    # ensure at least two accounts are in the wallet. Some accounts are imported by default, but the number changes.
+    def setup_accounts(self):
+        account_ids, account_map = self.get_all_accounts()
+        if len(account_ids) >= 2:
+            self.account_ids = account_ids
+            self.account_map = account_map
```
eranrund, line 161: These would get overridden by the `get_all_accounts` call below (at the end of the method). That call should probably live under the `else:` clause.

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+  SGX_MODE: SW
+  IAS_MODE: DEV
+  RUST_BACKTRACE: full
+  IAS_API_KEY: ${{ secrets.SGX_PRIMARY_API_KEY }}
+  IAS_SPID: ${{ secrets.SGX_SPID }}
+  MC_LOG: warn
+
+on:
+  workflow_dispatch:
+    inputs:
```
eranrund, line 13: Where are these inputs used?

eranrund (changes requested):
Nice, this is looking better, but there are still some open questions and a few nits.

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+        description: 'Tag or commit-hash to upgrade Full-Service to'
+        required: true
+        default: 'v1.9.2 '
+
+# TESTING WITH NEKTOS/ACT
+# nektos/act is a 3rd party github library that allows for the LOCAL testing of github actions
+# best nektox/act testing flow is:
+  # 2 change the runs-on tag to point to `ubuntu-latest` if its set to a self-hosted tag
+    # 2.1 `ubuntu-latest` will use Github Action minutes, but is required for testing with nektos/act
+    # 2.2 `[self-hosted, Linux, large]` will use a self-hosted machine (free for us), but fails with nektos/act
```
remoun, line 43: FYI `act` has a `-P` flag to override the platform mappings; I usually pass `-P self-hosted=mobilecoin/builder-install:latest` or `-P self-hosted=ubuntu-latest`

.github/workflows/mc-ledger-test.yml
```diff
@@ -0,0 +1,98 @@
...
+
+jobs:
+  integration-test:
+    # set runs-on to `[self-hosted, Linux, large]` for running on github actions
+    # and `ubuntu-latest` for running on locally with nektos/act. 
+    runs-on: [self-hosted, Linux, large]
+    container:
+      # image should be kept up to date with the url and tag listed in mobilecoin/.mobconf
+      # e.g. {{url}}:{{tag}} to prevent build failures
+      image: mobilecoin/builder-install:v0.0.10
```
remoun, line 67: latest is 17
```suggestion
      image: mobilecoin/builder-install:v0.0.17
```

briancorbin:
Closing for now, will revisit in the future.

PR #436, opened 2022-09-14
Feature/refactor transaction signer

BrianCorbin:


full-service/src/db/account.rs
```diff
@@ -574,6 +578,26 @@ impl AccountModel for Account {
...
     }
+
+    fn view_private_key(self) -> Result<RistrettoPrivate, WalletDbError> {
+        if self.view_only {
+            let view_account_key: ViewAccountKey = mc_util_serial::decode(&self.account_key)?;
+            return Ok(*view_account_key.view_private_key());
+        }
+
+        let account_key: AccountKey = mc_util_serial::decode(&self.account_key)?;
+        return Ok(*account_key.view_private_key());
```
eranrund, line 589: Can this be simplified and instead call `self.view_account_key()?.view_private_key()`?

full-service/src/service/transaction_builder.rs
```diff
@@ -416,6 +439,179 @@ impl<FPR: FogPubkeyResolver + 'static> WalletTransactionBuilder<FPR> {
...
+                    0
+                }
+            };
+
+            if ring.len() != membership_proofs.len() {
+                return Err(WalletTransactionBuilderError::RingSizeMismatch);
+            }
+
+            let onetime_key_derive_data = OneTimeKeyDeriveData::SubaddressIndex(
+                utxo.subaddress_index.unwrap_or_default() as u64,
```
eranrund, line 555: This seems problematic to me. Why would a utxo not have a subaddress index? If we just default to 0, I expect we can get weird and hard to debug errors.

full-service/src/service/transaction_builder.rs
```diff
@@ -416,6 +439,179 @@ impl<FPR: FogPubkeyResolver + 'static> WalletTransactionBuilder<FPR> {
...
+                        .or_insert(utxo.value as u64);
+                    acc
+                });
+
+        for (token_id, input_value) in input_value_per_token {
+            let total_value = total_value_per_token.get(&token_id).ok_or_else(|| {
+                WalletTransactionBuilderError::MissingInputsForTokenId(token_id.to_string())
+            })?;
+
+            println!("total_value: {}", total_value);
```
eranrund, line 596: debug code that shouldn't be here

transaction-signer/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
```
eranrund, line 1: probably shouldnt be here since this crate is empty at the moment

eranrund (commented):
Has the new transaction building code been tested in any way?

full-service/src/service/transaction_builder.rs
BrianCorbin: ah, just there for testing

BrianCorbin:
> Has the new transaction building code been tested in any way?

I've manually tested it and verified that it works on test net

full-service/src/db/account.rs
```diff
@@ -574,6 +580,35 @@ impl AccountModel for Account {
...
+    }
+
+    fn view_private_key(&self) -> Result<RistrettoPrivate, WalletDbError> {
+        if self.view_only {
+            let view_account_key: ViewAccountKey = mc_util_serial::decode(&self.account_key)?;
+            return Ok(*view_account_key.view_private_key());
+        }
+
+        let account_key: AccountKey = mc_util_serial::decode(&self.account_key)?;
+        return Ok(*account_key.view_private_key());
```
eranrund, line 610: Can this be simplified by calling `self.view_account_key()?.view_private_key()`?

full-service/src/service/gift_code.rs
```diff
@@ -459,7 +459,10 @@ where
         )?;
 
         let account_key: AccountKey = mc_util_serial::decode(&from_account.account_key)?;
-        let tx_proposal = unsigned_tx.sign(&account_key, fog_resolver)?;
+        let signer = LocalRingSigner::from(&account_key);
+        let mut rng = rand::thread_rng();
+        let tx = signing_data.sign(&signer, &mut rng)?;
+        let tx_proposal = TxProposal::new(tx, signing_data);
```
eranrund, line 465: I've seen this duplicated, is it possible to implement a method on `TxProposal` that does it?
`TxProposal::from_signing_data(&signing_data, &account_key)`?
Or maybe `TxProposal::new()` can be changed to do this?

transaction-signer/src/main.rs
```diff
@@ -1,3 +1,378 @@
+use bip39::{Language, Mnemonic, MnemonicType};
```
eranrund, line 1: Nit: Missing copyright notice

transaction-signer/src/main.rs
```diff
@@ -1,3 +1,378 @@
...
+    let account_key = account_key_from_mnemonic_phrase(&account_secrets.mnemonic.unwrap());
+
+    // Load input txos.
+    let sync_request_data =
+        fs::read_to_string(sync_request).expect("Could not open sync request file.");
+    let sync_request_json: serde_json::Value =
+        serde_json::from_str(&sync_request_data).expect("Malformed sync request.");
+    let account_id = sync_request_json
+        .get("account_id")
+        .unwrap()
```
eranrund, line 179: Using expect will make it easier to see what went wrong if this ever fails.

transaction-signer/src/main.rs
```diff
@@ -1,3 +1,378 @@
...
+        sync_request_json
+            .get("incomplete_txos_encoded")
+            .expect("Could not find \"incomplete_txos_encoded\".")
+            .clone(),
+    )
+    .expect("Malformed sync request.");
+    let input_txos: Vec<TxOut> = incomplete_txos_encoded
+        .iter()
+        .map(|tx_out_serialized| {
+            mc_util_serial::decode(&hex::decode(tx_out_serialized.as_bytes()).unwrap()).unwrap()
```
eranrund, line 194: Same, expect might make it easier to debug if this ever fails

transaction-signer/src/main.rs
```diff
@@ -1,3 +1,378 @@
...
+        .try_into_account_key("", "", &base64::decode("").unwrap())
+        .unwrap()
+}
+
+fn get_key_image_for_tx_out(
+    tx_out: &TxOut,
+    account_key: &AccountKey,
+    subaddress_spend_public_keys: &HashMap<RistrettoPublic, u64>,
+) -> Option<(KeyImage, u64)> {
+    let tx_public_key = match RistrettoPublic::try_from(&tx_out.public_key) {
```
eranrund, line 319: I think this can be simplified to `let tx_public_key = RistrettoPublic::try_from(&tx_out.public_key).ok()?;`

eranrund (approved):
I know most code here already existed before this PR, so I didn't review every part of it but for the future I think it could benefit from some more documentation - especially around the command line commands and their parameters, perhaps less unwraps/expects and using logging instead of print statements. It just has a feel of a proof of concept and not a production piece of code, which might be fine - I am not super familiar with its intended use.
I am also not a fan of the hardcoded file names, it is more conventional to have the filenames be passed in as command line arguments.

PR #437, opened 2022-09-14
Feature/update to fog resolver

BrianCorbin:


full-service/src/bin/transaction-signer.rs
```diff
@@ -352,7 +352,11 @@ fn tx_out_belongs_to_account(tx_out: &TxOut, account_view_private_key: &Ristrett
         Ok(k) => k,
     };
     let shared_secret = get_tx_out_shared_secret(account_view_private_key, &tx_out_public_key);
-    tx_out.masked_amount.get_value(&shared_secret).is_ok()
+    tx_out
+        .get_masked_amount()
+        .unwrap()
```
eranrund, line 357: unwrap seems wrong here, it appears the function should return false if this fails.

full-service/src/config.rs
```diff
@@ -120,7 +120,7 @@ impl APIConfig {
                 .build(),
         );
 
-        let conn = GrpcFogReportConnection::new(env, logger.clone());
+        let conn = GrpcFogReportConnection::new("".to_string(), env, logger.clone());
```
eranrund, line 123: We should add a new command line argument `--chain-id` and pass it here.

full-service/src/config.rs
```diff
@@ -213,6 +213,7 @@ impl PeersConfig {
             .iter()
             .map(|client_uri| {
                 ThickClient::new(
+                    "".to_string(),
```
eranrund, line 216: We should add a new command line argument `--chain-id` and pass it here.

full-service/src/db/transaction_log.rs
```diff
@@ -1069,7 +1069,7 @@ mod tests {
             &wallet_db.get_conn().unwrap(),
         )
         .unwrap();
-        assert_eq!(input_details0.value as u64, 7 * MOB);
+        assert_eq!(input_details0.value, associated_txos.inputs[0].value);
```
eranrund, line 1072: What caused this change?

full-service/src/error.rs
```diff
@@ -315,6 +315,9 @@ pub enum WalletTransactionBuilderError {
 
     /// Burn Redemption Memo must be exactly 128 characters (64 bytes) long.
     InvalidBurnRedemptionMemo(String),
+
+    /// Error converting a TxOut
+    TxOutConversion(mc_transaction_core::TxOutConversionError),
```
eranrund, line 320: This error, as well as a few ones above it, fail to render the argument (they are missing the `: {0}` in the docstring).

full-service/src/json_rpc/v1/models/amount.rs
```diff
@@ -60,14 +60,17 @@ impl TryFrom<&MaskedAmount> for mc_transaction_core::MaskedAmount {
             &hex::decode(&src.commitment)
                 .map_err(|err| format!("Could not decode hex for amount commitment: {:?}", err))?,
         );
-        Ok(Self {
+
+        let masked_amount = mc_transaction_core::MaskedAmountV1 {
```
eranrund, line 64: This is scary... whats going to happen when we begin supporting masked amount v2? Maybe not in the v1 full-service API?

full-service/src/json_rpc/v1/models/receiver_receipt.rs
```diff
@@ -95,7 +95,7 @@ mod tests {
         let account_key = AccountKey::random(&mut rng);
         let public_address = account_key.default_subaddress();
         let txo = TxOut::new(
-            BlockVersion::MAX,
+            BlockVersion::TWO,
```
eranrund, line 98: Why was this changed? Seems valuable to have a test that tests the newest block version.
In other places we put such tests in a loop that tries all relevant block versions.

full-service/src/json_rpc/v2/models/masked_amount.rs
```diff
@@ -54,14 +54,17 @@ impl TryFrom<&MaskedAmount> for mc_transaction_core::MaskedAmount {
             &hex::decode(&src.commitment)
                 .map_err(|err| format!("Could not decode hex for amount commitment: {:?}", err))?,
         );
-        Ok(Self {
+
+        let masked_amount_v1 = mc_transaction_core::MaskedAmountV1 {
```
eranrund, line 58: The v2 API should plan to support MaskedAmountV2...

validator/service/src/validator_api.rs
```diff
@@ -71,6 +71,7 @@ impl<UTC: UserTxConnection + 'static> ValidatorApi<UTC> {
             conn_manager,
             submit_node_offset: Arc::new(AtomicUsize::new(0)),
             fog_report_connection: GrpcFogReportConnection::new(
+                "".to_string(),
```
eranrund, line 74: Need to be passed from command line

eranrund (changes requested):
Spotted a few things that I think require attention:
1) tests that got pinned to an old block version (::TWO) - why not MAX? and can they instead be made to test all version (e.g. TWO..MAX)?
2) chain-id needs to be added to the command line config and used when creating grpc connections
3) MaskedAmountV2 is a thing that will need to be supported. Could be a future todo, but things will get messy once we switch to the block version that uses that.

full-service/src/json_rpc/v1/models/receiver_receipt.rs
BrianCorbin: yeah agreed, was testing to see what would change. I'll put TODO comments from now on, bad habit of mine not to.

full-service/src/config.rs
BrianCorbin: ah okay, yeah that works

BrianCorbin: Optional? Default value?

full-service/src/json_rpc/v2/models/masked_amount.rs
BrianCorbin: agreed, will add that in 👍

full-service/src/config.rs
eranrund: Nope, required

BrianCorbin: ah okay, thanks! Will add that as a required cla

BrianCorbin: Where can I find info on chain-ids?

full-service/src/json_rpc/v1/models/receiver_receipt.rs
BrianCorbin: Will create a gh issue to track that

full-service/src/db/transaction_log.rs
BrianCorbin: This should no longer be relevant, that particular code in the test has changed to better reflect the desired outcome

full-service/src/bin/transaction-signer.rs
BrianCorbin: This code in particular will be addressed in the following PR #436

full-service/src/json_rpc/v1/models/receiver_receipt.rs
eranrund: What breaks if it is set to ::MAX?

validator/service/src/config.rs
```diff
@@ -24,4 +24,7 @@ pub struct Config {
     /// How many seconds to wait between polling.
     #[structopt(long, default_value = "5", parse(try_from_str=parse_duration_in_seconds))]
     pub poll_interval: Duration,
+
+    #[structopt(long)]
+    pub chain_id: String,
```
eranrund, line 29: Would become unnecessary once its moves into `PeersConfig`

full-service/src/config.rs
eranrund: https://github.com/mobilecoinfoundation/mcips/blob/main/text/0049-chain-ids.md

full-service/src/json_rpc/v1/models/receiver_receipt.rs
BrianCorbin: Nothing now, it should be back to MAX

BrianCorbin: oh wait I missed this one, hahha

full-service/src/service/receipt.rs
```diff
@@ -144,7 +156,16 @@ impl TryFrom<&mc_api::external::Receipt> for ReceiverReceipt {
         let public_key: CompressedRistrettoPublic =
             CompressedRistrettoPublic::try_from(src.get_public_key())?;
         let confirmation = TxOutConfirmationNumber::try_from(src.get_confirmation())?;
-        let amount = MaskedAmount::try_from(src.get_masked_amount())?;
+
+        let amount = if let Ok(masked_amount_v2) = MaskedAmountV2::try_from(src.get_masked_amount())
```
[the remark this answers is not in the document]
BrianCorbin: To try to decode to V2 first, and then V1, which seems to be working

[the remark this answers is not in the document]
eranrund: Isn't this just a coincidence based on what block version the tests are using? If the fields are identical when would MaskedAmountV2::try_from ever fail and MaskedAmountV1::try_from succeed?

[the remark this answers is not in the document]
eranrund: It looks like @garbageslam put out https://github.com/mobilecoinfoundation/mobilecoin/pull/2532 so maybe upreving to that once it gets merged in a bit will remove the confusion here?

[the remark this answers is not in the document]
BrianCorbin: Yeah that's a good idea, let's wait on that

full-service/src/json_rpc/v1/models/amount.rs
```diff
@@ -27,26 +39,24 @@ pub struct MaskedAmount {
...
-    fn from(src: &mc_api::external::MaskedAmount) -> Self {
-        Self {
-            object: "amount".to_string(),
-            commitment: hex::encode(src.get_commitment().get_data()),
-            masked_value: src.get_masked_value().to_string(),
-            masked_token_id: hex::encode(&src.get_masked_token_id()),
-        }
-    }
+    /// The version of the masked amount.
+    pub version: MaskedAmountVersion,
```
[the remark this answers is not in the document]
BrianCorbin: I'm tempted to just remove the receipt stuff from v1 entirely and make it v2 only at this point like I did with view only accounts.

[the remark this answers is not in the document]
BrianCorbin: but also it really doesn't matter that much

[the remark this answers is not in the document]
BrianCorbin: no, only MOB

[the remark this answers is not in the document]
eranrund: then I think it's safe for the V1 API to only support MaskedAmountV1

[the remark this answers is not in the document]
eranrund: What made you resolve this? I think this still breaks the v1 API since there's now a new required field that didn't exist before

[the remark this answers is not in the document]
BrianCorbin: Ah, I misunderstood, this should always assume MaskedAmountV1

eranrund (approved):

PR #462, opened 2022-09-22
Get Block Index by TXO public key

BrianCorbin:
Adds the ability to check the block index of a txo's public key, if it exists on the ledger

Todo:

- [x] functionality
- [x] documentation
- [x] test(s)

### DIFF
--- a/docs/v2/api-endpoints/validate_tx_out.md
+++ b/docs/v2/api-endpoints/validate_tx_out.md
@@ -1,0 +1,60 @@
+# Validate TxOut
+
+Allows the public key of a tx out to be checked against the ledger, and if it exists will return the block index
+
+## Request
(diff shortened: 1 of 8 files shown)


docs/v2/api-endpoints/validate_tx_out.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Response
+
+## Example
+
+{% tabs %}
+{% tab title="Request Body" %}
+```text
+{
+    "method": "validate_tx_out",
```
eranrund, line 19: I think the name validate txout is problematic since no validation is taking place... how about `"get_tx_out_block_index"`?

eranrund (approved):

PR #473, opened 2022-09-28
Adding back old migrations

briancorbin:
This is in support of the changes for migrating to v2.0.0 directly via database migrations instead of an external tool. The first step is to add back the old migrations and to retroactively insert the history of migrations if it's missing (which would be the case for anyone who has used 1.10/2.0.0 before this update)


Shramp (approved):
glad to see there's a plan for doing the update via migrations instead of the external tool.

PR #482, opened 2022-09-30
Bugfix/incorrect account height reported

briancorbin:
fixes a bug where the next subaddress index was being reported as the account height in the v1 API response for get_balance_for_account and get_balance_for_address

also gets rid of several duplicate tests


full-service/src/json_rpc/v1/e2e_tests/account/account_other.rs
```diff
@@ -184,527 +184,4 @@ mod e2e_account {
         );
         let _account = result.get("account").unwrap();
     }
-
-    #[test_with_logger]
-    fn test_e2e_get_balance(logger: Logger) {
```
eranrund, line 189: Why was all of this removed?

briancorbin: Duplicate tests

briancorbin: There are more dupe tests than this which I'll go through, didn't realize until I was testing the ones specific to this bugfix

PR #496, opened 2022-10-02
Fix incomplete received tx_log response for v1 API

briancorbin:


david1alvarez (commented):
should be good, though it looks like some of the conditional flows are deeper than they need to be

Shramp (approved):

PR #502, opened 2022-10-04
Fix issues surfaced by block version 2

eranrund:
### Motivation

We discovered a few issues as a result of the recent switch to block version 2.

### In this PR
* Fixing `mc-validator-service` to properly forward the complete fee map and block version
* Change fee map handling - prior to this PR we chose the max fee for each token, but that allows a malicious node to do a grief attack and trick clients into higher than necessary fee. This PR changes the strategy around this, and instead enforces that all nodes are in agreement about the fee map. If they aren't, that indicates a network fork and should be carefully handled by the operator.
* Make `mc-validator-service` validate chain-id headers if the chain-id argument is passed to it.

### DIFF
--- a/full-service/src/service/ledger.rs
+++ b/full-service/src/service/ledger.rs
@@ -201,7 +201,7 @@ where
         }
 
         // Guaranteed not to fail since we verified last_block_infos is not empty.
-        return last_block_infos[0].minimum_fees.clone();
+        last_block_infos[0].minimum_fees.clone()
     }
 
     fn get_network_block_version(&self) -> BlockVersion {
--- a/full-service/src/service/ledger.rs
+++ b/full-service/src/service/ledger.rs
@@ -12,7 +12,9 @@ use crate::{
 };
 use mc_blockchain_types::{Block, BlockContents, BlockVersion, BlockVersionError};
 use mc_common::HashSet;
-use mc_connection::{BlockchainConnection, RetryableBlockchainConnection, UserTxConnection};
+use mc_connection::{
+    BlockInfo, BlockchainConnection, RetryableBlockchainConnection, UserTxConnection,
+};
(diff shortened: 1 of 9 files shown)


full-service/src/service/ledger.rs
```diff
@@ -176,25 +181,28 @@ where
...
+        }
+
+        // Return the fee map.
+        last_block_infos
+            .get(0)
+            .map(|info| info.minimum_fees.clone())
+            .ok_or(LedgerServiceError::NoLastBlockInfo)
     }
 
     fn get_network_block_version(&self) -> BlockVersion {
```
briancorbin, line 208: Should this also do the same as the above?

eranrund: That's a tricky one. There's an edgecase where a node could be behind for whatever reason, just at the time where a new block version was introduced. So expecting all nodes to agree on the block version could mean returning an intermittent error. I am not sure there's a right answer here, but I think the options are:
1) Erroring when all the available nodes do not agree on the block version, with the assumption that this should be a temporary condition
2) Choosing the most common block version at the time the query is made. I am leaning towards this option.
Thoughts?

briancorbin: Maybe the highest, since the rest _should_ be catching up to the one that has the highest?

briancorbin: Unless there is a malicious node that's setting the BlockVersion higher than the rest, causing it to always submit to that one

briancorbin: I think the intermittent error is acceptable, honestly, since how often are we assuming this is going to happen?

eranrund: Pushed a change to treat it similarly to the minimum_fee map

briancorbin (approved):

PR #504, opened 2022-10-05
Fix issues surfaced by block version 2 (#502)

briancorbin:
This is a cherry pick of #502 to the release branch for 1.9.5

### DIFF
--- a/full-service/src/service/gift_code.rs
+++ b/full-service/src/service/gift_code.rs
@@ -17,6 +17,7 @@ use crate::{
     service::{
         account::AccountServiceError,
         address::{AddressService, AddressServiceError},
+        ledger::LedgerService,
         transaction::{TransactionService, TransactionServiceError},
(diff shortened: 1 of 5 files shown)


full-service/src/service/gift_code.rs
```diff
@@ -53,6 +54,8 @@ use rand::Rng;
 use serde::{Deserialize, Serialize};
 use std::{convert::TryFrom, fmt, iter::empty, str::FromStr, sync::atomic::Ordering};
 
+use super::ledger::LedgerServiceError;
```
eranrund, line 57: Nit: import using `crate::` like done above

eranrund (approved):

full-service/src/service/gift_code.rs
briancorbin: ah, auto import likes to default to using super for some reason

PR #507, opened 2022-10-07
Build Script

### LINKED ISSUE #499: Script for building mainnet and testnet
**Is your feature request related to a problem? Please describe.**
It can be frustrating to build for different environments (mainnet, testnet, etc) because you need specific node endpoints and CSS files, and if you get those wrong it will cause errors during runtime that can be a bit cryptic if you are unfamiliar with Full Service.

**Describe the solution you'd like**
Add script(s) to the tools folder that builds mainnet and testnet binaries, including downloading the necessary consensus-enclave.css file. The ingest-enclave.css file isn't needed in the build step and can be downloaded exclusively in the run-<net> script.

**Describe alternatives you've considered**

**Additional context**

pouneh:
Create a build script that downloads sigstructs and builds full service for the specified version of the mobilecoin network

Resolves #499

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ env:
   CONSENSUS_ENCLAVE_CSS: /var/tmp/consensus-enclave.css
   INGEST_ENCLAVE_CSS: /var/tmp/ingest-enclave.css
 
-# only perform these build steps on pre-release
+# only perform these build steps on pre-release or forced dev build
 on:
   push:
     tags:
@@ -15,18 +15,27 @@ on:
       - '*.dev-build.*'
 
 jobs:
-  macos-x64:
-    runs-on: [self-hosted, macOS, X64, cargo]
+  build-and-pre-release:
     permissions:
       contents: write
     strategy:
       matrix:
+        runner-tags: [[self-hosted, macOS, X64, cargo], [self-hosted, macOS, ARM64, cargo], [self-hosted, Linux, large]]
+        namespace: [test, prod]
         include:
+          - runner-tags: [self-hosted, macOS, X64, cargo]
+            container: null
+          - runner-tags: [self-hosted, macOS, ARM64, cargo]
+            container: null
+          - runner-tags: [self-hosted, Linux, large]
+            container: mobilecoin/rust-sgx-base:latest
           - namespace: test
             network: testnet
           - namespace: prod
             network: mainnet
-            
+    runs-on: ${{ matrix.runner-tags }}
+    container: ${{ matrix.container }}
+
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -34,12 +43,9 @@ jobs:
           submodules: recursive
 
       - name: Brew Bundle
+        if: runner.os == 'macOS'
         run: |
           brew bundle
-      
-      - name: Git Submodule
-        run: |
-          git submodule update --checkout --init --recursive
 
       - name: Consensus SigStruct
         run: |
@@ -59,26 +65,26 @@ jobs:
         run: |
           mkdir -pv build_artifacts/${{ matrix.network }}/bin
           cp /var/tmp/*.css build_artifacts/${{ matrix.network }}
-          cp target/release/mc-full-service build_artifacts/${{ matrix.network }}/bin/
-          cp target/release/mc-transaction-signer build_artifacts/${{ matrix.network }}/bin/
-          cp target/release/mc-validator-service build_artifacts/${{ matrix.network }}/bin/
+          cp target/release/mc-full-service build_artifacts/${{ matrix.network }}
+          cp target/release/mc-transaction-signer build_artifacts/${{ matrix.network }}
+          cp target/release/mc-validator-service build_artifacts/${{ matrix.network }}
 
       - name: Create Artifact
         run: |
           mkdir -pv artifact
-          cd artifact && tar -czvf ${{ github.sha }}-${{ runner.os }}-x86-${{ matrix.network }}.tar.gz -C ../build_artifacts/${{ matrix.network }}/ .
+          cd artifact && tar -czvf ${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.network }}.tar.gz -C ../build_artifacts/${{ matrix.network }}/ .
 
       - name: Upload Artifact
         uses: actions/upload-artifact@v3
         with:
-          name: full-service_${{ runner.os }}_${{ matrix.network }}_x86
-          path: artifact/${{ github.sha }}-${{ runner.os }}-x86-${{ matrix.network }}.tar.gz
+          name: full-service_${{ runner.os }}-${{ runner.arch }}-${{ matrix.network }}
+          path: artifact/${{ github.sha }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.network }}.tar.gz
 
       - name: Create Prerelease
         if: startsWith(github.ref, 'refs/tags/v')
         run: |
           mkdir -pv release
-          cd release && tar -czvf ${{ github.ref_name }}-${{ runner.os }}-x86-${{ matrix.network }}.tar.gz -C ../build_artifacts/${{ matrix.network }}/ .
+          cd release && tar -czvf ${{ github.ref_name }}-${{ runner.os }}-${{ runner.arch }}-${{ matrix.network }}.tar.gz -C ../build_artifacts/${{ matrix.network }}/ .
 
       - name: Upload Prerelease
         if: startsWith(github.ref, 'refs/tags/v')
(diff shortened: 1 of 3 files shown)


tools/build-fs.sh
```diff
@@ -0,0 +1,45 @@
...
+NET=$1
+fi
+
+if [ $NET == "local" ]; then
+echo "Local networks are not currently supported with this script"
+exit 1
+elif [ $NET == "test" ]; then
+echo "Building Full Service with testnet settings"
+NAMESPACE="test"
+SM=SW
```
eranrund, line 19: Can be simplified by just setting SGX_MODE/IAS_MODE

tools/build-fs.sh
```diff
@@ -0,0 +1,45 @@
...
+
+CONSENSUS_ENCLAVE_CSS="$WORK_DIR/consensus-enclave.css"
+
+(cd ${WORK_DIR} && CONSENSUS_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.${NAMESPACE}.mobilecoin.com/production.json | grep consensus-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
+curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${CONSENSUS_SIGSTRUCT_URI})
+
+echo "building full service..."
+SGX_MODE=$SM \
+IAS_MODE=$IM \
+CONSENSUS_ENCLAVE_CSS=$CONSENSUS_ENCLAVE_CSS \
```
eranrund, line 43: if you change the environment variable assignments to be prefixed with `export` they will make their way to cargo build and you wont have to re-iterate it here

tools/build-fs.sh
```diff
@@ -0,0 +1,45 @@
...
+cargo build --release -p mc-full-service
```
eranrund, line 44: I'd remove the `-p mc-full-service`, is there a reason not to build the other binaries in the full-service workspace?

tools/build-fs.sh
```diff
@@ -0,0 +1,45 @@
...
+IM=PROD
+else
+echo "Network specified is not valid. Try 'test' or 'main'. Default is test"
+exit 1
+fi
+
+WORK_DIR="$HOME/.mobilecoin/${NET}"
+WALLET_DB_DIR="${WORK_DIR}/wallet-db"
+LEDGER_DB_DIR="${WORK_DIR}/ledger-db"
+mkdir -p ${WORK_DIR}
```
eranrund, line 33: This seems unrelated to building full-service. Should this be part of a "run full service" script instead?

tools/build-fs.sh
briancorbin: Agreed, can remove the project specific tag

tools/build-fs.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+if [ $NET == "local" ]; then
+echo "Local networks are not currently supported with this script"
+exit 1
+elif [ $NET == "test" ]; then
+echo "Building Full Service with testnet settings"
+NAMESPACE="test"
+export SGX_MODE=SW
+export IAS_MODE=DEV
```
eranrund, line 20: I missed this at a previous review, but testnet uses HW/PROD like mainnet.
Local network uses SW/DEV

tools/build-fs.sh
```diff
@@ -1,39 +1,41 @@
...
+    NAMESPACE="test"
+    export SGX_MODE=HW
+    export IAS_MODE=PROD
+    elif [ $NET == "main" ]; then
+    NAMESPACE="prod"
+    export SGX_MODE=HW
+    export IAS_MODE=PROD
+    fi 
+    
+    WORK_DIR="$HOME/.mobilecoin/${NET}"
```
eranrund, line 33: What will happen here if NET is not set?

eranrund (changes requested):

tools/build-fs.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+if [ -z ${NET} ] || [ $NET -ne "test" ] || [ $NET -ne "main" ]; then
+echo "Using current environment's SGX_MODE, IAS_MODE, CONSENSUS_ENCLAVE_CSS "
+else 
+    if [ $NET == "test" ]; then
+    echo "Building Full Service with testnet settings"
+    NAMESPACE="test"
+    export SGX_MODE=HW
+    export IAS_MODE=PROD
```
eranrund, line 26: Nit: Needs to be indented

tools/build-fs.sh
```diff
@@ -0,0 +1,39 @@
...
+#   version of the following variables are used
+#   - SGX_MODE
+#   - IAS_MODE
+#   - CONSENSUS_ENCLAVE_CSS
+#   - NAMESPACE (if no namespace is present, we use 'default')
+
+
+# Net can be main/test/local
+if [ $# -gt 0 ]; then
+NET=$1
```
eranrund, line 15: This should be indented.

tools/build-fs.sh
eranrund: IT appears that we can get here with `NET` not being set to anything, which would just set `WORK_DIR` to `$HOME/.mobilecoin/`

tools/build-fs.sh
```diff
@@ -0,0 +1,39 @@
...
+        export IAS_MODE=PROD
+    fi 
+    
+    WORK_DIR="$HOME/.mobilecoin/${NET}"
+    mkdir -p ${WORK_DIR}
+    export CONSENSUS_ENCLAVE_CSS="$WORK_DIR/consensus-enclave.css"
+    (cd ${WORK_DIR} && CONSENSUS_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.${NAMESPACE}.mobilecoin.com/production.json | grep consensus-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
+        curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${CONSENSUS_SIGSTRUCT_URI})
+else
+    echo "Using current environment's SGX_MODE, IAS_MODE, CONSENSUS_ENCLAVE_CSS "
```
eranrund, line 35:
```suggestion
    echo "Using current environment's SGX_MODE, IAS_MODE, CONSENSUS_ENCLAVE_CSS"
```

tools/build-fs.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# Copyright (c) 2018-2022 The MobileCoin Foundation
+
+# To use this script, run build-fs test or build-fs main
+# If no network is specified, or a different network is specified, the env's 
+#   version of the following variables are used
+#   - SGX_MODE
+#   - IAS_MODE
+#   - CONSENSUS_ENCLAVE_CSS
+#   - NAMESPACE (if no namespace is present, we use 'default')
```
eranrund, line 10: I don't see the string `default` mentioned anywhere in this script other than this comment.
On this topic, what do we get from allowing this script to run without a namespace?

eranrund (changes requested):

tools/build-fs.sh
eranrund: Ah, my mistake. I misread the `if` statement.
Looks like it can be simplified to just `[ $NET == "test" ] || [ $NET == "main" ]` since that already implies `$NET` is not empty

tools/build-fs.sh
```diff
@@ -0,0 +1,39 @@
...
+    WORK_DIR="$HOME/.mobilecoin/${NET}"
+    mkdir -p ${WORK_DIR}
+    export CONSENSUS_ENCLAVE_CSS="$WORK_DIR/consensus-enclave.css"
+    (cd ${WORK_DIR} && CONSENSUS_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.${NAMESPACE}.mobilecoin.com/production.json | grep consensus-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
+        curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${CONSENSUS_SIGSTRUCT_URI})
+else
+    echo "Using current environment's SGX_MODE, IAS_MODE, CONSENSUS_ENCLAVE_CSS"
+fi
+
+exit 0
```
eranrund, line 36: Shouldn't be here

eranrund (changes requested):

tools/build-fs.sh
eranrund: This should be mentioned in a comment then.
Also, it would be helpful for me if conversations are not marked as resolved until I have a chance to look at them, otherwise I can't tell what really was addressed and what still has a followup (such as documenting this behavior in a comment). Thanks!

eranrund: Unrelated, I believe that if this is changed to `"$NET" == "test"` then there won't be need for the `-z` test...

tools/build-fs.sh
```diff
@@ -0,0 +1,38 @@
...
+#   - SGX_MODE
+#   - IAS_MODE
+#   - CONSENSUS_ENCLAVE_CSS
+
+# Net can be main/test/local
+if [ $# -gt 0 ]; then
+    NET=$1
+fi
+
+if ([ "$NET" == "test" ] || [ "$NET" == "main" ]); then
```
eranrund, line 16: Nit: Outer parenthesis are likely not required

tools/build-fs.sh
eranrund: Thank you!

eranrund (approved):
LGTM once the nit is addressed

briancorbin (approved):

PR #508, opened 2022-10-07
Run script 

### LINKED ISSUE #440: The `tools/run-mainnet.sh` script is broken
1. It doesn’t really get the css files since I the URLs are wrong (the hostnames for mainnet are `.prod.`)
2. It doesn’t do anything with the consensus one - it should be building using it (or there should be a `build-mainnet` script)

pouneh:
The script to run FS with mainnet wasn't working correctly
1. The URL composition to get the sigstructs was incorrect, so it failed
2. The fetched consensus sigstruct wasn't being used (needs to be built into full service )

This PR 
1. Changes the URL composition to correctly download the sigstruct
2. Calls the build script to build FS with the current mainnet consensus sigstruct before launching FS

Depends on #507 , resolves #440

### DIFF
--- a/tools/build-fs.sh
+++ b/tools/build-fs.sh
@@ -38,7 +38,7 @@ WORK_DIR="$HOME/.mobilecoin/${NET}"
 CONSENSUS_DOWNLOAD_LOCATION="$WORK_DIR/consensus-enclave.css"
 mkdir -p ${WORK_DIR}
 
-if ! test -f "$CONSENSUS_DOWNLOAD_LOCATION" && [ $CONSENSUS_SIGSTRUCT_URI != ""]; then
+if ! test -f "$CONSENSUS_DOWNLOAD_LOCATION" && [ "$CONSENSUS_SIGSTRUCT_URI" != "" ]; then
     (cd ${WORK_DIR} && curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${CONSENSUS_SIGSTRUCT_URI})
 fi
 
--- a/tools/run-fs.sh
+++ b/tools/run-fs.sh
@@ -32,7 +32,7 @@ INGEST_DOWNLOAD_LOCATION="$WORK_DIR/ingest-enclave.css"
 mkdir -p ${WORK_DIR}
 
 
-if ! test -f "$INGEST_DOWNLOAD_LOCATION" && [ $INGEST_SIGSTRUCT_URI != ""]; then
+if ! test -f "$INGEST_DOWNLOAD_LOCATION" && [ "$INGEST_SIGSTRUCT_URI" != "" ]; then
     (cd ${WORK_DIR} && curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${INGEST_SIGSTRUCT_URI})
 fi
 


eranrund (changes requested):
Nice, this will be very handy once it works.
I am hoping we can get to a version that works for both testnet and mainnet.

briancorbin (approved):

tools/run-mainnet.sh
```diff
@@ -1,22 +1,39 @@
...
-WORK_DIR="$HOME/.mobilecoin/${NAMESPACE}"
+WORK_DIR="$HOME/.mobilecoin/${NET}"
 WALLET_DB_DIR="${WORK_DIR}/wallet-db"
 LEDGER_DB_DIR="${WORK_DIR}/ledger-db"
-mkdir -p ${WORK_DIR}
 
-(cd ${WORK_DIR} && CONSENSUS_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.${NAMESPACE}.mobilecoin.com/production.json | grep consensus-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
-curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${CONSENSUS_SIGSTRUCT_URI})
+# Pass "--no-build" if the user just wants to run what they have in  
+# .mobilecoin/main instead of building and copying over a new exectuable
```
eranrund, line 11:
```suggestion
# WORK_DIR instead of building and copying over a new exectuable
```

tools/run-mainnet.sh
```diff
@@ -1,22 +1,39 @@
...
+if [ $# -eq 0 ] || [ $1 != "--no-build" ]; then
+    echo "Building"
+    mkdir -p ${WORK_DIR}
+    SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
-(cd ${WORK_DIR} && INGEST_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.${NAMESPACE}.mobilecoin.com/production.json | grep ingest-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
-curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${INGEST_SIGSTRUCT_URI})
+    (cd ${WORK_DIR} && INGEST_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.${NAMESPACE}.mobilecoin.com/production.json | grep ingest-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
+    curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${INGEST_SIGSTRUCT_URI})
+    INGEST_ENCLAVE_CSS="$WORK_DIR/ingest-enclave.css"
```
eranrund, line 19: This doesn't seem related to building full-service, it is related to running it.
Any reason not to always try and download this file if it doesn't exist in the work directory?
That would also remove the need to abort if the user didn't explicitly set an environment variable with it.

eranrund: I don't think we should overwrite files that already exist, but no reason not to download it if it doesn't exist and we know where to get it from

tools/run-mainnet.sh
```diff
@@ -1,22 +1,48 @@
...
+fi
+
+
+if [ -z "$INGEST_ENCLAVE_CSS" ] ; then
+    if ! test -f "$INGEST_DOWNLOAD_LOCATION"; then
+        (cd ${WORK_DIR} && INGEST_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.${NAMESPACE}.mobilecoin.com/production.json | grep ingest-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
+        curl -O https://enclave-distribution.${NAMESPACE}.mobilecoin.com/${INGEST_SIGSTRUCT_URI})
+        INGEST_ENCLAVE_CSS=$INGEST_DOWNLOAD_LOCATION
+    else
+        echo "If running with --no-build, please export a path for the fog-ingest-enclave-css as INGEST_ENCLAVE_CSS"
```
eranrund, line 33: So it looks like if I run the script twice in a row, the behavior would be different. I haven't tested it locally but what I think is going to happen if INGEST_ENCLAVE_CSS is empty, the first time it will download the file because INGEST_DOWNLOAD_LOCATION doesn't exist, but the second time I will get this error since INGEST_DOWNLOAD_LOCATION was created. Is that right or am I missing something?

eranrund: Whats the reason to not have this run script right now?
Looks like a script that differentiates between a few networks would not be substantially more work.

Separately, on the topic of the ingest CSS file. I think it's fine downloading if it doesn't exist and we know where to get it from. I don't think it makes sense for the script to behave differently between runs, where one succeeds and the other one fails due to this else clause.

eranrund:
Please see my reply here https://github.com/mobilecoinofficial/full-service/pull/508#discussion_r992468127
I think there's no reason not to have a run script that supports at the very least both mainnet and testnet, and that it is not acceptable for the script to have different behavior the first time it runs and the second time it runs.

eranrund:
Please let me know when this is ready for review.

eranrund (approved):

PR #509, opened 2022-10-07
Ignore _pycache_ and generated bin folders/files

pouneh:
Ignore locally generated files such as __pycache__ and bin files/folders

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,4 +78,4 @@ dbml-error.log
 
 __pycache__
 
-*bin
+/tools/*.bin


.gitignore
```diff
@@ -75,3 +75,7 @@ cli/build/**
 dbml-error.log
 
 *.profraw
+
+__pycache__
+
+*bin
```
eranrund, line 81: What is generating this? I wonder if it can be narrowed down a bit.

eranrund: If its a single folder then perhaps `/bin` or something of that nature would be more adequate. That other `.bin` file could be ignored just in that directory (e.g. `/tools/*.bin` or something like that)

briancorbin (approved):

PR #510, opened 2022-10-07
Update github actions to consolidate duplicated code for building

briancorbin:
Consolidating duplicated code
Changing the copied binaries for full-service to mc-full-service
Adjusting the release script to reflect these changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,9 +65,9 @@ jobs:
         run: |
           mkdir -pv build_artifacts/${{ matrix.network }}/bin
           cp /var/tmp/*.css build_artifacts/${{ matrix.network }}
-          cp target/release/mc-full-service build_artifacts/${{ matrix.network }}/bin/
-          cp target/release/mc-transaction-signer build_artifacts/${{ matrix.network }}/bin/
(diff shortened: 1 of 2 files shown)


eranrund:
I'm looking at https://github.com/mobilecoinofficial/full-service/actions/runs/3207691446 (which is from the last run) and seeing this:
image

Shouldn't there be a mac arm artifact?

briancorbin:
> I'm looking at https://github.com/mobilecoinofficial/full-service/actions/runs/3207691446 (which is from the last run) and seeing this: image
> 
> Shouldn't there be a mac arm artifact?

I'll run again to double check, but yes there should be

briancorbin:
> What happened to the build job? Looks like this is now only testing one configuration?

renamed it to build_and_pre_release

eranrund:
> > What happened to the build job? Looks like this is now only testing one configuration?
> 
> renamed it to build_and_pre_release

I'm only seeing these checks:
image

Previously this had:
image

briancorbin:
weird, idk, but it definitely ran https://github.com/mobilecoinofficial/full-service/actions/runs/3229524670

maybe something up with github?

briancorbin:
I'll just rerun it

briancorbin:
@eranrund checks showed up!

eranrund (approved):

PR #519, opened 2022-10-12
Python Library

cxloe:
Soundtrack of this PR: [Machine Girl - Crystalline Chute](https://youtu.be/UVa-I9q6xyE)

( old commit history can be seen [here](https://github.com/mobilecoinofficial/full-service/commits/fuller_service_library) )

## Motivation

A readable, friendly async library that encompasses all of the full service endpoints. This is needed for the test suite as well as enabling python devs to quickly incorporate mobilecoin into their applications, and an updated version of mobcli. 

### In this PR
- All of the full service endpoints
- Async request code
- forest_utils for connecting to a remote instance using SSL + easy config file.

Per @briancorbin this should close https://github.com/mobilecoinofficial/full-service/issues/432 ? 

### Future Work

CLI. There is a rough but working version in the working branch ( which is .. messy ) 😅 at https://github.com/mobilecoinofficial/full-service/tree/fuller_service_library/cli/hacky_prototype but a proper cli that takes switch arguments and isn't a slightly modified version of the library is in the cards. 

Following best practices, refactor into classes for organization, maintainability and ease of use.

## Open Questions

Merge [wallet_muncher.py](https://github.com/mobilecoinofficial/full-service/blob/fuller_service_library/cli/wallet_muncher.py) under tools?

Should v1 and v2 endpoints be in separate files? It's more elegant to import this way, but makes for a much longer file. 

Rename the cli folder to imply it's a python library? There's a few folders like python-utils, etc that could also use some cleanup and clearer documentation.

Update the MobileCoin PyPi module? 

Not sure if it needs any more documentation.


cli/fullservice.py
```diff
@@ -0,0 +1,1237 @@
...
+    client.flush()
+
+    ssl_context.load_verify_locations("rootcrt.pem")
+    ssl_context.verify_mode = ssl.CERT_REQUIRED
+    ssl_context.load_cert_chain(certfile="client.full.pem")
+
+
+
+class Request:
+    url = 'http://localhost:9090/wallet/v2'
```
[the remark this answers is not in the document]
briancorbin: Yeah, I second that. Let's have the base url be fully configurable (address:port)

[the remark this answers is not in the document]
cxloe: Meant to do something better here but argparse would hinder importability, could ship it with a config ( dev_secrets perhaps ) file that gets loaded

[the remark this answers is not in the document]
cxloe: Maybe this configs which version of the API to use also..

cxloe:
> Looks pretty good, have a few questions I noted in the conversations portion. Also, what's going to happen to the old folder? Why are we keeping it around? If the plan is to get rid of it, what needs to happen before removal occurs?

I think the qualifier here would be a CLI that works similarly to the old one. But probably something to discuss a little bit.

cli/fullservice.py
```diff
@@ -0,0 +1,1237 @@
...
+    ssl_context.load_cert_chain(certfile="client.full.pem")
+
+
+
+class Request:
+    url = 'http://localhost:9090/wallet/v2'        
+    async def req(self, request_data: dict) -> dict:
+        logging.info("request: %s", request_data.get("method"))
+        request_data['params'] = {k:v for k,v in request_data['params'].items() if v} # handle optional params 
+        response_data = await self.request(request_data)
```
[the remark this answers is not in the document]
cxloe: how do you mean? suggestions appreciated

briancorbin:
There seems to be a huge folder called old, could that be removed?

cli/forest_utils.py
```diff
@@ -0,0 +1,115 @@
+#!/usr/bin/python3.9
+# Copyright (c) 2021 MobileCoin Inc.
+# Copyright (c) 2021 The Forest Team
+import functools
+import logging
+import os
+from typing import Optional, cast, Dict
+
+
+def FuckAiohttp(record: logging.LogRecord) -> bool:
```
briancorbin, line 10: we should probably rename this to something else

cli/forest_utils.py
```diff
@@ -0,0 +1,115 @@
...
+IAS_SPID = get_secret('IAS_SPID')  # 16 bytes
+MOBILECOIN_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '../..', 'mobilecoin'))
+FULLSERVICE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '../..'))
+MOB_RELEASE = get_secret('MOB_RELEASE', '1')
+TARGET_DIR = 'target/release'
+KEY_DIR = os.path.join(MOBILECOIN_DIR, 'target/sample_data/keys')
+WORK_DIR = os.path.join(MOBILECOIN_DIR, TARGET_DIR, 'mc-local-network')
+LEDGER_BASE = os.path.join(MOBILECOIN_DIR, 'target', "sample_data", "ledger")
+MINTING_KEYS_DIR = os.path.join(WORK_DIR, 'minting-keys')
+CLI_PORT = 31337
```
briancorbin, line 104: These look like they are from spinning up a local mobilecoin network, is this all supposed to be here?

cli/fullservice.py
```diff
@@ -0,0 +1,1173 @@
...
+                    "input_txo_ids": input_txo_ids,
+                    "fee_value": fee_value,
+                    "fee_token_id": fee_token_id,
+                    "tombstone_block": tombstone_block,
+                    "max_spendable_value": max_spendable_value,
+                    "comment": comment,
+                },
+            }
+        )
+
```
briancorbin, line 96: what happens when these aren't all set?

briancorbin (changes requested):

cli/forest_utils.py
cxloe: image

fine, fine

cli/forest_utils.py
cxloe: oops, thanks for catching that. more leftovers from refactor_test_utils

cli/fullservice.py
cxloe: should be handled on line 38-39 in the req function, it rebuilds the dictionary every request

briancorbin (approved):

PR #524, opened 2022-10-13
Feature/add mirror

briancorbin:
moving the [mirror service](https://github.com/mobilecoinofficial/full-service-mirror) to this repo and adding it to CI

todo:

- [x] update test script for mirror service


Cargo.toml
```diff
@@ -4,6 +4,7 @@ cargo-features = ["resolver"]
 resolver = "2"
 members = [
     "full-service",
+    "mirror",
```
cxloe, line 7: All of this seems standalone, but perhaps this should be an optional feature? 🤔

briancorbin: not sure what you mean by this?

cxloe: incase someone wants to build FS and isn't going to use the mirror, but not a big deal since it's probably not much extra overhead

briancorbin: that's what the -p flag is for during build, so if you wanted to just specify full-service you would do cargo build -p mc-full-service

cxloe: O

PR #534, opened 2022-10-18
drop txo field from txos table in db

### LINKED ISSUE #431: Drop txo field from txos table in db
This is a redundant field that can be derived from the ledger when needed, as the ledger holds the same info as this column in the database

pouneh:
Resolves #431

### DIFF
--- a/.github/actions-rs/grcov.yml
+++ b/.github/actions-rs/grcov.yml
@@ -1,0 +1,12 @@
+binary-path: ./target/debug/deps/
+branch: true
+ignore:
+  - "../*"
+  - "/*"
+  - "mobilecoin/*"
+  - "target/*"
+  - "*/e2e_tests/*"
+ignore-not-existing: true
+output-type: lcov
+output-path: ./tests.lcov
+source-dir: .
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,6 +90,29 @@ jobs:
           INGEST_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.test.mobilecoin.com/production.json | grep ingest-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
           (cd /var/tmp && curl -O https://enclave-distribution.test.mobilecoin.com/${INGEST_SIGSTRUCT_URI})
 
-      - name: Cargo Test
-        run: |
-          cargo test
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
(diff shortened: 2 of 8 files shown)


full-service/src/service/ledger.rs
```diff
@@ -176,8 +176,7 @@ where
     fn get_txo_object(&self, txo_id_hex: &str) -> Result<TxOut, LedgerServiceError> {
         let conn = self.get_conn()?;
         let txo_details = Txo::get(txo_id_hex, &conn)?;
-
-        let txo: TxOut = mc_util_serial::decode(&txo_details.txo)?;
+        let txo = self.ledger_db.get_tx_out_by_index(self.ledger_db.get_tx_out_index_by_public_key(&txo_details.public_key()?)?)?;
```
briancorbin, line 179: For the sake of readability, I would recommend breaking this up into multiple lines.

In fact, this should probably be refactored into the ledger.rs file as a helper function, since it seems duplicated a few times

full-service/src/test_utils.rs
```diff
@@ -508,16 +478,40 @@ pub fn create_test_received_txo(
...
+    recipient: PublicAddress,
+    value: u64,
+    wallet_db: WalletDb,
+    ledger_db: LedgerDB,
+) -> TxProposal {
+    let (_, txp) =
+        create_test_minted_and_change_txos(src_account_key, recipient, value, wallet_db, ledger_db);
+    txp
+}
+
```
briancorbin, line 504: What is the purpose of having these 2 functions? I would assume whatever is using them could just call the function create_test_minted_and_change_txos and ignore the piece they don't want

full-service/src/test_utils.rs
```diff
@@ -365,12 +333,14 @@ pub fn add_block_with_db_txos(
...
+            let txo = &Txo::get(&txo_id_hex.to_string(), &wallet_db.get_conn().unwrap()).unwrap();
+            ledger_db
+                .get_tx_out_by_index(
+                    ledger_db
+                        .get_tx_out_index_by_public_key(&txo.public_key().unwrap())
+                        .unwrap(),
+                )
+                .unwrap()
         })
         .collect();
```
briancorbin, line 346: This function may not be needed, or even work anymore? Since the txo doesn't exist in the ledger yet, the call to get_tx_out_by_public_key will just fail

briancorbin (changes requested):
This looks great!

briancorbin (approved):

cxloe (approved):

PR #535, opened 2022-10-20
Add code coverage measurements to test script

brucek:

### Motivation

Adding test coverage metrics

### In this PR
* Updated the test.sh script and related docs


eranrund:
Might want to consider https://about.codecov.io/
The SGX repository is using it so there is some reference around it: https://github.com/mobilecoinfoundation/sgx/blob/main/.github/workflows/build.yaml#L220

briancorbin:
> Might want to consider https://about.codecov.io/ The SGX repository is using it and can be used as a reference: https://github.com/mobilecoinfoundation/sgx/blob/main/.github/workflows/build.yaml#L220

Oh that's awesome, thanks for sharing that

briancorbin (changes requested):
This looks great, could we add the codecov.io to this before we merge it?

PR #556, opened 2022-11-02
Fixed to provide default values when offline

briancorbin:
Previously this would error with "NoLastBlockInfo", which was technically correct but not the function we actually want it to do when in offline mode. Now, if it's in offline mode it will respond with default values defined in the mobilecoin library

### DIFF
--- a/full-service/src/service/balance.rs
+++ b/full-service/src/service/balance.rs
@@ -18,11 +18,12 @@ use crate::{
     },
 };
 use displaydoc::Display;
+use mc_blockchain_types::BlockVersion;
 use mc_common::HashMap;
 use mc_connection::{BlockchainConnection, UserTxConnection};
 use mc_fog_report_validation::FogPubkeyResolver;
 use mc_ledger_db::Ledger;
-use mc_transaction_core::TokenId;
+use mc_transaction_core::{tokens::Mob, Token, TokenId};
 
 /// Errors for the Address Service.
 #[derive(Display, Debug)]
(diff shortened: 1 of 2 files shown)


briancorbin:
@eranrund Do you have a suggestion for dealing with eUSD? It doesn't seem to be defined the same as Mob in the mc library. Should we add an implementation for it's default to the mc lib?

Shramp (approved):
lgtm! thanks!

full-service/src/service/balance.rs
```diff
@@ -215,34 +216,47 @@ where
     }
 
     fn get_network_status(&self) -> Result<NetworkStatus, BalanceServiceError> {
+        let (network_block_height, fees, block_version) = match self.offline {
+            true => {
+                let mut fees = BTreeMap::new();
+                fees.insert(Mob::ID, Mob::MINIMUM_FEE);
+                fees.insert(TokenId::from(1), 1024);
```
remoun, line 223: Where does this 1024 come from? 

FWIW [in Moby the eUSD fee is $0.00256, or 2560 micro-eUSD tokens](https://github.com/mobilecoinofficial/mobilecoin-app/blob/4cd4611e6320eeedc6d938cbab87b12e95245cd5/lib/domain/mobilecoin_config.dart#L88)

remoun (approved):

full-service/src/service/balance.rs
briancorbin: Thanks for catching that, it's 1024 by default. Will update to reflect that. Is there anywhere this is defined in the mc library?

remoun: Not that I know of.

PR #585, opened 2022-11-16
Fix codecov yaml file, and explicitly tell it to look at our develop branch as its default

pouneh:
Codecov is broken for our repo. This is the first thing I found that could be fixed. 

Code coverage acts as a simple measure of how effective our tests are. We can use it to identify areas that we could use more confidence in, and make tests to help build confidence and drive quality

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,9 +50,18 @@ jobs:
           INGEST_SIGSTRUCT_URI=$(curl -s https://enclave-distribution.test.mobilecoin.com/production.json | grep ingest-enclave.css | awk '{print $2}' | tr -d \" | tr -d ,)
           (cd /var/tmp && curl -O https://enclave-distribution.test.mobilecoin.com/${INGEST_SIGSTRUCT_URI})
 
+      - name: Cargo sort
+        run: |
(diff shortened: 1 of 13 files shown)


codecov.yml
```diff
@@ -1,13 +1,18 @@
+codecov:
+  branch: develop
```
brucek, line 2: I think this is unnecessary based [on this](https://docs.codecov.com/docs/codecov-yaml#default-branch):

"Codecov will use the default branch from Git for your repository as the primary source of the repositories' Codecov.yml."

ie if the default branch ever changes in git, this would prevent codecov from moving to use that new default branch is how I read it - does that seem right to you?

brucek (changes requested):

codecov.yml
brucek: ok gotcha - that makes sense, thx 👍

brucek (approved):

codecov.yml
```diff
@@ -1,13 +1,18 @@
...
 coverage:
   status:
     project:
-      default: false
+      #default: false
```
cxloe, line 6: curious why this is commented out

cxloe (approved):

PR #589, opened 2022-11-19
GHA deploy and test on PR

jgreat:
### Motivation

Add a CD path for integration testing.  Starting with the goal:  
1. Deploy "simple" full-service to the Development cluster
2. Point full-service at TestNet
3. Run integration tests against full-service.

### In this PR

*`tools/build-fs.sh`*
* bit of a rework on this script to pull down consensus and ingest enclave measurements.
* Allow override of WORK_DIR
* Add build options (BUILD_OPTIONS)
* set the base css repo url and override the index json file so you can specify the mobilecoin release version
  * We are now publishing the latest as `production.json` and a copy with the semver tag `production-v3.0.0.json`.
* set NET to prod if you call `main` - follow up task to replicate the .prod. s3 bucket to .main. so it aligns with chain_id
* follow up - get alpha and other dev builds to publish signed.so and css files.

*mob and .mobconf*
* make .mobconf a link to `mobilecoin/.mobconf`

*docker*
* Create `.internal-ci` path for all the docker/helm and any CI/CD bits that might be required. We use this pattern in mobilecoin and other places. 
* Split docker files into "full-service" that uses pre-built binaries and "full-service-with-build" that does the full build.
* Move dockerfiles to `.internal-ci/docker` and link `.internal-ci/docker/Dockerfile.full-service-with-build` to the root of the repo for "easy mode" builds.
* Update builder and runtime base container images.
* Update containers to use ENV vars instead of command line args
* Add entrypoint script to containers for backwards compatibility for running command line argument style.
* Add build arguments to create a "batteries included" config. 
  * If we need different binaries/containers for test/main/dev... We might as well include the chain_id, peers and tx_source_urls in the build.  You can always customize them at runtime with the appropriate command line args or ENV vars.

*helm*
* Add helm chart for full-service `.internal-ci/helm/full-service`
* includes backups/restore sidecar - discuss if we want this here and making this public.

*GitHub Actions*
* Some small cleanups in existing workflows to pass `actionlint`
* fixed??? some misaligned cache action output variables in existing workflows.  
  * Need :eyes: on this, I think these changes reflect what's intended here.
* add standard caching actions
* dev-cd workflow
  * build code - with caching 
  * build and publish docker container image and helm chart.
  * create namespace in the development cluster and deploy.
  * TODO: add tests. Placeholder python with example poetry environment setup.

### Future Work
* Caching actions are copies from mobilecoin, we probably want to make extract these into a "actions library" repo so we can keep it up to date in one place and make it more reusable.  
* Replicate `prod` css repo bucket to a `main` bucket to align with chain_id
* put poetry install in builder-install container


.github/workflows/build.yml
```diff
@@ -24,9 +24,9 @@ jobs:
         namespace: [test, prod]
         include:
           - runner-tags: [self-hosted, macOS, X64, cargo]
-            container: null
+            container: ""
```
jgreat, line 27: `actionlint` fix:  null isn't a valid value for matrix definition

.internal-ci/test/fs-integration/pyproject.toml
```diff
@@ -0,0 +1,19 @@
+[tool.poetry]
```
cxloe, line 1: Is there a way to keep this file current with the one in the main FS directory?

.github/workflows/dev-cd.yaml
```diff
@@ -0,0 +1,217 @@
...
+        curl -sSL https://install.python-poetry.org | python3 -
+
+    - name: Test full-service - TestNet
+      run: |
+        # Switch to testing directory and install dependencies.
+        pushd .internal-ci/test/fs-integration || exit 1
+        "${POETRY_HOME}/bin/poetry" install --without=dev
+
+        # Run test
+        "${POETRY_HOME}/bin/poetry" run ./basic.py
```
cxloe, line 201: This should be `[poetry] run python3 ./basic.py`

.internal-ci/test/fs-integration/pyproject.toml
jgreat: Check out the latest changes and the README I included as part of the  `.internal-ci/test/fs-integration` directory for setting up the environment.  We should move the work that's in `ineg-tests` in to this directory.  We can use the same `poetry` project for multiple test scripts if we want.

As full-service is not a python project, I don't think the testing frameworks dependencies or setups should be part of the main repo structure. 

In the longer term, I think we should move/rename the `cli` directory and possibly clean up or delete the `python-utils`.  Some of the tools that get reused here should get broken out into a separate repo and properly versioned instead of copied around to various projects.

jgreat: Check out the latest changes and the README I included as part of the  `.internal-ci/test/fs-integration` directory for setting up the environment.  We should move the work that's in `ineg-tests` in to this directory.  We can use the same `poetry` project for multiple test scripts if we want.

As full-service is not a python project, I don't think the testing frameworks dependencies or setups should be part of the main repo structure. 

In the longer term, I think we should move/rename the `cli` directory and possibly clean up or delete the `python-utils`.  Some of the tools that get reused here should get broken out into a separate repo and properly versioned instead of copied around to various projects.

.github/workflows/dev-cd.yaml
```diff
@@ -0,0 +1,235 @@
...
+        chart_version: ${{ needs.metadata.outputs.tag }}.test
+        chart_set: |
+          --set=persistence.enabled=false
+        release_name: full-service
+        namespace: ${{ needs.metadata.outputs.namespace }}
+        rancher_cluster: ${{ secrets.DEV_RANCHER_CLUSTER }}
+        rancher_url: ${{ secrets.DEV_RANCHER_URL }}
+        rancher_token: ${{ secrets.DEV_RANCHER_TOKEN }}
+
+  # TODO: Add integration tests.
```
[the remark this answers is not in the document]
jgreat: Sure,  I know the python scripts are a work in progress, so I just left a placeholder here until they are ready to go.

.github/workflows/release.yml
```diff
@@ -71,6 +71,6 @@ jobs:
         uses: softprops/action-gh-release@v1
         with:
           draft: true
-          prerelease: ${{ steps.prerelease.outputs.value }}
+          prerelease: false
```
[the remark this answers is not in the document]
jgreat: At the moment there is no step with the id of `prerelease` so I just set it to the value that's always going to be assigned.  I think we can do a follow up on this PR, discuss what we want the workflow to look like and then we can come back and implement it.

PR #595, opened 2022-11-22
prettyprint the response from full service

cxloe:
Quick little quality of life improvement: This makes the response from FS much more readable in the Python Library, using the fantastic [rich](https://pypi.org/project/rich/) library.

### DIFF
--- a/cli/fullservice.py
+++ b/cli/fullservice.py
@@ -64,9 +64,6 @@ class Request:
 
 
 class FullServiceAPIv2(Request):
-    def __init__(self, logLevel=logging.ERROR):
-        logging.basicConfig( level=logLevel)
-
     async def assign_address_for_account(self, account_id, metadata=""):
         return await self.req(
             {
--- a/cli/fullservice.py
+++ b/cli/fullservice.py
@@ -8,6 +8,8 @@ import base64
 from typing import Optional
 import ssl
 import forest_utils as utils
+from rich import print_json
+
 
 if not utils.get_secret("ROOTCRT"):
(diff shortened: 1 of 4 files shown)


cli/fullservice.py
```diff
@@ -56,7 +58,7 @@ async def request(self, request_data: dict):
                     data=json.dumps(request_data),
                     headers={"Content-Type": "application/json"},
                 ) as resp:
-                    # print(resp.json)
+                    print_json(data=await resp.json(), indent=4)
```
brucek, line 61: It seems like if this was a logging item before, I would think it should remain as a logging item?

So:
1. the output actually gets added into logs
2. esp if you want to redirect / save / quiet logs via `logging`

Or am I misunderstanding the python idioms here?

Seems [somewhat straightforward](https://rich.readthedocs.io/en/stable/console.html?highlight=print_json#printing-json) IIUC

Also, is there an advantage to moving the output here from lines 46-47?

cli/pyproject.toml
```diff
@@ -9,6 +9,7 @@ license = "GPLv3"
 python = "^3.10"
 asyncio = "^3.4.3"
 aiohttp = "^3.8.3"
+rich = "^12.6.0"
```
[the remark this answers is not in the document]
cxloe: Rich works on 3.6.3 or later

cli/fullservice.py
cxloe: Cleaned this up a bit! Now it plays nicer with `logging` and prettyprints if the log level is set to `DEBUG`

PR #598, opened 2022-11-23
Exit codes for GHA

cxloe:


integ-tests/basic.py
```diff
@@ -50,6 +50,8 @@ async def get_account(i):
 
 
 async def main():
+    while (await fs.get_wallet_status())['result']['wallet_status']['is_synced_all'] != True:
+        await asyncio.sleep(5)
```
brucek, line 54: Is the `fs.get_wallet_status` call expensive?  Maybe we can lower this sleep to more like 1 or 0.5?

PR #599, opened 2022-11-24
test config as cmdline param

pouneh:
allow the test caller to specify a different config file

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3653,6 +3653,7 @@ dependencies = [
  "mc-util-uri",
  "mc-validator-api",
  "rayon",
+ "structopt",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


integ-tests/basic.py
```diff
@@ -111,4 +112,12 @@ async def does_it_go(amount_pmob: int = 5) -> bool:
 
 
 if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Basic test")
+    parser.add_argument("config_path", type=str, default=default_config_path)
```
cxloe, line 116: This should be optional so the default option actually comes into play

jgreat (approved):
Was just working on something like this, you beat me to it. LGTM

integ-tests/basic.py
```diff
@@ -111,4 +112,13 @@ async def does_it_go(amount_pmob: int = 5) -> bool:
 
 
 if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Basic test")
+    parser.add_argument("config_path", type=str, default=default_config_path, 
+                        required=False)
+    args = parser.parse_args()
+
+    print(args.__dict__)
```
cxloe, line 120: what is this print for?

PR #600, opened 2022-11-24
Add public_key to OutputTxo

plarson:
### Motivation

http://www.pivotaltracker.com/slack_links/story/183842633

Public key is often used to identify TXOs by ramps and it simplifies ramp code to include the output public key from `build_and_submit_transaction`.

### In this PR
* Add `public_key` to `OutputTxo` json

### DIFF
--- a/full-service/src/json_rpc/v2/models/transaction_log.rs
+++ b/full-service/src/json_rpc/v2/models/transaction_log.rs
@@ -126,6 +126,8 @@ impl InputTxo {
 pub struct OutputTxo {
     pub txo_id_hex: String,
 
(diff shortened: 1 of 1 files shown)


briancorbin (changes requested):
Would you mind updating the e2e tests to reflect this change? If your not familiar with our unit testing setup I'd be happy to do it :)

plarson:
> Would you mind updating the e2e tests to reflect this change? If your not familiar with our unit testing setup I'd be happy to do it :)

@briancorbin which files is that in?

briancorbin (approved):

PR #610, opened 2022-11-30
Move basic.py to where ci-cd will pick it up from

pouneh:
### Motivation
We want to ensure we do quality checks on ci by running integration tests

### In this PR
The ci-cd pipeline currently uses a sparse test. Moving the most up-to-date integ test to where ci reads the tests it uses.  
 Addresses PT#183925740

### Test Plan
You can run this test by navigating to the directory and running `poetry run python3 basic.py`
Mostly, testing this change is seeing the test run successfully on CI as part of this PR

### DIFF
--- a/.internal-ci/test/fs-integration/basic.py
+++ b/.internal-ci/test/fs-integration/basic.py
@@ -10,7 +10,7 @@ sys.path.append("{}/python-library".format(repo_root_dir))
 from fullservice import FullServiceAPIv2 as v2
 from FSDataObjects import Response, Account 
 
-sleepy_time = 5
+sleepy_time = 15
 default_config_path = "./test_config.json"
 config = []
 account_ids = []
@@ -106,10 +106,10 @@ async def does_it_go(amount_pmob: int = 600000000) -> bool:
     """Test Setup """
     pmob_to_send = amount_pmob
 
-    await preclean_this_test()
+    # await preclean_this_test()
 
-    alice = await get_account(0, "alice")
-    bob = await get_account(1, "bob")
+    alice = await get_account(0, "alice", True)
+    bob = await get_account(1, "bob", True)
 
     await wait_for_account_to_sync(alice.id)
     await wait_for_account_to_sync(bob.id)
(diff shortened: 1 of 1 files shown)


cxloe (approved):

cxloe:
maybe this should be a symlink instead. we could have the integ-tests folder in python-library, then symlink it to the CI dir.

brucek:
> You can run this test by navigating to the directory and running `poetry python3 basic.py`

2 Q's for you here:
1. I seem to need to run `poetry run python3 basic.py`, am I missing something, or is the description here?
2. Is there a sample config file somewhere?  I also can't seem to run locally without one (I can add my own mnemonics if needed)

.internal-ci/test/fs-integration/test_config.json
```diff
@@ -0,0 +1,10 @@
+{
```
jgreat, line 1: Just a question.  Did you intend this to be an example template?  I see that this file is listed in the .gitignore

jgreat (approved):
LGTM.   Good catch on the docs, not sure how I forgot to put the `poetry install` step in the README.

jgreat:
> maybe this should be a symlink instead. we could have the integ-tests folder in python-library, then symlink it to the CI dir.
> 
> i wonder if this approach would work?

I think having the integration tests in the `test` folder makes sense here.   Its not a library. 

For ease of use and only having to maintain a single set of dependencies,  we can have multiple scripts/tests in this `fs-integration` dir. 

Also breaking it up into other directories or renaming it under the `test` dir is fine.

I kinda think that if we want tests that are integrated with the main repo structure, we should probably write them as rust utilities.

cxloe:
> > 
> 
> mm I'm not sure how to do the symlinking in a good way .. like who creates the symlink/maintains it?

a symlink is just a file that points to another file.. there's not anything to maintain?

cxloe:
> > 
> 
> Do people have to set up that symlink manually on their machine, if not we're checking it in? Can we even checkin a symlink? If we do, will we hit snags when people clone this across different machines? There's just too many questions I don't know how to answer .. and I think fidling til I find out the answer isn't ... really what I want to do with this PR :(

You can check in a symlink. It should work as long as it's not pointing all the way from a specific home dir.. so you'd just want to move back a few directories instead of doing something like `ln -s /home/user/workspace/some-repo/file .`

PR #624, opened 2022-12-06
Patch a race condition in the wait for account to sync 

cxloe:
### Motivation

There's a problem with this function where it does not reliably sleep for long enough for a transaction to complete. 

### In this PR
Just sleeps for another 15 seconds. 

### Future Work
* [Refactor that function entirely](https://www.pivotaltracker.com/story/show/183966533)

### DIFF
--- a/.internal-ci/test/fs-integration/basic.py
+++ b/.internal-ci/test/fs-integration/basic.py
@@ -24,6 +24,7 @@ async def wait_for_account_to_sync(id):
            != account_status.get("result").get("local_block_height")) or (account_status.get("result").get("balance_per_token").get("0").get("pending") != "0")):
         await asyncio.sleep(sleepy_time)
         account_status = await fs.get_account_status(id)
+    await asyncio.sleep(sleepy_time)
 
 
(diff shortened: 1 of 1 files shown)


.internal-ci/test/fs-integration/basic.py
```diff
@@ -24,6 +24,7 @@ async def wait_for_account_to_sync(id):
            != account_status.get("result").get("local_block_height")) or (account_status.get("result").get("balance_per_token").get("0").get("pending") != "0")):
         await asyncio.sleep(sleepy_time)
         account_status = await fs.get_account_status(id)
+    await asyncio.sleep(sleepy_time)
```
queenbdubs, line 27: why not just increase the value of `sleepy_time` used on line 25?

cxloe: For one i'm not sure if the conditions are right in that that `while` loop.. It's definitely hacky but having the sleep at the end of the function makes sure it sleeps for at **least** 15sec.. since on line 144-145 it's being awaited twice in a row.

PR #625, opened 2022-12-06
Simple test for create_account

cxloe:
### In this PR
Simple test that gets the number of accounts, creates one, and checks to make sure the number of accounts increased.

### Future Work
* [We will have to orchestrate the entire test suite as a whole. ](https://www.pivotaltracker.com/story/show/183966740)


cxloe:
@briancorbin im assuming this is out of scope since it's covered in the unit tests?

PR #626, opened 2022-12-07
Feature/track build signed submitted for tx log

briancorbin:
### Motivation

Previously, we haven't been tracking transactions that were built but not signed. In order to support view only accounts with the offline signer and hardware wallet better, this adds that so we can track which inputs and outputs are for the built, but unsigned, transaction.

### In this PR
* Adding logging for built but unsigned transactions
* fixing bug where ```tombstone_block_index``` wasn't updated correctly if transaction log exists when submitting
* Changing how TransactionId is derived to allow it to be the same from UnsignedTx -> Tx (Using TxPrefix now)

### Future Work
Continued changes to support hardware wallet and offline transaction signing

Moving TransactionId, TxoId, and AccountId to the mobilecoin library to provide a central source of truth for these

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -25,7 +25,7 @@ use crate::{
     service::models::tx_proposal::{TxProposal, UnsignedTxProposal},
 };
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct TransactionId(pub String);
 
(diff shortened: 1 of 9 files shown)


cxloe:
Should this logging be configurable / disableable?

And why the change of `TransactionId' to `TransactionID` ?

cxloe (approved):

PR #633, opened 2022-12-12
Transaction signer build script

cxloe:
### Motivation
Build script to easily get the transaction signer set up.

### In this PR
- `build-signer.sh` script, built off of `build-fs.sh`


queenbdubs (approved):
I'd love to have Brian take a glance too

PR #634, opened 2022-12-13
check for xcode 12

cxloe:
### Motivation

If FS is built with Xcode 13 or newer, this leads to lots of `BadSignature` errors which can be frustrating. Now the build script checks if the version is 12, and prompts to switch if not.


tools/.shared-functions.sh
```diff
@@ -44,3 +43,20 @@ get_css_file()
...
     echo "${css_file}"
 }
+
+
+check_xcode() {
+  if [[ $(uname) == "Darwin" ]]; then
+    xcode_version=$(xcodebuild -version | grep "Xcode" | awk '{print $2}' | sed 's/[.].*//')
+    minimum_xcode_version=12
+    if [[ $(printf '%s\n' "$xcode_version" "$minimum_xcode_version" | sort -V | head -n1) == "$xcode_version" ]]; then
+      echo
```
brucek, line 53: According to the description of the issue, I think anything other than XCode 12 should fail here, right?
1. maybe `minimum_xcode_version` should be something more like `supported_xcode_version`?
2. This test as written passes with XCode 13 & 14, which it should not, correct?

brucek (changes requested):

eranrund:
@the-real-adammork are your changes to mbedtls supposed to remedy the issue with the newer xcode versions?

tools/.shared-functions.sh
```diff
@@ -44,3 +43,20 @@ get_css_file()
     debug "  css file saved ${css_file}"
     echo "${css_file}"
 }
+
+
+check_xcode() {
+  if [[ $(uname) == "Darwin" ]]; then
+    xcode_version=$(xcodebuild -version | grep "Xcode" | awk '{print $2}' | sed 's/[.].*//')
+    minimum_xcode_version=12
+    if [[ $(printf '%s\n' "$xcode_version" "$minimum_xcode_version" | sort -V | head -n1) == "$xcode_version" ]]; then
```
eranrund, line 52: Nit: If the `if` clause is empty and the `else` code has the logic, it means the condition the `if` is testing for needs to be reversed

the-real-adammork:
> @the-real-adammork are your changes to mbedtls supposed to remedy the issue with the newer xcode versions?

yes it fixes the issue with xcode 13/14. I got stuck trying to fix the mac-catalyst target for libmobilecoin, but fs doesnt need that so Ill make a mobilecoin PR with what I have.

the-real-adammork:
> > @the-real-adammork are your changes to mbedtls supposed to remedy the issue with the newer xcode versions?
> 
> yes it fixes the issue with xcode 13/14. I got stuck trying to fix the mac-catalyst target for libmobilecoin, but fs doesnt need that so Ill make a mobilecoin PR with what I have.

PR - https://github.com/mobilecoinfoundation/mobilecoin/pull/2962

briancorbin:
Not necessary as issue is resolved with PR - https://github.com/mobilecoinfoundation/mobilecoin/pull/2962

PR #639, opened 2022-12-14
Transaction Signer Support for Integration Tests

cxloe:
### Motivation

This makes the transaction signer natively available to the integration tests. I'm not sure if this approach is relevant since the signer seems like it may be folded into FS itself? @briancorbin 

### In this PR

- Transaction signer split into `lib.rs` and `main.rs `
- Build script that installs the signer
- pyo3 deps, decorators and a wrapper function to expose needed parts to the compiler.


tools/build-signer-maturin.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
```
eranrund, line 1: What is this file? How is it different from tools/build-signer.sh?

tools/build-signer.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
```
eranrund, line 1: It looks like a lot of stuff was copied from https://github.com/mobilecoinofficial/full-service/blob/zd/transaction-signer-lib-pyo3/tools/build-fs.sh
What is the reason for not having a single script that builds all the binaries in this repo?

transaction-signer/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+    get_tx_out_shared_secret,
+    onetime_keys::{recover_onetime_private_key, recover_public_subaddress_spend_key},
+    ring_signature::KeyImage,
+    tx::TxOut,
+};
+use std::{convert::TryFrom, fs};
+use pyo3::prelude::*;
+use pyo3::wrap_pyfunction;
+
+#[pymodule]
```
eranrund, line 34: I don't think we should be testing Rust code using something that exposes it to Python.
If the goal is to call actual rust functions for tests, the test should be written in Rust.
If the goal is to test the Rust code as a block box, the Python code should run the binary like regular users are expected to run it.
Nothing is currently expected to use these bindings, so exercising them in a test is in my opinion a questionable approach and I'm not sure how much work it saves us, compared to just running the binary.

eranrund (changes requested):

transaction-signer/src/lib.rs
cxloe: yeah i just took the simpler approach and am about to PR that

PR #644, opened 2022-12-17
removing account syncing requiring regenerating subaddress keys

briancorbin:
This update was necessary to increase account sync speed for accounts with large amounts of subaddresses, as well as not causing the API to lock when syncing any account. The main fix is that we are no longer using a transaction nor pre-generating all of the subaddress keys for each iteration, since the subaddress keys already exist in the database and we can simply perform a SQL query to match against them.

### DIFF
--- a/full-service/migrations/2023-01-20-231044_add_index_to_subaddress_spend_key/down.sql
+++ b/full-service/migrations/2023-01-20-231044_add_index_to_subaddress_spend_key/down.sql
@@ -1,0 +1,1 @@
+-- This file should undo anything in `up.sql`
--- a/full-service/migrations/2023-01-20-231044_add_index_to_subaddress_spend_key/down.sql
+++ b/full-service/migrations/2023-01-20-231044_add_index_to_subaddress_spend_key/down.sql
@@ -1,1 +1,1 @@
--- This file should undo anything in `up.sql`
+DROP INDEX idx_assigned_subaddresses__spend_public_key;
--- a/full-service/src/service/sync.rs
+++ b/full-service/src/service/sync.rs
@@ -110,20 +110,18 @@ pub fn sync_all_accounts(
         .num_blocks()
         .expect("failed getting number of blocks");
 
+    let conn = wallet_db.get_conn()?;
+
     // Go over our list of accounts and see which ones need to process more blocks.
-    let accounts: Vec<Account> = {
-        let conn = &wallet_db
-            .get_conn()
-            .expect("Could not get connection to DB");
-        Account::list_all(conn, None, None).expect("Failed getting accounts from database")
-    };
+    let accounts: Vec<Account> =
+        { Account::list_all(&conn, None, None).expect("Failed getting accounts from database") };
 
     for account in accounts {
         // If there are no new blocks for this account, don't do anything.
         if account.next_block_index as u64 > num_blocks - 1 {
             continue;
         }
-        sync_account(ledger_db, wallet_db, &account.id, logger)?;
+        sync_account(ledger_db, &conn, &account.id, logger)?;
     }
 
     Ok(())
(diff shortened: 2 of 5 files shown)


full-service/src/service/sync.rs
```diff
@@ -143,141 +142,132 @@ pub fn sync_account(
...
+            .into_par_iter()
+            .filter_map(|(block_index, tx_out)| {
+                let amount = match decode_amount(&tx_out, view_account_key.view_private_key()) {
+                    None => return None,
+                    Some(a) => a,
+                };
+                let subaddress_index = decode_subaddress_index(
+                    &tx_out,
+                    view_account_key.view_private_key(),
+                    &wallet_db.get_conn().unwrap(),
```
eranrund, line 217: This should not call unwrap, it should propagate the error

eranrund (changes requested):

full-service/src/service/sync.rs
briancorbin: Thanks for catching, leftover from draft

full-service/src/service/sync.rs
```diff
@@ -143,141 +142,132 @@ pub fn sync_account(
...
-        sync_account_next_chunk(ledger_db, &conn, logger, account_id_hex)?
+        sync_account_next_chunk(ledger_db, wallet_db, logger, account_id_hex)?
     {}
 
     Ok(())
 }
 
 fn sync_account_next_chunk(
     ledger_db: &LedgerDB,
-    conn: &Conn,
```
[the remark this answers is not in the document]
briancorbin: This is because the function to decode a subaddress index now requires a database connection, which is not thread safe, so in order to maintain the `into_par_iter().filter_map(..)` when decoding the individual received txos, I pass the wallet_db, which is thread safe, and pool for the connection when required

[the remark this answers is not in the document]
briancorbin: Was necessary until a refactor, so it can probably just go back so we don't worry about connection failures during syncing a chunk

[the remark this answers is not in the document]
eranrund: I believe the biggest advantage for using the same connection during sync is that it allows you to execute the entire block sync in a single SQL transaction, so you can't end up with a partially synced block if something fails midway.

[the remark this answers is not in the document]
briancorbin: right, okay, lemme fix the transaction chunk, too

full-service/migrations/2023-01-20-231044_add_index_to_subaddress_spend_key/down.sql
```diff
@@ -0,0 +1 @@
+-- This file should undo anything in `up.sql`
```
eranrund, line 1: Shouldn't it drop the index?

full-service/src/service/sync.rs
```diff
@@ -409,9 +415,16 @@ pub fn decode_subaddress_index(
         Err(_) => return None,
     };
 
-    let subaddress_spk: RistrettoPublic =
+    let subaddress_spend_public_key: RistrettoPublic =
         recover_public_subaddress_spend_key(view_private_key, &tx_out_target_key, &tx_public_key);
-    subaddress_keys.get(&subaddress_spk).copied()
+
+    match AssignedSubaddress::find_by_subaddress_spend_public_key(
```
eranrund, line 421: Nit: No need for a match statement, this can be `.ok().map(|idx, _b58| idx)`

eranrund (approved):

full-service/migrations/2023-01-20-231044_add_index_to_subaddress_spend_key/down.sql
briancorbin: Ideally yes, which I can add to this branch, but we haven't been adding steps for downgrading the database. Might be worth going back and adding them all.

PR #650, opened 2023-01-03
Using openssl-vendored with grpcio

briancorbin:
This resolves an issue where a user doesn't have the correct version of openssl installed on their machine by using a bundled version of openssl instead of the system version (if available)


eranrund (approved):

eranrund:
Good idea but looks like something else is linking the system openssl and resulting in all the duplicate symbols... not sure what the solution for that would be but the first step is probably figuring out which crate is doing that.

PR #654, opened 2023-01-07
Bugfix/input txos too large

briancorbin:
Fixes a bug where inputs were being restricted to < u64 max total, which is not necessary


full-service/src/service/transaction_builder.rs
```diff
@@ -726,6 +722,67 @@ mod tests {
...
+            .add_recipient(recipient.clone(), 7_000_000 * MOB, Mob::ID)
+            .unwrap();
+        builder
+            .add_recipient(recipient.clone(), 7_000_000 * MOB, Mob::ID)
+            .unwrap();
+        builder
+            .add_recipient(recipient.clone(), 7_000_000 * MOB, Mob::ID)
+            .unwrap();
+
+        // This is the value that will overflow u64, which should be valid.
```
brucek, line 777: Maybe also worth a check here that 7_000_000 * MOB * 3 is > u64::MAX ?

brucek (approved):

eranrund:
I think this needs more more tests, including complete transaction building. Wrap-around issues are hard to detect.
For example, I can see https://github.com/mobilecoinofficial/full-service/blob/bf0fe688f4828cdf554ff35907a19b50e6ed9b9d/full-service/src/db/txo.rs#L1200 being a problem - since the values are i64s in the database, I am not sure sorting will behave correctly. A potential workaround is to sort in Rust, once the value is converted to u64.
In a bunch of places we are using https://github.com/mobilecoinofficial/full-service/blob/bf0fe688f4828cdf554ff35907a19b50e6ed9b9d/full-service/src/db/models.rs#L100 directly, without casting to u64, so that is also a potential issue.

Some example tests might include:
1) Two input TXOs `u64::MAX - 10`, `20`, and an outlay of 30 - verifying that the change and outlay ends up being correct (ideally by actually decrypting the masked amount in the real constructed Tx)
2) A reverse of that, where the outlay is `u64::MAX`, as well as a `u64::MAX - 1` version. 
3) A similar version with multiple outlays

Imbalanced transactions will get rejected by consensus so that scenario is less catastrophic, but we should be very very certain that under no circumstance we end up paying more than intended due to some wraparound bug.

eranrund (changes requested):
Marking as changes requested since I think we need to really have a lot of tests around this that we currently do not. Too much mixing of i64/u64, sometimes with casting and sometimes without, letting SQLite do the sorting (as i64s), unchecked math (e.g. https://github.com/mobilecoinofficial/full-service/blob/bf0fe688f4828cdf554ff35907a19b50e6ed9b9d/full-service/src/db/txo.rs#L1291).

IMO this needs some pretty significant test coverage we don’t currently have. From unit tests around stuff like the sorting, input selection, change calculation, adding fees math, etc all the way to end to end that ensure the generated Tx is correct

briancorbin:
> 

Agreed, will work on building those tests right now

briancorbin:
Current dependencies on/for this PR:
* develop
  * **PR #655** <a href="https://app.graphite.dev/github/pr/mobilecoinofficial/full-service/655" target="_blank">Graphite</a> 
    * **PR #654** <a href="https://app.graphite.dev/github/pr/mobilecoinofficial/full-service/654" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinofficial/full-service/654?utm_source=stack-comment).

briancorbin:
Dupe of #654

eranrund:
> Dupe of #654

This is #654...

PR #658, opened 2023-01-12
Update mbedTLS to solve for Xcode 13 & 14 compatability

the-real-adammork:
### Motivation

`mobilecoin` core would have runtime issues when compiled on macOS w/ Xcode versions 13+. The reason for this is best explained in the [patch in mbedTLS](https://github.com/mobilecoinfoundation/rust-mbedtls/pull/66) which was the root cause area of the problem.

This PR brings that patch full-service.

### In this PR
* Patch rust-mbedtls to latest for xcode fix patch
* Remove Xcode version disclaimers in README.md
* Update submodule commit for `mobilecoin` repo

### Future Work

Update the submodule commit to point to a commit in `master` on the mobilecoin repo, right now it points to a commit in a branch. I will update this PR after that branch merges to master.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,36 +27,16 @@ dependencies = [
 ]
 
 [[package]]
-name = "aead"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c192eb8f11fc081b0fe4259ba5af04217d4e0faddd02417310a927911abd7c8"
-dependencies = [
- "crypto-common",
- "generic-array",
-]
-
-[[package]]
 name = "aes"
 version = "0.7.5"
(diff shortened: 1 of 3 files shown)


briancorbin:
tests look good, only grcov that's failing

briancorbin:
Trying this out on my machine with Xcode 14, will update momentarily

the-real-adammork:
ok, I did do as I suggested above, so the submodule commit is what it was + one commit with my `mbedtls` changes.

the-real-adammork:
> Trying this out on my machine with Xcode 14, will update momentarily

rad

briancorbin (approved):
awesome, looks and works great! I will merge if ready

briancorbin:
Actually, will use this as reference and will update full service once these changes make it into a release branch of the mobilecoin library. Thank you!

PR #660, opened 2023-01-12
Hotfix/input txos too large

briancorbin:
Fixes a bug where inputs were being restricted to < u64 max total, which is not necessary. Also fixes several bugs related to u64 conversions and generation of change txos.

## Todo

- Currently we only support change txos that are of size u64::MAX or less. Unsure if we should create as many as necessary, or force users to be explicit about their transaction inputs/outputs if the auto builder cant pick them normally


full-service/src/db/txo.rs
```diff
@@ -1815,6 +1822,79 @@ mod tests {
...
+            &wallet_db,
+        );
+
+        // Create some small TXOs for the account
+        // [100, 200, 300, ... 2000]
+        for i in 1..20 {
+            let (_txo_hex, _txo, _key_image) = create_test_received_txo(
+                &account_key,
+                0,
+                Amount::new(100, Mob::ID), // 100.0 MOB * i
```
eranrund, line 1873: This isn't multiplying by `i`

full-service/src/db/txo.rs
```diff
@@ -1815,6 +1822,79 @@ mod tests {
...
+            0,
+            Amount::new(u64::MAX, Mob::ID),
+            146,
+            &mut rng,
+            &wallet_db,
+        );
+
+        // Create some small TXOs for the account
+        // [100, 200, 300, ... 2000]
+        for i in 1..20 {
```
eranrund, line 1869: Should probably be 1..=20

full-service/src/db/txo.rs
```diff
@@ -1815,6 +1822,79 @@ mod tests {
...
+            Some(&account_id_hex.to_string()),
+            None,
+            None,
+            0,
+            Mob::MINIMUM_FEE,
+            &conn,
+        )
+        .unwrap();
+
+        // Max spendable SHOULD be u64::MAX * 2 + (100 * 14) - Mob::MINIMUM_FEE =
```
eranrund, line 1890: If the loop above created txos in the range 100-2000 then the max should be u64::max * 2 + 21000 - fee

full-service/src/service/transaction_builder.rs
```diff
@@ -726,6 +714,60 @@ mod tests {
...
+            .add_recipient(recipient.clone(), u64::MAX, Mob::ID)
+            .unwrap();
+
+        // This is the value that will overflow u64, which should be valid.
+        builder
+            .set_txos(
+                &conn,
+                &vec![txos[0].id.clone(), txos[1].id.clone(), txos[2].id.clone()],
+            )
+            .unwrap();
```
eranrund, line 768: Is it possible to fully construct a transaction with large inputs and see that all the outputs are correct?

full-service/src/service/transaction_builder.rs
```diff
@@ -1093,11 +1135,9 @@ mod tests {
...
             .unwrap();
 
-        match builder.select_txos(&wallet_db.get_conn().unwrap(), None) {
-            Ok(_) => panic!("Should not be able to select txos with > u64::MAX output value"),
-            Err(WalletTransactionBuilderError::OutboundValueTooLarge) => {}
-            Err(e) => panic!("Unexpected error {:?}", e),
-        }
+        builder
+            .select_txos(&wallet_db.get_conn().unwrap(), None)
+            .unwrap();
```
eranrund, line 1140: Would be nice to verify that the transaction that gets built actually contains the expected output values and change.

eranrund (changes requested):
Nice work, thanks for iterating on this. I have some suggestions on stricter tests.

full-service/src/db/txo.rs
```diff
@@ -1815,6 +1822,79 @@ mod tests {
...
```
eranrund, line 1890: Why is it times 14 and not times 20?

full-service/src/service/transaction_builder.rs
```diff
@@ -726,6 +770,154 @@ mod tests {
...
+                &conn,
+                &vec![txos[0].id.clone(), txos[1].id.clone(), txos[2].id.clone()],
+            )
+            .unwrap();
+
+        // NOTE: We have to use an Empty memo here because the RTH memo will fail
+        // because of a u64::MAX limit on the total_outlay value.
+        // See https://github.com/mobilecoinfoundation/mobilecoin/blob/437133a545b85958278efcb655bce36929c8f72a/transaction/extra/src/memo/destination_with_payment_request_id.rs#L51
+        // for more details.
+        let _unsigned_tx_proposal = builder.build(TransactionMemo::Empty, &conn).unwrap();
```
eranrund, line 834: Should validate the change and outputs are correct

full-service/src/db/txo.rs
briancorbin: only 16 can be used as inputs, so Max spendable is the largest 16 txos

briancorbin: minus the fee

eranrund: Can you please add that to the comment?

full-service/src/service/transaction_builder.rs
```diff
@@ -831,7 +831,34 @@ mod tests {
...
+        );
+        assert_eq!(
+            unsigned_tx_proposal.unsigned_input_txos[2].amount.value,
+            u64::MAX
+        );
+
+        // Check that the payload txos are correct
+        assert_eq!(unsigned_tx_proposal.payload_txos.len(), 2);
+        assert_eq!(unsigned_tx_proposal.payload_txos[0].amount.value, u64::MAX);
+        assert_eq!(unsigned_tx_proposal.payload_txos[0].amount.value, u64::MAX);
```
eranrund, line 854: This should check `[1]`

full-service/src/service/transaction_builder.rs
eranrund: Why was this marked as resolved?

full-service/src/service/transaction_builder.rs
eranrund: Why was this marked as resolved?

eranrund (commented):
Still some pending comments. Thank you!

full-service/src/service/transaction_builder.rs
briancorbin: I've updated that to check the change and outputs

full-service/src/service/transaction_builder.rs
briancorbin: I've expanded the test to include the suggestion

full-service/src/service/transaction_builder.rs
briancorbin: ah thanks for the catch

eranrund (approved):

full-service/src/service/transaction_builder.rs
briancorbin: Just kidding, this test can be removed (and was now) because it was covered in a different (new) test

PR #668, opened 2023-01-20
Add endpoints for getting recent blocks and range of blocks

eranrund:
### Motivation

These endpoints are needed for implementing a block explorer. In theory the `get_recent_blocks` endpoint can be implemented by using a combination of `get_blocks` and `get_network_status`, but it is simple enough that I figured it's nice to just have it.

### In this PR
Two new endpoints: `get_blocks` and `get_recent_blocks` in the V2 API.

### Test Plan
Unit tests were added.


briancorbin (approved):
LGTM! Thanks for the changes :)

PR #669, opened 2023-01-20
A basic search endpoint

eranrund:
### Motivation

We'd like to have the option to search the ledger by tx out public keys, key images, and later on TxoIDs (and maybe other things).

### In this PR
* Initial implementation of a basic `search_ledger` endpoint.

### Test Plan
Unit test was added to cover the basic functionality.

### Future Work
* Expand search to cover more queries, such as TxoIDs.

### DIFF
--- a/full-service/src/json_rpc/v2/api/request.rs
+++ b/full-service/src/json_rpc/v2/api/request.rs
@@ -250,4 +250,7 @@ pub enum JsonCommandRequest {
         address: String,
     },
     version,
+    search_ledger {
+        query: String,
+    },
 }
--- a/full-service/src/json_rpc/v2/api/request.rs
+++ b/full-service/src/json_rpc/v2/api/request.rs
@@ -227,6 +227,9 @@ pub enum JsonCommandRequest {
         num_mixins: u64,
         excluded_outputs: Vec<JsonTxOut>,
     },
+    search_ledger {
+        query: String,
+    },
     submit_transaction {
         tx_proposal: TxProposal,
         comment: Option<String>,
@@ -250,7 +253,4 @@ pub enum JsonCommandRequest {
         address: String,
     },
     version,
-    search_ledger {
-        query: String,
-    },
 }
--- a/full-service/src/json_rpc/v2/api/response.rs
+++ b/full-service/src/json_rpc/v2/api/response.rs
@@ -15,6 +15,7 @@ use crate::{
             balance::BalanceMap,
             block::{Block, BlockContents},
             confirmation_number::Confirmation,
+            ledger::LedgerSearchResult,
             network_status::NetworkStatus,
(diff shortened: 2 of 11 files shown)


full-service/src/json_rpc/v2/api/request.rs
```diff
@@ -250,4 +250,7 @@ pub enum JsonCommandRequest {
         address: String,
     },
     version,
+    search_ledger {
+        query: String,
+    },
```
briancorbin, line 255: Nit - The requests should be in alphabetical order

full-service/src/json_rpc/v2/api/response.rs
```diff
@@ -189,6 +190,9 @@ pub enum JsonCommandResponse {
         number: (String, String, String, String),
         commit: String,
     },
+    search_ledger {
+        results: Vec<LedgerSearchResult>,
+    },
```
briancorbin, line 195: Nit - The responses should be in alphabetical order

full-service/src/json_rpc/v2/api/wallet.rs
```diff
@@ -1127,6 +1127,12 @@ where
             ),
             commit: env!("VERGEN_GIT_SHA").to_string(),
         },
+        JsonCommandRequest::search_ledger { query } => {
+            let results = service.search_ledger(&query).map_err(format_error)?;
+            JsonCommandResponse::search_ledger {
+                results: results.iter().map(Into::into).collect(),
+            }
+        }
```
briancorbin, line 1135: Nit - The implementations should be in alphabetical order

briancorbin (changes requested):
So, this is already partially covered by the api request

```json
get_block {
        block_index: Option<String>,
        txo_public_key: Option<String>,
    },
```

which returns

```
get_block {
      block: Block,
      block_contents: BlockContents,
  }
```

I do like the idea of using a query parameter, but we could also add KeyImage to this request as an optional.

Thoughts?

eranrund:
> So, this is already partially covered by the api request
> 
> ```json
> get_block {
>         block_index: Option<String>,
>         txo_public_key: Option<String>,
>     },
> ```
> 
> which returns
> 
> ```
> get_block {
>       block: Block,
>       block_contents: BlockContents,
>   }
> ```
> 
> I do like the idea of using a query parameter, but we could also add KeyImage to this request as an optional.
> 
> Thoughts?

I actually think this is a pretty non-standard way to implement the API - there should probably be separate endpoints for getting by block id / txo public key instead of piling more options to this endpoint.

briancorbin (approved):
Looks good for now, we can chat about how this should be implemented better for v3 API whenever that comes around. Thanks!

remoun (approved):
LGTM

PR #684, opened 2023-01-30
Add optional CORS headers to http server.

Shramp:
### Motivation

The Block Explorer (web app) needs to use full-service, but full-service does not provide CORS headers, so it can't be used by browsers. 

### In this PR
Add an optional argument to start : `allowed-origin`. If provided, full-service will add CORS headers for the specified origin. providing `--allowed-origin "*"` would allow all origins.

### Test Plan
manually tested to make sure i could access FS from the browser when providing the new arg.


eranrund (commented):
Nice. And yeah I think it should be added to the validator mode as well.

christian-oudard (approved):

briancorbin (approved):

PR #686, opened 2023-01-31
updating logger to cancel resetting when falling out of scope

briancorbin:
This fixes a shutdown error that was caused by slog global logger being reset after dropping out of scope while still being used by rocket in an sync context

PR #687, opened 2023-01-31
Brian/updating rocket setup

briancorbin:
Refactoring the mirror service public main function to use rocket::main instead of rocket::launch, which allows the grpcio::Server to remain in scope until the exiting of the program

### DIFF
--- a/mirror/src/public/main.rs
+++ b/mirror/src/public/main.rs
@@ -239,7 +239,7 @@ fn rocket() -> Rocket<Build> {
 
     let query_manager = QueryManager::default();
 
-    start_grpc_server(config.clone(), query_manager.clone(), logger.clone());
+    start_grpc_server(&config, &query_manager, &logger);
 
     build_rocket(config, query_manager, logger)
 }
@@ -247,44 +247,43 @@ fn rocket() -> Rocket<Build> {
 /// Starts the GRPC server in its own thread. This is necessary because the GRPC
 /// server does not implement Sync, which is a requirement to be managed by
 /// Rocket OR pass over an async await block.
-fn start_grpc_server(config: Config, query_manager: QueryManager, logger: Logger) {
-    rocket::tokio::spawn(async move {
-        log::info!(
+fn start_grpc_server(config: &Config, query_manager: &QueryManager, logger: &Logger) {
+    log::info!(
             logger.clone(),
             "Starting wallet service mirror public forwarder, listening for mirror requests on {} and client requests on {}",
             config.mirror_listen_uri.addr(),
             config.client_listen_uri.addr(),
         );
 
-        // Start the mirror-facing GRPC server.
-        log::info!(logger.clone(), "Starting mirror GRPC server");
+    // Start the mirror-facing GRPC server.
+    log::info!(logger.clone(), "Starting mirror GRPC server");
 
-        let build_info_service = BuildInfoService::new(logger.clone()).into_service();
-        let health_service = HealthService::new(None, logger.clone()).into_service();
-        let mirror_service =
-            MirrorService::new(query_manager.clone(), logger.clone()).into_service();
+    let build_info_service = BuildInfoService::new(logger.clone()).into_service();
+    let health_service = HealthService::new(None, logger.clone()).into_service();
+    let mirror_service = MirrorService::new(query_manager.clone(), logger.clone()).into_service();
 
-        let env = Arc::new(
-            EnvBuilder::new()
-                .name_prefix("Mirror-RPC".to_string())
-                .build(),
-        );
+    let env = Arc::new(
+        EnvBuilder::new()
+            .name_prefix("Mirror-RPC".to_string())
+            .build(),
+    );
 
-        let ch_builder = ChannelBuilder::new(env.clone())
-            .max_receive_message_len(-1)
-            .max_send_message_len(-1);
+    let ch_builder = ChannelBuilder::new(env.clone())
+        .max_receive_message_len(-1)
+        .max_send_message_len(-1);
 
-        let server_builder = ServerBuilder::new(env)
-            .register_service(build_info_service)
-            .register_service(health_service)
-            .register_service(mirror_service)
-            .channel_args(ch_builder.build_args())
-            .bind_using_uri(&config.mirror_listen_uri, logger.clone());
+    let server_builder = ServerBuilder::new(env)
+        .register_service(build_info_service)
+        .register_service(health_service)
+        .register_service(mirror_service)
+        .channel_args(ch_builder.build_args())
+        .bind_using_uri(&config.mirror_listen_uri, logger.clone());
 
-        let mut server = server_builder
-            .build()
-            .expect("Failed to build mirror GRPC server");
+    let mut server = server_builder
+        .build()
+        .expect("Failed to build mirror GRPC server");
 
+    rocket::tokio::spawn(async move {
         server.start();
 
         loop {
--- a/mirror/src/public/main.rs
+++ b/mirror/src/public/main.rs
@@ -243,6 +243,11 @@ fn rocket() -> Rocket<Build> {
     build_rocket(config, query_manager, logger)
 }
 
+/// Starts the GRPC server in it's own thread. This is required because rocket
+/// runs in an async context, which either requires the main function to return
+/// the Rocket<Build> which drops the grpc server out of scope, or requires the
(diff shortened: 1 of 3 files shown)


briancorbin:
Current dependencies on/for this PR:
* release/v2.2.1
  * **PR #686** <a href="https://app.graphite.dev/github/pr/mobilecoinofficial/full-service/686" target="_blank">Graphite</a> 
    * **PR #687** <a href="https://app.graphite.dev/github/pr/mobilecoinofficial/full-service/687" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/mobilecoinofficial/full-service/687?utm_source=stack-comment).

eranrund:
Hm, this seems odd to me. Other parts of the mobilecoin codebase are able to do it with a standard async main, for example see https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/overseer/server/src/bin/main.rs

briancorbin:
And correct me if I'm mistaken, but that is a client, not a server, in the example you provided?

briancorbin:
Or are you suggesting it can be wrapped in Arc to allow it to function (I haven't tried that yet)

briancorbin:
This is the error I'm receiving (currently wrapped in Arc, but the same error when it isn't)

```
error: future cannot be sent between threads safely
   --> mirror/src/public/main.rs:235:46
    |
235 |   async fn main() -> Result<(), rocket::Error> {
    |  ______________________________________________^
236 | |     mc_common::setup_panic_handler();
237 | |     let _sentry_guard = mc_common::sentry::init();
238 | |
...   |
249 | |     Ok(())
250 | | }
    | |_^ future created by async block is not `Send`
    |
    = help: the trait `Sync` is not implemented for `(dyn ServerChecker + 'static)`
note: future is not `Send` as this value is used across an await
   --> mirror/src/public/main.rs:247:65
    |
246 |     let grpc_server = start_grpc_server(config.clone(), query_manager.clone(), logger.clone());
    |         ----------- has type `Arc<grpcio::Server>` which is not `Send`
247 |     let _ = build_rocket(config, query_manager, logger).launch().await?;
    |                                                                 ^^^^^^ await occurs here, with `grpc_server` maybe used later
...
250 | }
    | - `grpc_server` is later dropped here
note: required by a bound in `async_main`
   --> /tmp/mobilenode/.mob/cargo/registry/src/github.com-1ecc6299db9ec823/rocket-0.5.0-rc.2/src/lib.rs:262:66
    |
262 | pub fn async_main<R>(fut: impl std::future::Future<Output = R> + Send) -> R {
    |                                                                  ^^^^ required by this bound in `async_main`

error: future cannot be sent between threads safely
   --> mirror/src/public/main.rs:235:46
    |
235 |   async fn main() -> Result<(), rocket::Error> {
    |  ______________________________________________^
236 | |     mc_common::setup_panic_handler();
237 | |     let _sentry_guard = mc_common::sentry::init();
238 | |
...   |
249 | |     Ok(())
250 | | }
    | |_^ future created by async block is not `Send`
    |
    = help: the trait `Sync` is not implemented for `(dyn grpcio::server::CloneableHandler + 'static)`
note: future is not `Send` as this value is used across an await
   --> mirror/src/public/main.rs:247:65
    |
246 |     let grpc_server = start_grpc_server(config.clone(), query_manager.clone(), logger.clone());
    |         ----------- has type `Arc<grpcio::Server>` which is not `Send`
247 |     let _ = build_rocket(config, query_manager, logger).launch().await?;
    |                                                                 ^^^^^^ await occurs here, with `grpc_server` maybe used later
...
250 | }
    | - `grpc_server` is later dropped here
note: required by a bound in `async_main`
   --> /tmp/mobilenode/.mob/cargo/registry/src/github.com-1ecc6299db9ec823/rocket-0.5.0-rc.2/src/lib.rs:262:66
    |
262 | pub fn async_main<R>(fut: impl std::future::Future<Output = R> + Send) -> R {
    |                                                                  ^^^^ required by this bound in `async_main`
```

briancorbin:
Ooooooh I got it, I was missing wrapping the server in a Mutex >_>

briancorbin:
We'll make a rust programmer out of me yet ;D

mirror/src/public/main.rs
```diff
@@ -282,8 +294,14 @@ fn rocket() -> _ {
     let mut server = server_builder
         .build()
         .expect("Failed to build mirror GRPC server");
+
     server.start();
```
ryankurte, line 298: you could equally use `tokio::task::spawn(async move { server.start() })` then return the `JoinHandle` to hold onto, but it's pretty moot as the `server` seems to already pretty much be a handle.

briancorbin: Yeah it seems like there's a number of ways we _could_ do this, and I don't think any of them have any real implication one way or another.. @eranrund do you have any preference whether we use a thread or Arc Mutex?

briancorbin: Honestly, I prefer using the `tokio::task::spawn` approach myself, it feels cleaner than using Arc Mutex

eranrund:
I am confused... why does the server need to move anywhere?
This feels like it is going in a messier direction that needed, unless I am missing something.
You should be able to start the GRPC server, then start Rocket similar to what is done here:
https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/overseer/server/src/bin/main.rs#L48-L49
I don't see why that would result in the server object being sent anywhere, or for main() returning while rocket is still running.

briancorbin:
grpcio::Server does not implement Sync, which is a requirement for it to either

A) be managed by rocket as a state
B) remain in scope over an async await

So the 2 ways around that that I have figured out is either

A) wrapping the server in an Arc<Mutex<>>
B) starting and maintaining the server in its own thread

briancorbin:
I'm happy to show you in a screen share if you'd like, but it will not let me compile it like that without wrapping it in an Arc Mutex or starting its own thread

eranrund:
> grpcio::Server does not implement Sync, which is a requirement for it to either
> 
> A) be managed by rocket as a state B) remain in scope over an async await
> 
> So the 2 ways around that that I have figured out is either
> 
> A) wrapping the server in an Arc<Mutex<>> B) starting and maintaining the server in its own thread

And why can't it remain in the scope? Isn't that what is happening in https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/overseer/server/src/bin/main.rs#L48-L49 where OverseerService remains in the scope and rocket is launched and awaited in the end?

eranrund:
> I'm happy to show you in a screen share if you'd like, but it will not let me compile it like that without wrapping it in an Arc Mutex or starting its own thread

Sure, we can do that, feel free to put something on my calendar tomorrow or Fri.

briancorbin:
> > grpcio::Server does not implement Sync, which is a requirement for it to either
> > A) be managed by rocket as a state B) remain in scope over an async await
> > So the 2 ways around that that I have figured out is either
> > A) wrapping the server in an Arc<Mutex<>> B) starting and maintaining the server in its own thread
> 
> And why can't it remain in the scope? Isn't that what is happening in https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/overseer/server/src/bin/main.rs#L48-L49 where OverseerService remains in the scope and rocket is launched and awaited in the end?

Yeah but overseer service isn't a grpcio::Server instance as far as I can tell, is it?

briancorbin:
> > > grpcio::Server does not implement Sync, which is a requirement for it to either
> > > A) be managed by rocket as a state B) remain in scope over an async await
> > > So the 2 ways around that that I have figured out is either
> > > A) wrapping the server in an Arc<Mutex<>> B) starting and maintaining the server in its own thread
> > 
> > 
> > And why can't it remain in the scope? Isn't that what is happening in https://github.com/mobilecoinfoundation/mobilecoin/blob/master/fog/overseer/server/src/bin/main.rs#L48-L49 where OverseerService remains in the scope and rocket is launched and awaited in the end?
> 
> Yeah but overseer service isn't a grpcio::Server instance as far as I can tell, is it?

Besides, OverseerService is ultimately spinning of its own thread under the hood anyways for the grpc client(s)

https://github.com/mobilecoinfoundation/mobilecoin/blob/4df4bd6b07e47ec3d8ba8e36406a569528e5bcee/fog/overseer/server/src/worker.rs#L44-L77

eranrund (approved):

PR #691, opened 2023-02-01
adding info logging for command success and failure message

briancorbin:
### Motivation

This is needed to better identify errors that are reported to the API that aren't currently reported in the logs. This only reports the outcome and any error that comes back with it.

### In this PR
* Added logging to API response


briancorbin:
> Approved, but question: Would mirror/validator node also benefit from increased logging?

Yes, but out of scope for this PR

PR #698, opened 2023-02-06
Feat/reintroduce secreted status

pouneh:
## Motivation

Currently the API responses include a "secreted" when listing balances. However, this field was unhelpfully always set to zero. Between removing the API, and reintroducing the status, we reintroduce the status.

## In this PR

We introduce 2 new txo status.

    "Created" - for output TXOs that are proposed as part of a built transaction (either not submitted, or failed)
    "Secreted" - for output TXOS that are build as part of successfully submitted transactions that go to a wallet that does NOT belong to the sender

### DIFF
--- a/tools/test.sh
+++ b/tools/test.sh
@@ -4,12 +4,18 @@
 
 # RUSTFLAGS="-C instrument-coverage" \
 
-if ! [[ -f "$(pwd)/consensus-enclave.css" ]]; then
-  echo "Please copy the consensus-enclave.css to $(pwd)"
-  exit 1
-fi
+# Grab current location and source the shared functions.
+# shellcheck source=.shared-functions.sh
+location=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+source "${location}/.shared-functions.sh"
 
-echo "Testing in $PWD"
-SGX_MODE=SW IAS_MODE=DEV CONSENSUS_ENCLAVE_CSS=$(pwd)/consensus-enclave.css \
+net="test"
+WORK_DIR="${WORK_DIR:-"${HOME}/.mobilecoin/${net}"}"
+mkdir -p "${WORK_DIR}"
+
+SGX_MODE=SW
+IAS_MODE=DEV
+CONSENSUS_ENCLAVE_CSS=$(get_css_file "${net}" "${WORK_DIR}/consensus-enclave.css")
+export SGX_MODE IAS_MODE CONSENSUS_ENCLAVE_CSS
+
 cargo test $1
-echo "Testing in $PWD complete."
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -1376,7 +1376,7 @@ impl TxoModel for Txo {
         )?;
 
         if spendable_txos.is_empty() {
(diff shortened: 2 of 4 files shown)


full-service/src/json_rpc/v1/models/txo.rs
```diff
@@ -66,6 +66,7 @@ impl From<&db::txo::TxoStatus> for TxoStatus {
             db::txo::TxoStatus::Spent => TxoStatus::Spent,
             db::txo::TxoStatus::Unspent => TxoStatus::Unspent,
             db::txo::TxoStatus::Unverified => TxoStatus::Unspent,
+            db::txo::TxoStatus::Created | db::txo::TxoStatus::Secreted => todo!(),
```
briancorbin, line 69: How come this is left as todo?

tools/build-fs.sh
```diff
@@ -90,7 +90,7 @@ case ${net} in
         if test -f "${WORK_DIR}/consensus-enclave.css"; then
             CONSENSUS_ENCLAVE_CSS="${WORK_DIR}/consensus-enclave.css"
         else
-            CONSENSUS_ENCLAVE_CSS=`pwd`/target/docker/release/consensus-enclave.css
+            CONSENSUS_ENCLAVE_CSS=`pwd`/mobilecoin/target/docker/release/consensus-enclave.css
```
briancorbin, line 93: Can we move these changes to a separate PR? Out of scope for this review

tools/run-fs.sh
```diff
@@ -130,7 +130,7 @@ case "${net}" in
         if [[ -z "${offline}" ]]
         then
             MC_PEER="insecure-mc://localhost:3200/,insecure-mc://localhost:3201/"
-            MC_TX_SOURCE_URL="http://localhost:4566/node-0-ledger/,http://localhost:4566/node-1-ledger/"
+            MC_TX_SOURCE_URL="file:///home/$USER/full-service/mobilecoin/target/docker/release/mc-local-network/node-ledger-distribution-0,file:///home/$USER/full-service/mobilecoin/target/docker/release/mc-local-network/node-ledger-distribution-1"
```
briancorbin, line 133: Can we move these changes to a separate PR? Out of scope for this review

tools/test.sh
```diff
@@ -4,6 +4,12 @@
 
 # RUSTFLAGS="-C instrument-coverage" \
 
+
+if ! [[ -f "$(pwd)/consensus-enclave.css" ]]; then
+  echo "Please copy the consensus enclave to $(pwd)"
+  exit 1
+fi
+
```
briancorbin, line 12: Can we move these changes to a separate PR? Out of scope for this review

briancorbin (changes requested):
This looks good so far, but it could use a bunch more tests, both directly in the db/txo.rs file and in the e2e tests. I would recommend creating tests for a variety of possible situations that the database could be in to ensure we get the correct status at the correct times.

It would also be a good idea to test that querying an individual txo results in the correct status, again both in the e2e tests as well as the db/txo.rs tests

christian-oudard (approved):

briancorbin (approved):
This looks good, thanks for all of the changes!

PR #700, opened 2023-02-07
Fix error message for no txos of token type.

christian-oudard:
-


briancorbin (approved):

briancorbin (commented):
This needs a lint fix @christian-oudard

PR #701, opened 2023-02-08
Add watcher sync thread to rocket.manage

briancorbin:
This prevents the sync thread from dropping out of scope when the function it's created in returns, similar to how we are handling the ledger sync thread.


joekottke (approved):
Ran into this last night, and while I can't be trusted to review rust code, this looks good to me.   Someone else should review it for actual accuracy.  I'm just the plant operator in this ;)

eranrund (approved):

PR #702, opened 2023-02-08
Script improvements

pouneh:
### Motivation

Test.sh
- Previously we just ran it, and it may fail part way through if a consensus sigstruct doesn't exist in the full-service root directory. This is an issue mostly on a fresh clone of the repository. This change just makes the script fail earlier if it would fail on a missing sigstruct file anyway.

Build.sh
- The sigstruct gets build as part of building the mobilecoin submodule, and should therefore be grabbed from its `target` directory.

Run.sh
- Unfortunately, the `localhost/port#` way of getting the ledger-distribution data isn't working the way it used to when running full-service against a local-network in a different container. We just grab the file directly.


tools/run-fs.sh
```diff
@@ -130,7 +130,7 @@ case "${net}" in
         if [[ -z "${offline}" ]]
         then
             MC_PEER="insecure-mc://localhost:3200/,insecure-mc://localhost:3201/"
-            MC_TX_SOURCE_URL="http://localhost:4566/node-0-ledger/,http://localhost:4566/node-1-ledger/"
+            MC_TX_SOURCE_URL="file:///home/$USER/full-service/mobilecoin/target/docker/release/mc-local-network/node-ledger-distribution-0,file:///home/$USER/full-service/mobilecoin/target/docker/release/mc-local-network/node-ledger-distribution-1"
```
eranrund, line 133: This path is very specific to one's setup - is it possible to figure it out from the current directory this script is running in?

tools/test.sh
```diff
@@ -4,6 +4,11 @@
 
 # RUSTFLAGS="-C instrument-coverage" \
 
+if ! [[ -f "$(pwd)/consensus-enclave.css" ]]; then
+  echo "Please copy the consensus enclave to $(pwd)"
```
eranrund, line 8:
```suggestion
  echo "Please copy the consensus-enclave.css to $(pwd)"
```

tools/run-fs.sh
eranrund: Who is the main user of this script?

eranrund: What prevents something like `MC_TX_SOURCE_URL="file:///$PWD/mobilecoin/target/docker/r...` from working?

eranrund (approved):

briancorbin (approved):

PR #714, opened 2023-02-15
Update txos fully when syncing

pouneh:
### Motivation

When a txo is synced from the database but already exists, it should update all of the information, not just the new expected info, in the case that the information from the TxProposal was malformed or incorrect in any way.

### In this PR
We additionally update
1. Amount (value, token_id)
2. target_key
3. public_key
4. e_fog_hint


full-service/src/db/txo.rs
```diff
@@ -376,10 +380,14 @@ impl TxoModel for Txo {
...
             Ok(txo) => {
                 txo.update_as_received(
-                    account_id_hex,
                     subaddress_index,
                     key_image,
                     received_block_index,
+                    account_id_hex,
+                    amount,
+                    &txo.target_key,
+                    &txo.public_key,
```
briancorbin, line 389: So, these 2 should definitely be encoded as things currently work, and it's a bit concerning that they're passing tests without...

full-service/src/db/txo.rs
```diff
@@ -462,10 +470,14 @@ impl TxoModel for Txo {
...
-        received_account_id_hex: &str,
-        received_subaddress_index: Option<u64>,
+        subaddress_index: Option<u64>,
         received_key_image: Option<KeyImage>,
         block_index: u64,
+        account_id_hex: &str,
+        amount: Amount,
+        target_key: &Vec<u8>,
+        public_key: &Vec<u8>,
+        e_fog_hint: &Vec<u8>,
```
briancorbin, line 480: Perhaps we leave these not updated for now until we understand better why they are passing tests without being protobuf encoded.

briancorbin (changes requested):
This is good, but I would like to understand why we're not encoding it the second time around with mc util serial like we are the first time

full-service/src/db/txo.rs
briancorbin: Oh I see what's going on...

full-service/src/db/txo.rs
```diff
@@ -376,10 +380,14 @@ impl TxoModel for Txo {
             // transaction), we need to update it
             Ok(txo) => {
```
briancorbin, line 381: This line is overwriting the `TxOut` that is passed in under the variable name `txo`, which is why the encoding is weird. I would recommend changing the name of either one of them as to preserve both of their information sets

briancorbin (changes requested):

christian-oudard (approved):

briancorbin (approved):
Thanks for the changes, this looks good now :)

PR #717, opened 2023-02-16
updating branches from develop->main

briancorbin:
Since the [gitbook docs](https://mobilecoin.gitbook.io/full-service-api/) are attached to the `main` branch but we had been using `develop` as our primary branch (and not really using `main`), the easiest thing to do was just to rename `develop` to `main`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,13 +11,11 @@ env:
 on:
   pull_request:
     branches:
-      - develop
       - main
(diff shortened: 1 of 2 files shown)


brucek (changes requested):
Can you just quickly explain why we're doing this in the description please?

brucek (approved):
It turns out I can edit your comment(!). So I just added a bit more details

PR #718, opened 2023-02-17
adding the collection of v1 endpoints to mirror allowed endpoints

briancorbin:
This is to enable all v1 endpoints that were missing compared to the full-service-mirror repo


brucek (approved):
Future work:  add v1 tests [#184497561](https://www.pivotaltracker.com/story/show/184497561)

PR #720, opened 2023-02-17
removing mac builds - only for v2.3 series

briancorbin:
This is to allow CI to run smoothly for the 2.3 series, which is not working on mac at the moment. This has been resolved in 2.4, so this PR should not be merged back in.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,8 +11,8 @@ env:
 on:
   push:
     tags:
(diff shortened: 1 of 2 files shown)


brucek (approved):
A little sloppy to have quote changes here too but whatevs

briancorbin:
> A little sloppy to have quote changes here too but whatevs

Oh that must have been the auto formatting

PR #730, opened 2023-02-23
Introduce API to reimport/rescan accounts

pouneh:
### Motivation

Sometimes customers may wish to resync their full-service accounts to deal with transient issues  

### In this PR
We introduce a new api in the v2 jsonrpc that allows us to reimport an account by rescanning blocks from the account's `first_block_index`

### Test Plan
Set SGX_MODE, IAS_MODE, and CONSENSUS_ENCLAVE_CSS. Then run: 

`cargo test --package mc-full-service --lib service::account::tests::test_reimport_account -- --exact`


full-service/src/json_rpc/v2/api/request.rs
```diff
@@ -230,6 +230,9 @@ pub enum JsonCommandRequest {
         first_block_index: Option<String>,
         next_subaddress_index: Option<String>,
     },
+    reimport_account {
```
briancorbin, line 233: I think we should call this "resync_account" or something, as it doesn't have anything to do with importing an account

full-service/src/json_rpc/v2/api/response.rs
```diff
@@ -181,6 +181,7 @@ pub enum JsonCommandResponse {
     import_view_only_account {
         account: Account,
     },
+    reimport_account,
```
briancorbin, line 184: Rename to "resync_account"

briancorbin (changes requested):
This looks good, just a comment about naming :) Also, would like to see a test that exercises a rescan fixing an incorrectly stored txo, and making sure this rescan doesn't futz with existing transaction logs, as tx logs are updated from the sync process.

briancorbin:
Could we also have this change made against release/2.5 branch

briancorbin (changes requested):
This still has the "reimport_account" name instead of "resync_account"

full-service/src/json_rpc/v2/api/request.rs
```diff
@@ -230,6 +230,9 @@ pub enum JsonCommandRequest {
         first_block_index: Option<String>,
         next_subaddress_index: Option<String>,
     },
+    resync_account {
```
briancorbin, line 233: One more thing is to order this alphabetically

full-service/src/json_rpc/v2/api/response.rs
```diff
@@ -181,6 +181,7 @@ pub enum JsonCommandResponse {
     import_view_only_account {
         account: Account,
     },
+    resync_account,
```
briancorbin, line 184: Order alphabetically

full-service/src/json_rpc/v2/api/wallet.rs
```diff
@@ -1113,6 +1113,13 @@ where
 
             JsonCommandResponse::import_view_only_account { account }
         }
+        JsonCommandRequest::resync_account { account_id } => {
```
briancorbin, line 1116: Order alphabetically

full-service/src/service/sync.rs
```diff
@@ -122,8 +122,9 @@ pub fn sync_all_accounts(
         { Account::list_all(conn, None, None).expect("Failed getting accounts from database") };
 
     for account in accounts {
-        // If there are no new blocks for this account, don't do anything.
-        if account.next_block_index as u64 > num_blocks - 1 {
+        // If there are no new blocks for this account, or we are in offline mode (so
```
briancorbin, line 125: Being in offline mode doesn't mean there are no blocks, as a user can manually load a recent ledger_db from an online machine, which would trigger this. This shouldn't affect the logic here, but I feel justifies changing the comment.

briancorbin: Also, this check for if num_blocks == 0 might be better before this loop through the accounts, since it would apply to the whole wallet (basically this is a redundant check)

briancorbin (changes requested):

full-service/src/service/sync.rs
```diff
@@ -121,13 +121,14 @@ pub fn sync_all_accounts(
...
-        if account.next_block_index as u64 > num_blocks - 1 {
-            continue;
+    if num_blocks != 0 {
+        for account in accounts {
+            // If there are no new blocks for this account, don't do anything.
+            if account.next_block_index as u64 > num_blocks - 1 {
+                continue;
+            }
+            sync_account_next_chunk(ledger_db, conn, &account.id, logger)?;
         }
```
briancorbin, line 131:
```suggestion
    if num_blocks == 0 {
        Ok(())
    }
    for account in accounts {
        // If there are no new blocks for this account, don't do anything.
        if account.next_block_index as u64 > num_blocks - 1 {
            continue;
        }
        sync_account_next_chunk(ledger_db, conn, &account.id, logger)?;
    }
```

briancorbin: Generally I think we want to try and avoid unnecessary nesting of code, so instead of wrapping the loop inside an if statement, we could instead invert the conditional and just return, otherwise continue with the execution of the program. This becomes exceedingly more and more important as conditions grow and become more complex.

briancorbin (changes requested):
This looks good other than 1 more change request! Thanks for all the hard work on this :)

full-service/src/service/sync.rs
briancorbin: Also, not sure if that suggestion is commit worthy, so I would advise just implementing it yourself

briancorbin (approved):

PR #731, opened 2023-02-24
Implement payment requests for all token types.

christian-oudard:
- Add token-id to payment request results from check_b58_type.
- Implement payment requests in Python API and CLI.


briancorbin (changes requested):
This looks great, but could we add a test to the e2e tests as well to exercise the endpoint? This would only need to be in v2

christian-oudard:
> This looks great, but could we add a test to the e2e tests as well to exercise the endpoint? This would only need to be in v2

End-to-end test is written.

briancorbin (approved):

PR #743, opened 2023-03-03
v2.3 - fix: set chain-id to main for main|prod network

jgreat:
### Motivation

Small change in tools/run-fs.sh helper script to fix the chain-id logic when running on `main` or the legacy `prod` networks.

### Test Plan

```
# build fs for main network
tools/build-fs.sh main

# start fs with run-fs.sh helper script with main or prod networks
tools/run-fs.sh prod

OR

tools/run-fs.sh main

# watch for blocks to start downloading
```

PR #744, opened 2023-03-03
Cleaning up some main/prod build & run issues

brucek:
### Motivation

the build & run scripts were broken for main

### In this PR
* Fixes [#184616514](https://www.pivotaltracker.com/story/show/184616514)

### Test Plan

run:
* `tools/build-fs.sh main`
* `tools/build-fs.sh prod`
* `tools/run-fs.sh main`
* `tools/run-fs.sh prod`
* `tools/run-fs.sh --build main`
* `tools/run-fs.sh --build prod`

### DIFF
--- a/tools/build-fs.sh
+++ b/tools/build-fs.sh
@@ -56,23 +56,21 @@ WORK_DIR="${WORK_DIR:-"${HOME}/.mobilecoin/${net}"}"
 mkdir -p "${WORK_DIR}"
 
 case ${net} in
-    test|prod|main)
-        echo "Setting '${net}' SGX, IAS and enclave values"
-
+    prod|main)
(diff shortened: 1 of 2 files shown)


briancorbin (approved):

jgreat (changes requested):
I think this is going to cause conflicts with: https://github.com/mobilecoinofficial/full-service/pull/743 that's pointed at release/v2.3

We should probably do one or the other.  starting at v2.3 and promoting the changes _should_ be the cleaner way I think.

brucek:
> I think this is going to cause conflicts with: #743 that's pointed at release/v2.3
> 
> We should probably do one or the other. starting at v2.3 and promoting the changes _should_ be the cleaner way I think.

Yep, yours looks good, I'll close this

PR #745, opened 2023-03-04
Add v1 api mirror tests

jgreat:
### Motivation

Primary: Add v1 API mirror tests.

Secondary: 
- Improve helper scripts to launch v1 and v2 copies of mirror. Listening of ports 9091 and 9092 respectively.
- Separate target directories for main vs test cargo builds. 
- Detect `mob prompt` container and allow full-service and public mirror endpoint "publish" and use storage for ledger/wallet db's that will be persistent (.mob directory)

### Test Plan

```
# test builds
tools/build-fs.sh main

# start mirror services
cd mirror/test
./start.sh --network main

# run v1 and v2 mirror tests
./test.sh --mnemonic "<24 words>"
```

### DIFF
--- a/mirror/test/start.sh
+++ b/mirror/test/start.sh
@@ -20,7 +20,7 @@ do
             exit 0
             ;;
         --network)
-            network="${2}"
+            net="${2}"
             shift 2
             ;;
         *)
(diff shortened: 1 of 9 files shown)


mirror/test/start.sh
```diff
@@ -20,7 +20,7 @@ do
             exit 0
             ;;
         --network)
-            network="${2}"
+            net="${2}"
```
jgreat, line 23: use "$net" for compatibility with .shared-functions

briancorbin (approved):

PR #754, opened 2023-03-09
Augment get_network_status with more information

pouneh:
### Motivation

We want to augment `get_network_status` so that users can provide more information about their setup when debugging or reporting issues.

### In this PR
We augment `get_network_status` to give more information about the launching configuration of the full-service instance.

### DIFF
--- a/full-service/src/bin/main.rs
+++ b/full-service/src/bin/main.rs
@@ -204,7 +204,7 @@ fn consensus_backed_full_service(
         ledger_db,
         watcher_db,
         peer_manager,
-        Some(APIConfig.peers_config.clone()),
+        Some(config.peers_config.clone()),
         network_state,
         config.get_fog_resolver_factory(logger.clone()),
         config.offline,
@@ -280,7 +280,7 @@ fn validator_backed_full_service(
         ledger_db,
         None,
         conn_manager,
-        Some(APIConfig.peers_config.clone()),
+        Some(config.peers_config.clone()),
         network_state,
         Arc::new(move |fog_uris| -> Result<FogResolver, String> {
             if fog_uris.is_empty() {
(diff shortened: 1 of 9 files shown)


full-service/src/bin/main.rs
```diff
@@ -274,11 +295,32 @@ fn validator_backed_full_service(
...
+        ),
+    };
+
+    let network_setup_config = NetworkSetupConfig {
+        offline: config.offline.clone(),
+        chain_id,
+        peers,
+        tx_sources,
+    };
+
```
briancorbin, line 317: Could we move this to the higher level function (main) so it isn't duplicated and just pass it in as part of the params of `consensus_backed_` and `validator_backed_full_service`?

full-service/src/bin/main.rs
```diff
@@ -199,11 +200,31 @@ fn consensus_backed_full_service(
...
```
briancorbin, line 221: Could we move this to the higher level function (`main`) so it isn't duplicated and just pass it in as part of the params of `consensus_backed_` and `validator_backed_full_service`?

full-service/src/config.rs
```diff
@@ -187,6 +188,16 @@ pub struct PeersConfig {
     pub chain_id: String,
 }
 
+/// The Network Setup object.
+/// This holds a copy of the network parameters used to start full-service
+#[derive(Default, Clone, Debug, Deserialize, Serialize)]
+pub struct NetworkSetupConfig {
```
briancorbin, line 194: We can probably just call this `NetworkConfig`

briancorbin (changes requested):
Thanks for this enhancement! I left a few comments below. Also, could you rebase and target onto a release branch? release/v2.5 should be fine, or if it slips to release/v2.6 that's okay, too (once it's created)

briancorbin (approved):

PR #773, opened 2023-03-15
Expose max tombstone block as part of network status

pouneh:
### Motivation

Full-service clients want to be better-informed when decided the tombstone block when building a transaction. This gives them information about what the maximum delta of tombstone block they can use compared to the network height

Resolves PV183840879

### DIFF
--- a/full-service/src/json_rpc/v2/models/network_status.rs
+++ b/full-service/src/json_rpc/v2/models/network_status.rs
@@ -25,6 +25,8 @@ pub struct NetworkStatus {
     /// The current block version
     pub block_version: String,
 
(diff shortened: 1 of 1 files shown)


full-service/src/json_rpc/v2/models/network_status.rs
```diff
@@ -25,6 +25,8 @@ pub struct NetworkStatus {
     /// The current block version
     pub block_version: String,
 
+    pub max_tombstone_blocks: String,
```
briancorbin, line 28: Make sure to add doc comments :)

briancorbin (changes requested):
See inline comments

PR #781, opened 2023-03-21
CLI support for Ledger Nano S-Plus hardware wallet.

christian-oudard:
-


briancorbin:
This isn't going into main yet, and should be stacked on top of the other branches for ledger hw support

PR #793, opened 2023-03-27
Add legacy entropy support for creating view-only accounts in tx signer

Shramp:
### Motivation

We have an investor who needs to prove to their auditor that they hold some amount of MOB. We want to give them a way to provide the investor with the view-only account, but the tx signer does not support legacy entropy accounts.

### In this PR
Add `ImportFromLegacy` option to the tx signer to create view-only acount import packages for legacy entropies.

NOTE: this does not (yet) provide support for syncing txos for legacy entropy accounts


Shramp:
Going to close this and revisit solutions if it comes up gain

brucek:
I think we may want this support after all - I'll circle with @briancorbin to discuss

briancorbin (approved):
This looks good, just need to fix the CI issues

PR #803, opened 2023-04-10
store decrypted memo in txo table

pouneh:
### Motivation

Customers are interested in querrying txos based on the address stored in the RTH memo. This is prerequisite work for enabling that feature


briancorbin (changes requested):
So, this is a good start, but it still needs a few things

1. Tests to exercise the new functionality
2. A way to backfill memo info for all of the existing txos in the database when the migration is first applied

PR #804, opened 2023-04-11
updating rust toolchain for ci test

briancorbin:
Fixing an out of date rust toolchain causing compilation to fail when running tests because it requires a rust toolchain higher than the one set.

This should have been updated in the past anyways, as they should all be running on the same toolchain.


briancorbin:
Oh, no it's for some stuff in a different PR that needs the toolchain to be updated (unit test stuff). The tests themselves aren't failing, it's the compilation that is

This particular fix is from clap I think, one of the cargo crates

PR #805, opened 2023-04-11
Update API's TXO model to return the memo type and address hash

pouneh:
### Motivation

Although we now store this info in the db, FS users can't access it. With this change, they will get this information on all APIs that return a TXO

### Future Work
Add tests

### DIFF
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -72,6 +72,14 @@ pub enum TxoStatus {
     Unverified,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum RTHMemoType {
+    AuthenticatedSender,
+    AuthenticatedSenderWithPaymentIntentId,
+    AuthenticatedSenderWithPaymentRequestId,
+    Unsupported,
+}
+
 impl fmt::Display for TxoStatus {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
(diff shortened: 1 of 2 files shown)


full-service/src/db/txo.rs
```diff
@@ -72,6 +72,14 @@ pub enum TxoStatus {
     Unverified,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum RTHMemoType {
+    AuthenticatedSender,
+    AuthenticatedSenderWithPaymentIntentId,
+    AuthenticatedSenderWithPaymentRequestId,
+    Unsupported,
+}
```
briancorbin, line 81: This should also maybe be just a general Memo enum, not Memo Type, and include the data in the actual enum that goes along with each of them

briancorbin:
```suggestion
pub enum Memo {
    AuthenticatedSender { address_hash: String },
    AuthenticatedSenderWithPaymentIntentId { address_hash: String, payment_intent_id: String },
    AuthenticatedSenderWithPaymentRequestId { address_hash: String, payment_request_id: String },
    ...remaining memos after this
}
```

briancorbin (changes requested):

briancorbin:
This also needs an API endpoint that allows users to exercise this additional data

full-service/src/db/txo.rs
briancorbin: Actually, it looks like it can only be 1 to 1 relationship, so maybe at the database level it's broken into memo type and underlying data, and at the model level it combines them into a single enum that takes in values

PR #816, opened 2023-04-17
Integration testing fs/fsm+validator - TestNet

jgreat:
### Motivation

Integration test standalone full-service and full-service with mirror and validator on testnet.
 
### In this PR
Full-service
* minor fix for full-service option - MC_WATCHER_DB to use clap

Charts/Docker
* make dockerfile "batteries included" - sets MC_PEER, MC_TX_SOURCE_URL for test/main builds.
* update runtime container base image
* Improve fs container entrypoint script, MC_READ_ONLY option, MC_VALIDATOR detection, and restore from MC_LEDGER_DB_URL
* Optional: Import wallets on startup - still needs a bit of work.
* Add full-service-mirror helm chart for deploying setup in k8s clusters
* standardize configuration for full-service and full-service-mirror charts.

CI/CD
* Add cache key differentiation for test/main builds.
* Use mobilecoinofficial/gh-actions for "standard" cache and checkout actions.
* trigger "cd" testing on feature/* branch
*  "cd" testing use the `matrix` feature for testnet and full-service standalone and fs with mirror+validator (add main/alpha testing later).
* restore partial ledger from azure blob storage for faster startup times.
* wire up secrets for 2 test wallets so fs and fsm+v tests don't get confused.

Tools
* Clean up some differences in CARGO_TARGET_DIR between `mob prompt` and local development
* CARGO_TARGET_DIR includes "network" so test/main builds don't stomp on each other.
* Stop duplicating binaries, make WORK_DIR a symlink to  CARGO_TARGET_DIR path, and migrate ledger/wallet dbs if they exist.

### Test Plan

Automated testing under "Development CD"

### Future Work
* Need automatic update process for full-service ledger db being stored in azure blob storage
* Come back and move other ci/cd processes to use new standardized github actions (mco/gh-actions)
* wire up mainnet integration testing
* refactor other fs build processes to deduplicate workflows.

### DIFF
--- a/tools/build-fs.sh
+++ b/tools/build-fs.sh
@@ -49,7 +49,7 @@ fi
 # use main instead of legacy prod
 if [[ "${net}" == "prod" ]]
 then
-    echo "Detected \"prod\" legacy network setting. Using \"main\" instead."
+    echo "Detected 'prod' legacy network setting. Using 'main' instead."
     net=main
 fi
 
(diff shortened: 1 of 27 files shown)


tools/build-fs.sh
```diff
@@ -112,11 +141,4 @@ echo "building full service..."
...
 cargo build --release ${BUILD_OPTIONS}
 
-target_dir=${CARGO_TARGET_DIR:-"target"}
-
-if [[ "${target_dir}/release" != "${WORK_DIR}" ]]
-then
-    echo "  binaries are available in ${target_dir}/release and ${WORK_DIR}"
-    cp "${target_dir}/release/full-service" "${WORK_DIR}"
-    cp "${target_dir}/release/validator-service" "${WORK_DIR}"
-fi
```
briancorbin, line 122: How come we're getting rid of this? It's been very useful to have to build script copy the built files directly into my ~/.mobilecoin/{NET} directory.

briancorbin:
Also, I think it's fine to share build directories for main vs test vs other, as they share most of the same build stuff and having them be separate causes SO much more space to be taken up if you're building for multiple networks. It takes like, 10 seconds to switch networks and rebuild

tools/build-fs.sh
jgreat: This is to take up less space.  We haven't been building to that directory, but copying the binaries around. 

Here are some problems with the current setup

- This is to take up less space.  We haven't been building to that directory, but copying the binaries around. 
- Using $HOME/.mobilecoin  I can't work on multiple repo copies the the same time (which is how I keep multiple workflows straight), builds over write each other. 
- This isn't really compatible with mob prompt since $HOME is local to the container

Consolidating the files into CARGO_TARGET_DIR and making "WORK_DIR" a link allows us to have one standard path place that is relative to the repo we're working in.

briancorbin (approved):

PR #841, opened 2023-04-27
remove key derivation version param

briancorbin:
This is a legacy field from a long time ago when the import and import_from_legacy_root_entropy methods were combined. Since they have been separated, it no longer is necessary to specify the KDV


christian-oudard (approved):
easy cleanup, obvious win.

PR #851, opened 2023-05-03
Store data as raw bytes - txo - key_image

briancorbin:
### Motivation

< The motivation for the changes in this PR. "Currently we...", "This is needed because..." >

### In this PR
* < Additions, removals, fixes, features >

< Ticket status, e.g. "fixes #issue number" > 

### Test Plan

< How to run the unit test, e2e test etc. >
< Or what are the actual request/response might look like (like grpcurl) >

### Future Work
* < Out of scope non-goals for this PR >
* < These should be links to tickets. If the tickets do not exist, make them. >


full-service/src/db/assigned_subaddress.rs
```diff
@@ -3,6 +3,8 @@
 //! A subaddress assigned to a particular contact for the purpose of tracking
 //! funds received from that contact.
 
+use std::convert::TryInto;
+
```
eranrund, line 7: Nit: no new line, all imports grouped together

full-service/src/db/wallet_db_error.rs
```diff
@@ -170,6 +170,9 @@ pub enum WalletDbError {
 
     /// MemoDecoding: {0}
     MemoDecoding(MemoDecodingError),
+
+    /// Ring Signature error: {0}
+    RingSignature(mc_transaction_core::ring_signature::Error),
```
eranrund, line 175: Curious why this was needed?

eranrund (commented):
This appears to be a very breaking change since it alters the format inside the database. Whats the migration plan for it?

PR #852, opened 2023-05-04
Fix the git sha returned by the version endpoint. It was broken by the Renovate bot.

christian-oudard:
-


briancorbin (commented):
This seems to be causing the version unit test to fail now

briancorbin (approved):

PR #856, opened 2023-05-12
Deprecate fog report id.

christian-oudard:
-


briancorbin (approved):
This looks good, just need to update unit / e2e tests cause they're failing

PR #860, opened 2023-05-16
Implement get_balance endpoint for API v2.

christian-oudard:
-


briancorbin (changes requested):
This looks good, but I would add the block height that it's reporting the balance for as part of the response, as the balance could be called during a partial or delayed sync

christian-oudard:
> This looks good, but I would add the block height that it's reporting the balance for as part of the response, as the balance could be called during a partial or delayed sync
Done.

briancorbin (approved):

PR #862, opened 2023-05-18
Implement search_ledger endpoint searching of txo id from the ledger db.

christian-oudard:
-

### DIFF
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -8,7 +8,6 @@ use diesel::{
 };
 use mc_account_keys::AccountKey;
 use mc_common::{logger::global_log, HashMap};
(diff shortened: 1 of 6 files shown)


full-service/src/db/txo.rs
```diff
@@ -101,8 +100,7 @@ pub struct TxoID(pub String);
 
 impl From<&TxOut> for TxoID {
     fn from(src: &TxOut) -> TxoID {
-        let digest: [u8; 32] = src.digest32::<MerlinTranscript>(b"txo_data");
-        Self(hex::encode(digest))
+        Self(hex::encode(src.hash()))
```
eranrund, line 103: Wouldn't this be a breaking change for existing databases?

PR #869, opened 2023-05-25
Filter by account and address

briancorbin:
This is fixing a bug where we weren't filtering properly on account as well as address index when querying for txos at a specific subaddress, so if there were multiple accounts in the same instance of full service and one tried to query for balance at a specific address, it would report all txos at the specified subaddress index across all accounts.


holtzman (approved):
Looks good.

PR #887, opened 2023-06-23
Allow getting transaction logs by txo_id in the v1 api for get_transaction_log.

christian-oudard:
-

### DIFF
--- a/full-service/src/json_rpc/v1/api/wallet.rs
+++ b/full-service/src/json_rpc/v1/api/wallet.rs
@@ -741,14 +741,39 @@ where
             .map_err(format_error)?,
         },
         JsonCommandRequest::get_transaction_log { transaction_log_id } => {
-            let (transaction_log, associated_txos, _) = service
-                .get_transaction_log(&transaction_log_id)
-                .map_err(format_error)?;
+            // Check whether the transaction_log_id actually refers to the txo_id of a
+            // received transaction.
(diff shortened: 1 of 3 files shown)


full-service/src/json_rpc/v1/api/wallet.rs
```diff
@@ -741,14 +741,39 @@ where
...
+                            subaddress_index,
+                        )
+                        .map(|assigned_sub| assigned_sub.public_address_b58)
+                        .ok()
+                    } else {
+                        None
+                    }
+                } else {
+                    None
+                };
```
briancorbin, line 763: This can be cleaned up by using a single match statement

```suggestion
                let subaddress_b58 = match (txo.subaddress_index, txo.account_id) {
                    (Some(subaddress_index), Some(account_id)) => {
                        let account_id = AccountID(account_id);
                        service
                            .get_address_for_account(&account_id, subaddress_index)
                            .map(|assigned_sub| assigned_sub.public_address_b58)
                            .ok()
                    }
                    _ => None,
                };
```

briancorbin (changes requested):

briancorbin:
Also, there's some lint issues to fix up

full-service/src/json_rpc/v1/api/wallet.rs
christian-oudard: Nice improvement.

briancorbin (approved):

PR #889, opened 2023-06-27
add public_key to v1 transaction_log txo_abbrevs

briancorbin:
This is to support more easily finding the public key for input and output txos in a transaction

### DIFF
--- a/full-service/src/json_rpc/v1/models/transaction_log.rs
+++ b/full-service/src/json_rpc/v1/models/transaction_log.rs
@@ -145,11 +145,6 @@ impl TransactionLog {
         txo: &db::models::Txo,
         assigned_address: Option<String>,
     ) -> Result<Self, String> {
-        let public_key_hex = match txo.public_key() {
-            Ok(pk) => hex::encode(pk.as_bytes()),
-            Err(_) => "".to_string(),
-        };
-
         Ok(TransactionLog {
             object: "transaction_log".to_string(),
             transaction_log_id: txo.id.clone(),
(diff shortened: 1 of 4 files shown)


full-service/src/json_rpc/v1/models/transaction_log.rs
```diff
@@ -249,6 +254,7 @@ impl TxoAbbrev {
             txo_id_hex: txo.id.clone(),
             recipient_address_id,
             value_pmob: (txo.value as u64).to_string(),
+            public_key: hex::encode(txo.public_key().unwrap().as_bytes()),
```
eranrund, line 257: unwrap here is sketchy, maybe default to an empty string if it fails?

full-service/src/json_rpc/v1/models/transaction_log.rs
```diff
@@ -159,6 +159,7 @@ impl TransactionLog {
                 txo_id_hex: txo.id.to_string(),
                 recipient_address_id: "".to_string(),
                 value_pmob: txo.value.to_string(),
+                public_key: hex::encode(txo.public_key().unwrap().as_bytes()),
```
eranrund, line 162: unwrap here is sketchy, maybe default to an empty string if it fails?

eranrund: Actually here you can just return an error since this already returns a `Result<>`, so that would be better.

eranrund (changes requested):
Should be possible to get rid of unwraps. Thanks!

eranrund (approved):

PR #914, opened 2023-09-06
Remove /app/target from build cache in Dockerfile

joekottke:
### Motivation

The RUN command to copy binaries after build doesn't work as the cache is not available in that step.  Removing the `/app/target` from the cache fixes this.

### In this PR
* Update to the dockerfile cargo build step, remove `/app/target` from cache directives

### Test Plan

* Tested locally

### Future Work
* None


briancorbin (approved):

briancorbin (approved):

jgreat:
These dockerfiles have been deprecated.  Closing.

PR #924, opened 2023-10-16
Feature/t3 transactions

briancorbin:
This PR adds a t3 api and connection crate to the full service repo

### TODO

- [x] README for t3 crates
- [x] finish implementations for t3 api methods

### FUTURE
- [ ] Add section to gitbook docs for how to use this


full-service/src/service/t3_sync.rs
```diff
@@ -0,0 +1,152 @@
...
+    // authenticated sender memo
+    log::debug!(logger, "Syncing txos to t3");
+
+    let txos = Txo::get_txos_that_need_to_be_synced_to_t3(Some(TXO_CHUNK_SIZE), conn)?;
+
+    for txo in txos {
+        let txo_memo = txo.memo(conn)?;
+        let memo = match txo_memo {
+            TxoMemo::AuthenticatedSender(memo) => memo,
+            _ => continue,
```
eranrund, line 123: isn't this an error since `get_txos_that_need_to_be_synced_to_t3` should only be returning txos with an AuthenticatedSender memo?

full-service/src/service/t3_sync.rs
```diff
@@ -0,0 +1,152 @@
...
+                            break;
+                        }
+
+                        match sync_txos(conn, &t3_connection, &logger) {
+                            Ok(()) => (),
+                            Err(e) => log::error!(&logger, "Error during t3 sync:\n{:?}", e),
+                        }
+
+                        // This sleep is to allow other API calls that need access to the database a
+                        // chance to execute, because the t3 sync process requires a write lock on
```
eranrund, line 79: where is this write lock?

full-service/src/service/t3_sync.rs
```diff
@@ -0,0 +1,152 @@
...
+#[derive(Clone, Debug, Parser)]
+pub struct T3Config {
+    #[clap(long, env = "T3_URI")]
+    pub t3_uri: T3Uri,
+    #[clap(long, env = "T3_API_KEY")]
+    pub t3_api_key: String,
+}
+
+// How many txos to sync per round
+const TXO_CHUNK_SIZE: usize = 5;
```
eranrund, line 35: any reason not to make this bigger?

full-service/src/service/t3_sync.rs
```diff
@@ -0,0 +1,152 @@
...
+    let txos = Txo::get_txos_that_need_to_be_synced_to_t3(Some(TXO_CHUNK_SIZE), conn)?;
+
+    for txo in txos {
+        let txo_memo = txo.memo(conn)?;
+        let memo = match txo_memo {
+            TxoMemo::AuthenticatedSender(memo) => memo,
+            _ => continue,
+        };
+
+        sync_txo(&txo, &memo, t3_connection)?;
```
eranrund, line 126: How does t3 behaves if the transaction has already been synced? e.g. what happens if we sync the same tx twice for some reason?

t3/api/proto/external/v1/external.proto
```diff
@@ -0,0 +1,309 @@
+/**
+ * DO NOT MODIFY THIS PROTOBUF
+ *
+ * This has been copied directly from the link below. This should likely never
+ * be modified in this repo, and instead will eventually be replaced by an
+ * externally published and versioned protobuf module. The only change is in
+ * the definition of the package name, which must be suffixed with a version
+ * number.
+ *
+ * https://github.com/mobilecoinofficial/android-sdk/blob/336b2efe4889ef21cfbc4e7e451b83aa1c13142a/android-sdk/src/main/proto/external.proto
```
eranrund, line 10: What does the android-sdk has to do with this repo? Why is this not copied from the main repo?

t3/api/proto/google/api/annotations.proto
```diff
@@ -0,0 +1,31 @@
+// Copyright 2015 Google LLC
```
eranrund, line 1: Can you include a README that explains where all these protos came from / how to update them?

t3/connection/src/error.rs
```diff
@@ -0,0 +1,21 @@
+use displaydoc::Display;
```
eranrund, line 1: Nit: missing copyright

t3/api/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2018-2022 MobileCoin, Inc.
```
eranrund, line 1:
```suggestion
// Copyright (c) 2018-2023 MobileCoin, Inc.
```

t3/connection/src/lib.rs
```diff
@@ -0,0 +1,157 @@
+mod error;
```
eranrund, line 1: Nit: missing copyright

t3/connection/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+        Self {
+            uri: uri.clone(),
+            api_key,
+            transaction_service_client,
+            logger,
+        }
+    }
+
+    pub fn find_transactions(&self) -> Result<FindTransactionsResponse, Error> {
+        let request = FindTransactionsRequest::new();
```
eranrund, line 61: Seems like this should be a parameter to this method

t3/connection/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+                log::warn!(
+                    self.logger,
+                    "t3_transaction_service find_transactions RPC call failed: {}",
+                    err
+                );
+                err
+            })?)
+    }
+
+    pub fn list_transactions() {}
```
eranrund, line 76: Remove or implement

t3/connection/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+impl PartialOrd for T3Connection {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.uri.addr().partial_cmp(&other.uri.addr())
+    }
+}
+
+impl Connection for T3Connection {
```
eranrund, line 148: Do you actually need this?

eranrund (changes requested):
Overall LGTM but has some nits/questions.
Thank you!

t3/connection/src/lib.rs
briancorbin: Ah, yeah I still need to go back and flesh out most of the t3 connection stuff, that's my immediate next task! Thanks for noting this missing :)

full-service/src/service/t3_sync.rs
```diff
@@ -0,0 +1,152 @@
...
+
+fn sync_txo(
+    txo: &Txo,
+    memo: &AuthenticatedSenderMemo,
+    t3_connection: &T3Connection,
+) -> Result<(), T3SyncError> {
+    let mut transparent_transaction = TransparentTransaction::new();
+
+    let sender_address_hash_bytes = hex::decode(&memo.sender_address_hash)?;
+    transparent_transaction.set_sender_address_hash(sender_address_hash_bytes);
```
starkriedesel, line 141: I don't see where we are setting the recipient address hash.

full-service/src/service/t3_sync.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    transparent_transaction.set_token_id(txo.token_id as u64);
+    transparent_transaction.set_amount(txo.value as u64);
+
+    let public_key_bytes = txo.public_key()?.as_bytes().to_vec();
+    transparent_transaction.set_public_key_hex(hex::encode(public_key_bytes));
+
+    let _ = t3_connection.create_transaction(transparent_transaction)?;
```
starkriedesel, line 149: In order to have parity with Moby we should also be filling in either `sender_institution` or `recipient_institution`. The institution enum value should be provided as a config arg (like the t3 api key) and then filled in either sender or recipient depending on if t3 is sending or receiving the transaction. If no enum is provided then `NONE` institution should be selected.

briancorbin: Could we just reference that based on the api key and information stored in the t3 database? It seems that for most use cases it wouldn't be applicable, only to the institutions that are established as part of the enum

starkriedesel: T3 server won't know if it was the sender or recipient address which belongs to the full service node. We could store the enum in the t3 server and return it to full service with some RPC but it would still be up to full service to assign the enum value to the correct field.

full-service/src/service/t3_sync.rs
briancorbin: That's a todo! Thanks 😊

full-service/Cargo.toml
```diff
@@ -70,6 +74,7 @@ grpcio = "0.12"
 hex = { version = "0.4", default-features = false }
 libsqlite3-sys = { version = "0.26", features = ["bundled-sqlcipher"] }
 num_cpus = "1.16"
+prost = "0.12"
```
remoun, line 77: Optional/Nit: `Cargo.lock` shows the rest of the project uses `prost 0.11.9`, so using 0.12 here winds up with 2 versions of that lib.
```suggestion
prost = "0.11"
```

full-service/src/db/txo.rs
```diff
@@ -802,32 +811,26 @@ impl TxoModel for Txo {
...
-                            diesel::insert_into(authenticated_sender_memos::table)
-                                .values(&new_memo)
-                                .execute(conn)?;
-                        }
-                        Err(e) => return Err(e.into()),
-                    }
-                }
-                _ => {} // The memo type is unknown, so don't record it to the database.
+        if let Ok(MemoType::AuthenticatedSender(memo)) = MemoType::try_from(&memo_payload) {
+            let new_memo = NewAuthenticatedSenderMemo {
```
remoun, line 815: Optional/Equivalent: Since this is only used in the `NotFound` branch below, move it in there

full-service/src/service/t3_sync.rs
remoun: The `t3.CreateTransaction` RPC returns `RESULT_ALREADY_EXISTS`, which can be treated equivalently to `RESULT_OK`

full-service/src/service/t3_sync.rs
remoun: Consider adding an explicit TODO
```suggestion
    transparent_transaction.set_sender_address_hash(sender_address_hash_bytes);
    // TODO: Set recipient_address_hash
```

t3/api/proto/pb/gubernator/gubernator.proto
remoun, line 1: Please remove this. Gubernator is only used by the server for rate-limiting, so clients don't need it.

t3/api/proto/t3/v1/admin.proto
remoun, line 1: FS doesn't need to use this AdminService, either.

t3/api/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+#[derive(Debug, Hash, Ord, PartialOrd, Eq, PartialEq, Clone)]
+pub struct T3Scheme {}
+impl UriScheme for T3Scheme {
+    /// The part before the '://' of a URL.
+    const SCHEME_SECURE: &'static str = "t3";
+    const SCHEME_INSECURE: &'static str = "insecure-t3";
+
+    /// Default port numbers
+    const DEFAULT_SECURE_PORT: u16 = 5553;
+    const DEFAULT_INSECURE_PORT: u16 = 5554;
```
remoun, line 33: IIUC these should be the HTTP/S ports
```suggestion
    const DEFAULT_SECURE_PORT: u16 = 443;
    const DEFAULT_INSECURE_PORT: u16 = 80;
```

t3/connection/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+        let mut request = CreateTransactionRequest::new();
+        request.set_transaction(transparent_transaction);
+
+        Ok(self
+            .transaction_service_client
+            .create_transaction_opt(&request, common_headers_call_option(&self.api_key))
+            .map_err(|err| {
+                log::warn!(
+                    self.logger,
+                    "t3_transaction_service find_transactions RPC call failed: {}",
```
remoun, line 91:
```suggestion
                    "T3.create_transaction RPC call failed: {}",
```

t3/connection/src/lib.rs
```diff
@@ -0,0 +1,157 @@
...
+    pub fn create_transaction(
+        &self,
+        transparent_transaction: TransparentTransaction,
+    ) -> Result<CreateTransactionResponse, Error> {
+        let mut request = CreateTransactionRequest::new();
+        request.set_transaction(transparent_transaction);
+
+        Ok(self
+            .transaction_service_client
+            .create_transaction_opt(&request, common_headers_call_option(&self.api_key))
```
remoun, line 87: It seems better for this method to check the `response.result` code and map it to a local `Error`. Maybe return `Result<(), Error>` or even `Result<TransparentTransaction, Error>`

t3/connection/src/lib.rs
briancorbin: That's true, although I wanted this to be a more standalone crate that isn't specific to full service, but a way for any rust app to connect to T3 in the future, so I think it's appropriate to have full functionality

full-service/src/service/t3_sync.rs
briancorbin: WIll do that in the future, bad habit that I don't! Thanks :)

t3/api/proto/t3/v1/admin.proto
briancorbin: This crate is more just for access to T3. I would like this to be a standalone crate in the future, so that any rust app can establish a connection to a T3 server

full-service/src/service/t3_sync.rs
briancorbin: I suppose an error or continue is fine? It should never happen, regardless, if we can trust that the function returning the txos is working as expected. Curious your thoughts on situations like this, does an error seem more appropriate?

full-service/src/service/t3_sync.rs
briancorbin: So, we had an issue during the account sync where if we didn't add a slight delay to the syncing, it would effectively keep relocking the database until the entire account was fully synced without ever letting anything else access the DB. I'm guessing the same thing is going to happen here, even though it's only locked during the single update (unlike the sync thread which creates an exclusive transaction for a large chunk of its operation).

I think 1000ms is actually too long and have lowered it to 10ms similar to the sync function, which should alleviate any potential issues.

Let me know if you have any thoughts on this approach though?

full-service/src/service/t3_sync.rs
briancorbin: I left it small during testing so I could iterate without resetting my database. Not sure there's much of a reason to keep it small, do you have a preference? It is really only impactful during the initial sync, which doesn't take long as it is. Under normal operation, it should only be syncing a couple txos at most at any given time.

t3/api/proto/external/v1/external.proto
briancorbin: No idea, this is what I found on buf.build

t3/api/proto/google/api/annotations.proto
briancorbin: Yes, definitely!

t3/connection/src/lib.rs
briancorbin: This seems to be the standard with the other Connection types found throughout the mobilecoin repo

t3/api/src/lib.rs
briancorbin: I will check the T3 docs for this

t3/connection/src/lib.rs
eranrund: I see, okay.

full-service/src/db/transaction_log.rs
```diff
@@ -1887,7 +1883,4 @@ mod tests {
 
         assert_eq!(tx_log.tx, expected_tx_log.tx);
     }
-
-    // FIXME: test_log_submitted for recovered
-    // FIXME: test_log_submitted offline flow
```
eranrund, line 1892: Do these tests exist?

full-service/src/db/txo.rs
```diff
@@ -950,6 +962,46 @@ impl TxoModel for Txo {
...
+    }
+
+    fn get_txos_that_need_to_be_synced_to_t3(
+        limit: Option<usize>,
+        conn: Conn,
+    ) -> Result<Vec<Txo>, WalletDbError> {
+        use crate::db::schema::txos;
+
+        let memo_types_predicate = txos::memo_type
+            .eq(Some(two_bytes_to_i32(
```
eranrund, line 982: Why does this `eq` requite `Some` but the others do not?

full-service/src/json_rpc/v1/api/wallet.rs
```diff
@@ -877,9 +891,9 @@ where
             }
         }
         JsonCommandRequest::get_txo { txo_id } => {
-            let (txo, status, _) = service.get_txo(&TxoID(txo_id)).map_err(format_error)?;
+            let txo_info = service.get_txo(&TxoID(txo_id)).map_err(format_error)?;
             JsonCommandResponse::get_txo {
-                txo: Txo::new(&txo, &status),
+                txo: Txo::new(&txo_info.txo, &txo_info.status),
```
eranrund, line 896: Might be cleaner to `impl From<&TxoInfo> for Txo` so that you can just do `Txo::from(&txo_info)` instead of having to repeat this

full-service/src/service/t3_sync.rs
eranrund: 5 every 10msec seems fine to me

full-service/src/service/t3_sync.rs
```diff
@@ -0,0 +1,164 @@
...
+    #[clap(long, env = "T3_API_KEY")]
+    pub t3_api_key: String,
+}
+
+// How many txos to sync per round
+// TODO - Discuss in PR if this is a reasonable value
+const TXO_CHUNK_SIZE: usize = 5;
+// How long to wait in milliseconds between sync rounds
+// TODO - discuss in PR if this is a reasonable value
+const T3_SYNC_INTERVAL: u64 = 10;
```
eranrund, line 36: any reason not to use the `Duration` type?

full-service/src/service/t3_sync.rs
eranrund: I think the only reason the sleep is needed here is because we don't know if we have more transactions to sync or not and without sleep this loop would just take 100% CPU.

t3/api/src/lib.rs
eranrund: Still shows as 2022

eranrund (commented):
Thanks for addressing the previous review comments, this looks good.
I'm ready to approve once Stark has had a chance to also review.

t3/api/src/lib.rs
eranrund: I have a feeling @remoun is right. I wouldn't count on the docs, I would look to see which port Moby is connecting to

full-service/src/db/transaction_log.rs
briancorbin: No, they don't, but I don't think they are applicable. This is from a very long time ago. I need to go through and clean up all of the TODO's and FIXME's, there are some that just don't apply anymore to the codebase.

full-service/src/db/txo.rs
briancorbin: That may have been from my autocomplete, I don't think the Some is required

remoun (approved):

starkriedesel (approved):
I'm good with these changes assuming we get sender address hash resolved in another PR and have some plan for the institution identifier.

tools/lint-fix.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Copyright (c) 2018-2020 MobileCoin Inc.
+
+set -e
+
+if [[ ! -z "$1" ]]; then
+    cd "$1"
+fi
+
+for toml in $(grep --exclude-dir cargo --exclude-dir rust-mbedtls --exclude-dir mobilecoin --exclude-dir ledger-mob --include=Cargo.toml -r . -e '\[workspace\]' | cut -d: -f1); do
```
remoun, line 11: Optional/Equivalent: Replace `cut` with `grep` flag `-l`
```suggestion
for toml in $(grep --exclude-dir cargo --exclude-dir rust-mbedtls --exclude-dir mobilecoin --exclude-dir ledger-mob --include=Cargo.toml -r . -e '\[workspace\]' -l); do
```

eranrund (approved):
LGTM once @remoun's comments are addressed.

briancorbin:
@starkriedesel could you take one more look at this since updating the protobuf and let me know if it looks good to you?

PR #925, opened 2023-10-27
Feature/txo sender memos/fs 37

briancorbin:
This adds authenticated sender memos to full service, which is a precursor for t3 integration into full service. A new table has been added to the database to track the information in the encrypted memo so that queries can be performed against the data using SQL.


full-service/src/json_rpc/v2/models/memo.rs
```diff
@@ -0,0 +1,39 @@
...
+            payment_intent_id: memo.payment_intent_id.clone(),
+        }
+    }
+}
+
+impl From<&TxoMemo> for Memo {
+    fn from(memo: &TxoMemo) -> Self {
+        match memo {
+            TxoMemo::Sender(memo) => Memo::Sender(memo.into()),
+            _ => Memo::Unused,
```
eranrund, line 36: Can you use `TxoMemo::Unused` here? Using `_` might result in forgetting to update this code if a new memo type is added.

full-service/src/json_rpc/v2/models/txo.rs
```diff
@@ -132,8 +141,11 @@ mod tests {
         let status = txo_details
             .status(&mut wallet_db.get_pooled_conn().unwrap())
             .unwrap();
+        let memo = txo_details
+            .memo(&mut wallet_db.get_pooled_conn().unwrap())
+            .unwrap();
         assert_eq!(txo_details.value as u64, 15_625_000 * MOB);
-        let json_txo = Txo::new(&txo_details, &status);
+        let json_txo = Txo::new(&txo_details, &status, &memo);
```
eranrund, line 148: Should anything be asserted on the memo in the json_txo?

full-service/src/service/memo.rs
```diff
@@ -0,0 +1,120 @@
+use crate::{
```
eranrund, line 1: Nit: Copyright

full-service/src/service/memo.rs
```diff
@@ -0,0 +1,120 @@
...
+    B58(B58Error),
+
+    /// Error decoding account key: {0}
+    Decode(mc_util_serial::DecodeError),
+
+    /// Error interacting with ledger database: {0}
+    Ledger(LedgerServiceError),
+
+    /// Key Error
+    Key(mc_crypto_keys::KeyError),
```
eranrund, line 35: Nit: This should be consistent with how we typically name and format the error variants

```suggestion
    /// WalletDb: {0}
    WalletDb(WalletDbError),

    /// B58: {0}
    B58(B58Error),

    /// Decode: {0}
    Decode(mc_util_serial::DecodeError),

    /// Ledger: {0}
    Ledger(LedgerServiceError),

    /// Key: {0}
    Key(mc_crypto_keys::KeyError),
```

full-service/src/service/memo.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+impl From<mc_crypto_keys::KeyError> for MemoServiceError {
+    fn from(src: mc_crypto_keys::KeyError) -> Self {
+        Self::Key(src)
+    }
+}
+
+#[rustfmt::skip]
```
eranrund, line 68: whats this for?

full-service/src/service/memo.rs
```diff
@@ -0,0 +1,120 @@
...
+        let txo = self.get_txo_object(txo_id_hex)?;
+        let shared_secret =
+            account.get_shared_secret(&RistrettoPublic::try_from(&txo.public_key)?)?;
+        let memo_payload = match txo.e_memo {
+            Some(e_memo) => e_memo.decrypt(&shared_secret),
+            None => UnusedMemo.into(),
+        };
+        Ok(if let Ok(memo_type) = MemoType::try_from(&memo_payload) {
+            match memo_type {
+                MemoType::AuthenticatedSender(memo) => memo
```
eranrund, line 107: What about the two other types?

full-service/src/service/memo.rs
```diff
@@ -0,0 +1,120 @@
...
+
+        let txo = self.get_txo_object(txo_id_hex)?;
+        let shared_secret =
+            account.get_shared_secret(&RistrettoPublic::try_from(&txo.public_key)?)?;
+        let memo_payload = match txo.e_memo {
+            Some(e_memo) => e_memo.decrypt(&shared_secret),
+            None => UnusedMemo.into(),
+        };
+        Ok(if let Ok(memo_type) = MemoType::try_from(&memo_payload) {
+            match memo_type {
```
eranrund, line 106: Nit: Simplify this by matching on `Ok(MemoType::AuthenticatedSender(memo)))`

full-service/src/service/receipt.rs
```diff
@@ -187,7 +187,7 @@ pub trait ReceiptService {
         &self,
         address: &str,
         receiver_receipt: &ReceiverReceipt,
-    ) -> Result<(ReceiptTransactionStatus, Option<(Txo, TxoStatus)>), ReceiptServiceError>;
+    ) -> Result<(ReceiptTransactionStatus, Option<(Txo, TxoStatus, TxoMemo)>), ReceiptServiceError>;
```
eranrund, line 190: The followup PR moves from this tuple to `TxoInfo`?

full-service/src/service/transaction_builder.rs
```diff
@@ -228,7 +229,11 @@ impl<FPR: FogPubkeyResolver + 'static> WalletTransactionBuilder<FPR> {
         let (fee, fee_token_id) = self.fee.unwrap_or((Mob::MINIMUM_FEE, Mob::ID));
         let fee_amount = Amount::new(fee, fee_token_id);
         let fog_resolver = self.get_fog_resolver(conn)?;
-        let memo_builder = memo.memo_builder(account)?;
+
+        let memo_builder = match account.account_key() {
+            Ok(account_key) => memo.memo_builder(&account_key),
+            Err(_) => Box::<EmptyMemoBuilder>::default(),
```
eranrund, line 235: Should probably log this if possible, or maybe just return an error? When is `account.account_key()` expected to fail?

python/mobilecoin/cli.py
```diff
@@ -421,10 +422,27 @@ def remove(self, account_id):
...
+                print()
+                print('id:', txo['id'])
+                print('amount:', Amount.from_storage_units(txo['value'], txo['token_id']))
+                print('subaddress:', txo['subaddress_index'])
+                print('received in block:', txo['received_block_index'])
+                print('spent in block:', txo['spent_block_index'])
+                print('status:', txo['status'])
+
+                mcp_txo = self.client.get_mc_protocol_txo(txo['id'])
+                print(mcp_txo['txo']['e_memo'])
```
eranrund, line 443:
```suggestion
                print('e_memo:', mcp_txo['txo']['e_memo'])
```

eranrund (changes requested):
Got some small things/some questions but overall I think this is good.
Is the idea to get this merged before https://github.com/mobilecoinofficial/full-service/pull/924?

.github/workflows/ci.yml
```diff
@@ -61,20 +61,26 @@ jobs:
           cargo clippy --all --all-features
 
   test:
-    runs-on: [self-hosted, Linux, large]
-    container:
-      image: mobilecoin/rust-sgx-base:v0.0.28
+    runs-on: [self-hosted, macOS, X64, cargo]
```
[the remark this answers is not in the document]
briancorbin: I'm using my own computer to test since I'm not able to rely on ones that we have. They are seemingly running out of memory randomly, like 55% of the time. Sometimes they will work for days at a time, and then suddenly stop. 🤷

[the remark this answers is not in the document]
briancorbin: So this is a self hosted runner on my work computer

full-service/src/service/transaction_builder.rs
briancorbin: When it is a view only account

Cargo.toml
```diff
@@ -46,6 +46,9 @@ schnorrkel-og = { git = "https://github.com/mobilecoinfoundation/schnorrkel.git"
 # Fix issues with recent nightlies, bump curve25519-dalek version
 x25519-dalek = { git = "https://github.com/mobilecoinfoundation/x25519-dalek.git", rev = "4fbaa3343301c62cfdbc3023c9f485257e6b718a" }
 
+# Fix linker errors when compiling gRPC C++ library on systemd based Linux.
+grpcio = { git = "https://github.com/tikv/grpc-rs.git", rev = "cc5d2ab4fa0a110d0fdaa49f12552d28a0348721" }
```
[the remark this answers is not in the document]
briancorbin: I would like to leave this on 0.12 until the mobilecoin repo has upgraded to 0.13 and instead just have it be patched for now.

full-service/migrations/2023-08-14-214222_authenticated_sender_memos/up.sql
```diff
@@ -0,0 +1,11 @@
...
+
+CREATE TABLE authenticated_sender_memos (
+  txo_id TEXT PRIMARY KEY NOT NULL,
+  sender_address_hash TEXT NOT NULL,
+  payment_request_id UNSIGNED BIG INT,
+  payment_intent_id UNSIGNED BIG INT,
+  FOREIGN KEY (txo_id) REFERENCES txos(id)
+);
+
+UPDATE accounts SET next_block_index = 0;
```
remoun, line 11: This seems out of place here. If it's a temp change for local dev, please mark it as such
```suggestion
// DO NOT MERGE: local testing
UPDATE accounts SET next_block_index = 0;

```

full-service/src/service/memo.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn from(e: KeyError) -> Self {
+        Self::Key(e)
+    }
+}
+
+impl From<MemoDecodingError> for MemoServiceError {
+    fn from(e: MemoDecodingError) -> Self {
+        Self::MemoDecoding(e)
+    }
+}
```
remoun, line 73: IIUC we can replace these `impl`s with `#[from]` in the `enum` declaration

full-service/src/test_utils.rs
```diff
@@ -378,11 +378,37 @@ pub fn create_test_txo_for_recipient(
...
     amount: Amount,
     rng: &mut StdRng,
+) -> (TxOut, KeyImage) {
+    return create_test_txo_for_recipient_with_memo(
+        recipient_account_key,
+        recipient_subaddress_index,
+        amount,
+        rng,
+        TransactionMemo::Empty,
+    );
```
remoun, line 388: Optional/Equivalent: Omit `return`
```suggestion
    create_test_txo_for_recipient_with_memo(
        recipient_account_key,
        recipient_subaddress_index,
        amount,
        rng,
        TransactionMemo::Empty,
    )
```

remoun (approved):

full-service/src/service/memo.rs
```diff
@@ -0,0 +1,123 @@
...
+    WalletService,
+};
+use mc_account_keys::AccountKey;
+use mc_connection::{BlockchainConnection, UserTxConnection};
+use mc_crypto_keys::{KeyError, RistrettoPublic};
+use mc_fog_report_validation::FogPubkeyResolver;
+use mc_transaction_extra::{MemoDecodingError, MemoType, UnusedMemo};
+use mc_util_serial::DecodeError;
+use std::{convert::TryFrom, ops::DerefMut};
+use thiserror::Error;
```
eranrund, line 18: Why this and not `displaydoc` like we've used everywhere else?

full-service/src/util/b58/errors.rs
```diff
@@ -1,29 +1,29 @@
-use displaydoc::Display;
+use thiserror::Error;
```
eranrund, line 1: Whats the reason for this change?

full-service/src/service/memo.rs
briancorbin: So, it requires that the error we are going from conforms to std::Error, otherwise this is still required. The goal is to make as many things conform to that as possible so we can use the #[from] macro, but that will be a later PR that is focused on that.

full-service/src/service/memo.rs
briancorbin: starting to move over to thiserror. It will reduce the cumbersome and boilerplate impl From<ErrorType> for <ErrorType>

full-service/migrations/2023-08-14-214222_authenticated_sender_memos/up.sql
eranrund: How was this tested?
How do we know this doesn't break integrations that don't expect the next_block_index/account_block_height to ever decrease?

full-service/src/util/b58/errors.rs
eranrund: FYI the reason we did not use `thiserror` in the `mobilecoin` repo is that it is not no_std compatible and cannot be used in enclaves or any other no_std code (e.g. wasm).
This might not affect `full-service` but most/all error types in `mobilecoin` do not implement `std::error::Error`, and if there are ever any parts in this repository that will need to be no_std, `thiserror` will have to be removed from them.

Cargo.toml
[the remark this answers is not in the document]
eranrund: This is now in https://github.com/mobilecoinfoundation/mobilecoin/tree/release/v5.1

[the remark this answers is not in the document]
briancorbin: oh? Awesome, I'll update to that! Thank you :)

full-service/migrations/2023-08-14-214222_authenticated_sender_memos/up.sql
briancorbin: I have tested it several times, it works as expected.

As for not breaking integrations, I don't feel like that's something to worry about. Do you think it's reasonable to assume that someone would be using this in a way that would cause their integration to break? Secondly, do we care if that is the case?

If we care, it's possible to do this another way but will require more work. I'm fine doing that but it doesn't seem worth the effort since the account will resync in a few minutes at current block height.

The other proposed solution was to run a query after the migration and iterate over all of the txos, manually calling the update, which would avoid requiring a resync.

full-service/src/util/b58/errors.rs
briancorbin: aaah, I see. I don't expect Full Service to ever run into that requirement, but if you think it's worth avoiding it altogether, I am fine with that.

.github/workflows/ci.yml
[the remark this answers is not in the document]
briancorbin: Will try to resolve this another way. That said, if another PR came through it would just run using my runner on my work computer(s)

full-service/src/util/b58/errors.rs
briancorbin: will revert this regardless and address thiserror stuff at a later point

full-service/src/service/memo.rs
briancorbin: no longer relevant, will be address thiserror in a different PR potentially

full-service/migrations/2023-08-14-214222_authenticated_sender_memos/up.sql
eranrund: I have no idea how people are using full-service, but given that it is used by exchanges we very much care about breaking things for them.
This feels like a questionable approach to me, for example it appears that right now if you did it twice in a row (`UPDATE accounts SET next_block_index = 0;`, start full-service and let it do its thing, `UPDATE accounts SET next_block_index = 0;` again) you will misbehave since `add_authenticated_memo_to_database` will be called twice.

I am worried of any other potential incidents where something like this might get missed.

How difficult would it be to add code before [running migrations](https://github.com/mobilecoinofficial/full-service/blob/main/full-service/src/bin/main.rs#L85) that checks if this migration is about to run, and if so then after it runs goes and processes all known TxOuts in the database for their memos?

Its possible I am being over-conservative but my reason for this is that the cost of accidentally breaking one of the exchanges integrations or databases because something we didn't predict had happened would be very high...

briancorbin: It isn't much work, maybe a few hours at most? I do agree that we want to avoid breaking things especially for exchanges, so I'll go ahead and implement this instead :) Thanks for being on top of these types of potential issues and bringing them up, appreciate it!

eranrund: Thank you!

mirror/Cargo.toml
```diff
@@ -49,4 +53,7 @@ mc-util-build-script = { path = "../mobilecoin/util/build/script" }
 cargo-emit = "0.2.1"
 # Even though this is unused, it needs to be here otherwise Cargo brings in some weird mixture of packages/features that refuses to compile.
 # Go figure ¯\_(ツ)_/¯
-serde = { version = "1", default-features = false, features = ["alloc", "derive"] }
+serde = { version = "1", default-features = false, features = [
+    "alloc",
+    "derive",
+] }
```
eranrund, line 59: Nit: We haven't been putting new lines in our dependencies, not sure what caused this change

eranrund (commented):
Is this missing the migration code?
I am seeing a new table `post_migration_processes` but nothing appears to be using it. And on that topic, why is that needed? Can't the `__diesel_schema_migrations` table be used to tell if `full-service/migrations/2023-08-14-214222_authenticated_sender_memos/up.sql` still needs to run?

briancorbin:
> Is this missing the migration code? I am seeing a new table `post_migration_processes` but nothing appears to be using it. And on that topic, why is that needed? Can't the `__diesel_schema_migrations` table be used to tell if `full-service/migrations/2023-08-14-214222_authenticated_sender_memos/up.sql` still needs to run?

So, my reason against that was if for any reason the operation gets interrupted, the next time full service starts it wouldn't finish because the migration would have already been applied. The other issue is that it will require the ledger_db, which is not available at the time of the migrations (it may be with some refactoring, but I didn't want to go down that path unnecessarily).

I have also opted not to do a backfill of the info after chatting with stark and henry, it seems that information really isn't too necessary, so long as it works going forward.

mirror/Cargo.toml
briancorbin: linter did this automatically

briancorbin: I can try to revert and see if it cares

full-service/src/db/txo.rs
```diff
@@ -1987,9 +2069,81 @@ impl TxoModel for Txo {
...
+        Ok(MemoType::AuthenticatedSenderWithPaymentRequestId(memo)) => {
+            add_authenticated_memo_to_database(
+                txo_id,
+                &memo.sender_address_hash().to_string(),
+                Some(memo.payment_request_id() as i64),
+                None,
+                conn,
+            )
+        }
+        Ok(_) => Ok(()),
```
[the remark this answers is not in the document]
briancorbin: When it's any of the other memo types available as defined in the mc transaction crate. We will eventually care about this information in a future update, but for now it is to be ignored. There will soon be a follow up PR that adds all of the DestinationMemo stuff, which would cause this to be updated

eranrund:
> I have also opted not to do a backfill of the info after chatting with stark and henry, it seems that information really isn't too necessary, so long as it works going forward.

So that implies `post_migration_processes` will now go away?

eranrund (approved):

PR #926, opened 2023-10-31
Adding shared secret to output txo

briancorbin:
### In this PR
An optional (for backwards compatibility with v1 json API) shared_secret field has been added to the service model for OutputTxo's. This enables us to re-decrypt the txo from the ledger, including its memo, at a future time when deemed necessary.

This will directly support the T3 integration by allowing authenticated sender memos of sent txos to be recorded

### Test Plan

Unit tests

### Future Work
Adjusting upstream PRs to utilize the new shared_secret information if available

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -592,14 +592,10 @@ impl TransactionLogModel for TransactionLog {
                 .execute(conn)?;
         }
 
-        // TODO - Get each payload txo and add it to the transaction_output_txos
-        // table for this transactions.
         for payload_txo in unsigned_tx_proposal.payload_txos.iter() {
             Txo::create_new_output(payload_txo, false, &transaction_log_id, conn)?;
         }
 
-        // TODO - Get each change txo and add it to the transaction_output_txos
-        // table for this transaction as change.
         for change_txo in unsigned_tx_proposal.change_txos.iter() {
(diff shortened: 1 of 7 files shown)


eranrund:
Is there a reason this needs to be stored in the database?
Unless I'm mistaken, it can always be calculated since you have the private view key and the txo pubkey.

briancorbin:
> > Is there a reason this needs to be stored in the database? Unless I'm mistaken, it can always be calculated since you have the private view key and the txo pubkey.
> 
> This is for the shared secret of a txo that you don't own (output of a transaction), which can be created with either the txo_private_key (at the time of generating the txo) + recipients view public key (in the public address of the recipient) OR the txo_public_key (available on the txo itself) + recipients view private key (which requires the private account keys).

https://github.com/mobilecoinfoundation/mobilecoin/blob/08d9f69d0c7e27faabbb0127912b330910239eae/transaction/builder/src/transaction_builder.rs#L868-L888

for reference

eranrund:
Gotcha, yeah, I missed that this is for TxOuts you are sending.

eranrund (approved):

briancorbin:
> Gotcha, yeah, I missed that this is for TxOuts you are sending.

😅 oh thank goodness, for a second I thought I was going crazy, hahah. Thanks so much!

PR #931, opened 2023-11-11
Account resyncing

briancorbin:
Forcing each account in the wallet to resync with the ledger from block index 0, which will cause each one to scan txos and pick up any Authorized Sender Memos that are found.

During this sync, normal API calls will be unavailable, and an error will be returned with the current percentage completed of the sync.

### DIFF
--- a/full-service/src/json_rpc/v1/api/wallet.rs
+++ b/full-service/src/json_rpc/v1/api/wallet.rs
@@ -120,12 +120,8 @@ where
     if service.resync_in_progress().map_err(format_error)? {
         let wallet_status = service.get_wallet_status().map_err(format_error)?;
 
-        let percent_complete = (wallet_status.min_synced_block_index as f64
-            / wallet_status.local_block_height as f64
-            * 100.0) as u64;
-
-        return Err(format_error(&format!(
-            "Resync in progress, please wait until it is completed to perform API calls... ({percent_complete}% complete)"
+        return Err(format_error(format!(
+            "Resync in progress, please wait until it is completed to perform API calls... ({}% complete)", wallet_status.percent_synced()
         )));
     }
 
--- a/full-service/src/json_rpc/v1/api/wallet.rs
+++ b/full-service/src/json_rpc/v1/api/wallet.rs
@@ -117,6 +117,18 @@ where
 {
     global_log::info!("Running command {:?}", command);
 
+    if service.resync_in_progress().map_err(format_error)? {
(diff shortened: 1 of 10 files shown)


full-service/src/json_rpc/v1/api/wallet.rs
```diff
@@ -117,6 +117,18 @@ where
 {
     global_log::info!("Running command {:?}", command);
 
+    if service.resync_in_progress().map_err(format_error)? {
+        let wallet_status = service.get_wallet_status().map_err(format_error)?;
+
+        let percent_complete = (wallet_status.min_synced_block_index as f64
+            / wallet_status.local_block_height as f64
+            * 100.0) as u64;
```
eranrund, line 125: You should be able to skip the floating point stuff by doing `wallet_status.min_synced_block_index * 100 / wallet_status.local_block_height`

full-service/src/json_rpc/v2/api/wallet.rs
```diff
@@ -139,6 +139,18 @@ where
 {
     global_log::info!("Running command {:?}", command);
 
+    if service.resync_in_progress().map_err(format_error)? {
+        let wallet_status = service.get_wallet_status().map_err(format_error)?;
+
+        let percent_complete = (wallet_status.min_synced_block_index as f64
+            / wallet_status.local_block_height as f64
+            * 100.0) as u64;
```
eranrund, line 147: Since this is duplicated it should be a method on the `WalletStatus` struct

full-service/src/json_rpc/v2/api/wallet.rs
```diff
@@ -139,6 +139,18 @@ where
 {
     global_log::info!("Running command {:?}", command);
 
+    if service.resync_in_progress().map_err(format_error)? {
```
eranrund, line 142: Its worth mentioning here that even a single account resyncing would block all API access, even if other accounts are fully synced. I think its acceptable at the moment since resyncing only takes place when the db migration runs, but if we ever added some method to force resyncing of a single account it might be odd that it blocks all access until thats done.

eranrund (approved):

full-service/src/json_rpc/v2/api/wallet.rs
briancorbin: Agreed. We currently do have a method for forcing a single account to resync that does _not_ cause the API to lock up, which is expected and intended behavior. Happy to revisit this if you think it's necessary

PR #936, opened 2023-11-17
Update mirror endpoints

briancorbin:
Updating the mirror to include validate sender memo as a whitelisted API call

Updating the validate sender memo to not require account id, as it can be derived from the txo that is being validated

### DIFF
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -2159,7 +2159,7 @@ impl TxoModel for Txo {
             .as_ref()
             .map(|account_id| Account::get(&AccountID(account_id.to_string()), conn))
             .transpose()?)
-     }
+    }
 }
 
 fn add_authenticated_memo_to_database(
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -2154,14 +2154,12 @@ impl TxoModel for Txo {
     }
 
     fn account(&self, conn: Conn) -> Result<Option<Account>, WalletDbError> {
-        match &self.account_id {
-            Some(account_id) => Ok(Some(Account::get(
-                &AccountID(account_id.to_string()),
-                conn,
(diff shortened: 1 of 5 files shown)


full-service/src/db/txo.rs
```diff
@@ -2150,6 +2152,16 @@ impl TxoModel for Txo {
...
+
+    fn account(&self, conn: Conn) -> Result<Option<Account>, WalletDbError> {
+        match &self.account_id {
+            Some(account_id) => Ok(Some(Account::get(
+                &AccountID(account_id.to_string()),
+                conn,
+            )?)),
+            None => Ok(None),
+        }
+    }
```
eranrund, line 2164: Cleaner imo

```suggestion
        Ok(self
            .account_id
            .as_ref()
            .map(|account_id| Account::get(&AccountID(account_id.to_string()), conn))
            .transpose()?)
     }
```

full-service/src/service/memo.rs
```diff
@@ -102,13 +95,18 @@ where
         let mut pooled_conn = self.get_pooled_conn()?;
         let conn = pooled_conn.deref_mut();
 
-        let account = Account::get(account_id, conn)?;
-        let account_key: AccountKey = mc_util_serial::decode(&account.account_key)?;
+        let txo = Txo::get(txo_id_hex, conn)?;
+        let account = match txo.account(conn)? {
+            Some(account) => account,
+            None => return Ok(false),
+        };
```
eranrund, line 102: Cleaner imo
```suggestion
        let Some(account) = txo.account(conn)? else {
            return Ok(false);
        };
```

eranrund (approved):

full-service/src/db/txo.rs
briancorbin: agreed, thanks!

PR #938, opened 2023-11-27
update default secure port to 443 for T3 URI

briancorbin:
Default port was incorrectly set to 80 before, it should be 443


starkriedesel (approved):
LGTM

PR #944, opened 2024-01-07
validate sender memo using recipient subaddress

holtzman:
### Motivation

validate_sender_memo currently always returns false because it is using the recipient's base view private key to recreate the memo's HMAC for matching against the one stored in the memo. Instead, the subaddress view private key for the subaddress that received the txo should be used.

### In this PR

This PR adds looking up the subaddress index for the address which received the txo and using it to pass the recipient's subaddress view private key to the authenticated memo validator.

It also adds a new internal error in case the txo is orphaned (no subaddress index for the txo in the walletdb) which could happen if the account is imported after having been used elsewhere and insufficient subaddress indexes to cover all of the previously assigned addresses were initialized.  We consider this an error rather than returning false to the validation because whether or not the sender_address is valid is unknown, and the full-service operator can correct the situation by generating more assigned addresses until there are no longer orphaned txxos.

### DIFF
--- a/full-service/src/service/memo.rs
+++ b/full-service/src/service/memo.rs
@@ -83,19 +83,19 @@ pub trait MemoService {
     ) -> Result<bool, MemoServiceError>;
 }
 
-// validate_sender_memo
-//
-// validating the sender memo includes:
-// 1. Is there a sender memo?
-// 2. Does the provided sender public address hash to the
-//    same value as the memo's sender address hash?
-// 3. When we recreate the HMAC, does it match the
-//    HMAC conveyed in the memo?
 impl<T, FPR> MemoService for WalletService<T, FPR>
 where
     T: BlockchainConnection + UserTxConnection + 'static,
     FPR: FogPubkeyResolver + Send + Sync + 'static,
 {
+    // validate_sender_memo
+    //
+    // validating the sender memo includes:
+    // 1. Is there a sender memo?
+    // 2. Does the provided sender public address hash to the
+    //    same value as the memo's sender address hash?
+    // 3. When we recreate the HMAC, does it match the
+    //    HMAC conveyed in the memo?
     fn validate_sender_memo(
         &self,
         txo_id_hex: &str,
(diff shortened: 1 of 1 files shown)


full-service/src/service/memo.rs
```diff
@@ -80,6 +83,14 @@ pub trait MemoService {
...
 }
 
+// validate_sender_memo
+//
+// validating the sender memo includes:
+// 1. Is there a sender memo?
+// 2. Does the provided sender public address hash to the
+//    same value as the memo's sender address hash?
+// 3. When we recreate the HMAC, does it match the
+//    HMAC conveyed in the memo?
```
eranrund, line 93: This should be right above the `fn validate_sender_memo` line

full-service/src/service/memo.rs
```diff
@@ -100,6 +111,15 @@ where
...
         };
 
+        // validating the HMAC requires the receipient's subaddress
+        // view private key, so fetch the recipient subaddress_index
+        // of the txo, and fail if this is not available (orphaned txo)
+        let subaddress_index = match txo.subaddress_index {
+            Some(subaddress_index) => subaddress_index as u64,
+            // None => return Ok(false),
+            None => return Err(MemoServiceError::TxoOrphaned(txo_id_hex.to_string())),
+        };
```
eranrund, line 121: This will be a bit more elegantly written like this:
```suggestion
        let Some(subaddress_index) = txo.subaddress_index else {
            return Err(MemoServiceError::TxoOrphaned(txo_id_hex.to_string()));
        };
```

eranrund (approved):
Small nits but other than that lgtm

cbeck88:
I would like to suggest that you write some tests on this

full-service/src/service/memo.rs
holtzman: moved! Thanks for the suggestion.

full-service/src/service/memo.rs
holtzman: did it that way because I wanted to cast to subaddress_index to u64 on assignment, but I've moved that to when used instead.  Also, meant to delete the commented out piece!

PR #949, opened 2024-01-22
add the txo public key to a few api response structs

eranrund:
Callers can benefit from having easy access to TxOut public keys since that is often a common method for checking if a TxOut has been included in the ledger.

### DIFF
--- a/full-service/src/json_rpc/v2/models/tx_proposal.rs
+++ b/full-service/src/json_rpc/v2/models/tx_proposal.rs
@@ -13,6 +13,7 @@ use std::convert::TryFrom;
 #[derive(Deserialize, Serialize, Default, Debug, PartialEq)]
 pub struct UnsignedInputTxo {
     pub tx_out_proto: String,
+    pub tx_out_public_key: String,
     pub amount: AmountJSON,
     pub subaddress_index: String,
 }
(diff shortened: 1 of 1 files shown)


kwon-t (approved):
LGTM. thank you!

full-service/src/json_rpc/v2/models/tx_proposal.rs
```diff
@@ -54,6 +57,7 @@ impl TryFrom<&crate::service::models::tx_proposal::UnsignedTxProposal> for Unsig
             .iter()
             .map(|input_txo| UnsignedInputTxo {
                 tx_out_proto: hex::encode(mc_util_serial::encode(&input_txo.tx_out)),
+                tx_out_public_key: hex::encode(input_txo.tx_out.public_key.as_bytes()),
```
nick-mobilecoin, line 60: Based on `input_txo.tx_out` and `input_txo.tx_out.public_key` it seems that `tx_out_public_key` is already in this payload, do we really want to duplicate data like this.
Guessing it's the lack of library like interface to be able to have something that can easily get to this nested field

nick-mobilecoin (approved):
LGTM if we're find ducplicating data that's already in the message

full-service/src/json_rpc/v2/models/tx_proposal.rs
eranrund: If you want to get it out of the proto you need to have the ability to parse that proto, which is not ergonomic.
I don't see an issue with duplicating some data if it makes the life of clients easier.
I think part of the design philosophy behind full-service is to provide a user-friendly API and having to parse a proto out of hex bytes inside a JSON response in order to get a commonly used field seems out of place to me.

PR #955, opened 2024-02-05
Feature/short address hash

holtzman:
### Motivation

full-service decrypts and returns sender and destination memos as part of the response when using the `get_txo` and `get_txos` API request methods.  These memo formats identify the sender or recipient of a MobileCoin transaction to enable recovering transaction history from the blockchain. To save space in the ledger, as well as preserve privacy, the counterparty is identified via a `short address hash` of their full public address. Users of the full-service API need a way to generate the `short address hash` from the B58 public addresses that they have for their contacts, to aid in identifying transaction counterparties via these memos.

(Side note: full-service can only decrypt the memo on txos that are owned by accounts for which it has the view private key.)

### In this PR

We `add short_address_hash` as part of the response to `verify_address` and `get_address_details`, two methods that full-service provides for working with B58 Public Addresses.

### DIFF
--- a/full-service/src/json_rpc/v2/api/response.rs
+++ b/full-service/src/json_rpc/v2/api/response.rs
@@ -102,6 +102,7 @@ pub enum JsonCommandResponse {
     },
     get_address_details {
         details: PublicAddress,
+        short_address_hash: String,
     },
     get_address {
         address: Address,
--- a/full-service/src/json_rpc/v2/api/response.rs
+++ b/full-service/src/json_rpc/v2/api/response.rs
@@ -102,7 +102,7 @@ pub enum JsonCommandResponse {
     },
     get_address_details {
         details: PublicAddress,
-        short_address_hash: String,
+        address_hash: String,
     },
     get_address {
         address: Address,
@@ -215,7 +215,7 @@ pub enum JsonCommandResponse {
     },
     verify_address {
         verified: bool,
-        short_address_hash: Option<String>,
+        address_hash: Option<String>,
     },
     version {
         string: String,
--- a/full-service/src/json_rpc/v2/api/response.rs
+++ b/full-service/src/json_rpc/v2/api/response.rs
@@ -215,6 +215,7 @@ pub enum JsonCommandResponse {
     },
     verify_address {
         verified: bool,
+        short_address_hash: Option<String>,
     },
     version {
         string: String,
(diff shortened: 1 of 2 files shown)


full-service/src/json_rpc/v2/api/response.rs
```diff
@@ -102,6 +102,7 @@ pub enum JsonCommandResponse {
     },
     get_address_details {
         details: PublicAddress,
+        short_address_hash: String,
```
nick-mobilecoin, line 105: I'm a bit on the fence with this suggestion

My initial reaction to this field name is that `address` is redundant.

Seeing the type `ShortAddressHash` later in the code, I can understand how this field name may have been chosen for consistency.  I think the use of `short_address_hash` is falling more into the camp of using the type name in the field/variable, than naming the field based on what it means to the containing struct.
```suggestion
        short_hash: String,
```
consider `address.short_address_hash` versus `address.short_hash`

if this change is made the field in verify_address should be renamed as well.

full-service/src/json_rpc/v2/api/wallet.rs
```diff
@@ -1443,8 +1444,17 @@ where
                 .map_err(format_error)?;
             JsonCommandResponse::validate_sender_memo { validated: result }
         }
-        JsonCommandRequest::verify_address { address } => JsonCommandResponse::verify_address {
-            verified: service.verify_address(&address).is_ok(),
+        JsonCommandRequest::verify_address { address } => match service.verify_address(&address) {
```
nick-mobilecoin, line 1447: :question: where does this, https://mobilecoin.gitbook.io/full-service-api/api-endpoints/v2/account/address/verify_address, bet updated at?

nick-mobilecoin (approved):
looks good, just a bit of bikeshedding on field names

full-service/src/json_rpc/v2/api/response.rs
holtzman: thanks. decided to go with `address_hash` instead as `short` too was redundant (it referred to the length of the hash, which is self-evident) and just `hash` seems too generic.

full-service/src/json_rpc/v2/api/wallet.rs
holtzman: We maintain the docs using gitbook itself.  [Here is a preview of the edits ](https://mobilecoin.gitbook.io/full-service-api/~/changes/AqCyVo43Lcl820LFbEoI/api-endpoints/v2/account/address/verify_address) that correspond to this PR

PR #957, opened 2024-02-05
update ci and release workflows to use mco-dev-large-x64 runners

holtzman:
### Motivation

ci workflows not being picked up by any runner.

### In this PR

changed ci and release workflows from using runs-on: [self-hosted, Linux, large] to runs-on: mco-dev-large-x64

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ concurrency:
 
 jobs:
   lint:
-    runs-on: [self-hosted, Linux, large]
+    runs-on: mco-dev-large-x64
     container:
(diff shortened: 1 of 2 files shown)


jgreat (approved):

nick-mobilecoin (approved):
looks good

Some other workflows in the `.github` directory are still using `[self-hosted ...]`, but my understanding is those don't work at all right now so not pressing to update them in this PR.

holtzman:
> looks good
> 
> Some other workflows in the `.github` directory are still using `[self-hosted ...]`, but my understanding is those don't work at all right now so not pressing to update them in this PR.

That was my thinking too.

PR #959, opened 2024-02-10
Chore/uprev ci tools

holtzman:
### In this PR

* update version of mobilecoin/rust-sgx-base to use to v0.0.32
* update various runners to current preferred ones
* fix to `Dockerfile.full-service-with-build` to copy a build's executable artifacts from the `/app/target` specialized cache within the same RUN layer as does the `cargo build`.  Previously, the numerous `RUN cp /app/target/...` commands failed because `/app/target/` has been mounted from a cache in the `cargo build` layer and was not present in these layers.

### DIFF
--- a/.internal-ci/docker/Dockerfile.full-service-with-build
+++ b/.internal-ci/docker/Dockerfile.full-service-with-build
@@ -28,7 +28,7 @@
 # In order to create a consistent and verifiable the build environment, only add
 # or update in the mobilecoin/rust-sgx-base image and refer to the image by its hash.
 
-FROM mobilecoin/rust-sgx-base:v0.0.28 as builder
+FROM mobilecoin/rust-sgx-base:v0.0.32 as builder
 
 ARG NAMESPACE=test
 ARG SIGNED_ENCLAVE_BASE=enclave-distribution.${NAMESPACE}.mobilecoin.com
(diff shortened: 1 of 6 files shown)


.internal-ci/docker/Dockerfile.full-service-with-build
```diff
@@ -59,15 +59,15 @@ RUN  --mount=type=cache,target=/opt/cargo/git \
...
-
-RUN cp /app/target/release/full-service /usr/local/bin/
-RUN cp /app/target/release/signer /usr/local/bin/
-RUN cp /app/target/release/signer-service /usr/local/bin/
-RUN cp /app/target/release/validator-service /usr/local/bin/
-RUN cp /app/target/release/wallet-service-mirror-private /usr/local/bin/
-RUN cp /app/target/release/wallet-service-mirror-public /usr/local/bin/
-RUN cp /app/target/release/generate-rsa-keypair /usr/local/bin/
+  cargo build --locked --release ${BUILD_OPTS} \
+  && cp /app/target/release/full-service \
```
nick-mobilecoin, line 63: I didn't know you could do multiple sources with cp
> Copy SOURCE to DEST, or multiple SOURCE(s) to DIRECTORY.

nick-mobilecoin (approved):

.internal-ci/docker/Dockerfile.full-service-with-build
holtzman: Thanks for the prompt, after-hours review!

and yeah, cp letting you have an arbitrary number of sources is particularly useful when combined with the shell's brace expansion pattern.  I didn't use it here for readability, but you can do something like:

cp /long/path/i/dont/want/type/over/and/over/{a,b,c}.txt /dest/path/thats/also/long

to copy a.txt, b.txt, and c.txt to a new place.

works with mv too.

PR #967, opened 2024-03-21
Bugfix/value pmob cast

holtzman:
### Motivation

The wallet database doesn't store u64s and full-service casts u64s to i64 for storage.  In a few places, those values being retrieved from the database were not being cast back to u64 before being used, notably in transaction_log processing and receipt processing.

### In this PR
* fixed two missing u64 casts in v1 transaction_log retrieval
* fixed a missing u64 cast when forming an response indicating that that a receipt's value doesn't match the txo.


nick-mobilecoin (approved):
looks good, there are a couple test cases left using i64, but they're also comparing to i64.
Very anecdotal, but this is a good argument for normalizing data before passing it around.

PR #971, opened 2024-04-04
Introduce Spend Only from Subaddress

sugargoat:
### Motivation

Some full-service users would prefer a mode that enables spending only from a given subaddress. While subaddresses were conceived as "turnstiles" - only tracking the amounts that come into a wallet through a given address, it is reasonable to want to instrument a wallet layer with this restriction. This PR introduces an API for spending only TXOs that were received at a given subaddress. 

### In this PR
- [x] Adds `spend_only_from_subaddress` field to `build_and_submit_transaction` (and affiliated API endpoints)
- [x] Updates documentation to add the new endpoint (see change request here: https://app.gitbook.com/o/-MW0kJdX5CaZTiJrSPwE/s/-MXOsD-rEx51AyOMlu_p-2910905616/~/changes/161/)
- [x] Ensures that change returns to the subaddress that was being spent from
- [x] Ensures the memo is constructed correctly 
- [x] Question: does `build_unsigned` and `build_unsigned_burn` need this as well? 
    - Note: I included, since they both ultimately call the same underlying function, but I did not expose the `spend_only_subaddress` in the API for these methods
- [x] Confirms T3 works correctly 
    - T3 syncs from TXOs, uses the AuthenticatedSenderMemo directly, and derives the Destination RecipientAddressHash from the recipient logged with the TXO data (https://github.com/mobilecoinofficial/full-service/blob/main/full-service/src/service/t3_sync.rs#L125). To select which TXOs to upload to T3, it [filters only on TXOs with a sender memo](https://github.com/mobilecoinofficial/full-service/blob/main/full-service/src/db/txo.rs#L1001) (so now that we are writing the change memo correctly, T3 syncing will continue to work as designed)

### Test Plan

- [x] Unittests added for new functionality
- [x] Integration test added and run (e2e) for new functionality
- [x] Tested against deployed network (Prod and/or TestNet)

### Future Work
* Add a specific mode that restricts usage to only the above features when it is enabled (need to think through what other functionality might be needed in this mode), and a concept of "subaccounts" to flesh this out
* Add a test for T3 uploading to confirm it filters the txos correctly (something that would have caught the change txo having a sender memo rather than a destination memo)

### DIFF
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -3035,23 +3035,20 @@ mod tests {
             .unwrap();
 
         // Create two TXOs for Alice's assigned address
-        let (_txo_hex, _txo, _key_image) = create_test_received_txo(
-            &account_key,
-            alice_subaddress_index as u64,
-            Amount::new(33 * MOB, Mob::ID),
-            145,
-            &mut rng,
-            &wallet_db,
-        );
-
-        let (_txo_hex, _txo, _key_image) = create_test_received_txo(
-            &account_key,
-            alice_subaddress_index as u64,
-            Amount::new(44 * MOB, Mob::ID),
-            146,
-            &mut rng,
-            &wallet_db,
-        );
+        let mut block_index = 145;
+        let alice_balance = [33, 44].iter().fold(0, |acc, value| {
+            let (_txo_hex, _txo, _key_image) = create_test_received_txo(
+                &account_key,
+                alice_subaddress_index as u64,
+                Amount::new(value * MOB, Mob::ID),
+                block_index,
+                &mut rng,
+                &wallet_db,
+            );
+            block_index += 1;
+            acc + value
+        });
+        assert_eq!(77, alice_balance);
 
         let spendable_txos = Txo::list_spendable(
             Some(&account_id_hex.to_string()),
@@ -3063,29 +3060,25 @@ mod tests {
         )
         .unwrap();
 
-        // Max spendable SHOULD be the sum of all TXOs at Alice's subaddress
         assert_eq!(
             spendable_txos.max_spendable_in_wallet,
-            ((77 * MOB) - Mob::MINIMUM_FEE) as u128
+            ((alice_balance * MOB) - Mob::MINIMUM_FEE) as u128
         );
 
-        // Add some coins at Bob's subaddress
-        let (_txo_hex, _txo, _key_image) = create_test_received_txo(
-            &account_key,
-            bob_subaddress_index as u64,
-            Amount::new(55 * MOB, Mob::ID),
-            147,
-            &mut rng,
-            &wallet_db,
-        );
-        let (_txo_hex, _txo, _key_image) = create_test_received_txo(
-            &account_key,
-            bob_subaddress_index as u64,
-            Amount::new(66 * MOB, Mob::ID),
-            148,
-            &mut rng,
-            &wallet_db,
-        );
+        // Add two coins at Bob's subaddress
+        let bob_balance = [55, 66].iter().fold(0, |acc, value| {
+            let (_txo_hex, _txo, _key_image) = create_test_received_txo(
+                &account_key,
+                bob_subaddress_index as u64,
+                Amount::new(value * MOB, Mob::ID),
+                block_index,
+                &mut rng,
+                &wallet_db,
+            );
+            block_index += 1;
+            acc + value
+        });
+        assert_eq!(121, bob_balance);
 
         // Max spendable SHOULD be the same at Alice's subaddress
         let spendable_txos = Txo::list_spendable(
@@ -3099,7 +3092,7 @@ mod tests {
         .unwrap();
         assert_eq!(
             spendable_txos.max_spendable_in_wallet,
-            ((77 * MOB) - Mob::MINIMUM_FEE) as u128
+            ((alice_balance * MOB) - Mob::MINIMUM_FEE) as u128
         );
 
         // Max spendable for Bob should be the sum of all TXOs at Bob's subaddress
@@ -3114,7 +3107,7 @@ mod tests {
         .unwrap();
         assert_eq!(
             spendable_txos.max_spendable_in_wallet,
-            ((121 * MOB) - Mob::MINIMUM_FEE) as u128
+            ((bob_balance * MOB) - Mob::MINIMUM_FEE) as u128
         );
 
         // Max spendable for the account should be the sum of all Alice's and Bob's
@@ -3129,7 +3122,7 @@ mod tests {
         .unwrap();
         assert_eq!(
             spendable_txos.max_spendable_in_wallet,
-            ((198 * MOB) - Mob::MINIMUM_FEE) as u128
+            (((alice_balance + bob_balance) * MOB) - Mob::MINIMUM_FEE) as u128
         );
     }
 
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -145,6 +145,7 @@ impl Txo {
     }
 }
 
+#[derive(Debug)]
 pub struct TxoInfo {
     pub txo: Txo,
     pub memo: TxoMemo,
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -645,9 +645,9 @@ pub trait TxoModel {
         account_id_hex: &str,
         target_value: u128,
         max_spendable_value: Option<u64>,
+        assigned_subaddress_b58: Option<&str>,
         token_id: u64,
         default_token_fee: u64,
-        only_from_subaddress: Option<&str>,
         conn: Conn,
     ) -> Result<Vec<Txo>, WalletDbError>;
 
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -631,6 +631,7 @@ pub trait TxoModel {
     ///| Name                  | Purpose                                                    | Notes                               |
     ///|-----------------------|------------------------------------------------------------|-------------------------------------|
     ///| `account_id_hex`      | The account id where the Txos from                         | Account must exist in the database. |
+    ///| `subaccount_address`  | The subaddress at which the list of Txos from              | (optional)                          |
     ///| `target_value`        | The value used to filter spendable Txos on its value       |                                     |
     ///| `max_spendable_value` | The upper limit for the spendable TxOut value to filter on |                                     |
     ///| `token_id`            | The id of a supported type of token to filter on           |                                     |
@@ -641,6 +642,7 @@ pub trait TxoModel {
     /// * Vector of TxoOut
     fn select_spendable_txos_for_value(
         account_id_hex: &str,
+        subaccount_address: Option<&str>,
         target_value: u128,
         max_spendable_value: Option<u64>,
         token_id: u64,
@@ -1858,19 +1860,23 @@ impl TxoModel for Txo {
 
     fn select_spendable_txos_for_value(
         account_id_hex: &str,
+        subaccount_address: Option<&str>,
         target_value: u128,
         max_spendable_value: Option<u64>,
         token_id: u64,
         default_token_fee: u64,
         conn: Conn,
     ) -> Result<Vec<Txo>, WalletDbError> {
+
+        // TODO: Logic - switch on option account_id vs subaccount_address
+
         let SpendableTxosResult {
             mut spendable_txos,
             max_spendable_in_wallet,
         } = Txo::list_spendable(
             Some(account_id_hex),
             max_spendable_value,
-            None,
+            None, // TODO: Note, if !None, this will only get spendable for an address
             token_id,
             default_token_fee,
             conn,
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -634,9 +634,9 @@ pub trait TxoModel {
     ///| `account_id_hex`      | The account id where the Txos from                         | Account must exist in the database. |
     ///| `target_value`        | The value used to filter spendable Txos on its value       |                                     |
     ///| `max_spendable_value` | The upper limit for the spendable TxOut value to filter on |                                     |
+    ///| `assigned_subaddress_b58`  | The subaddress where the spendable Txos can be sourced from |                                      |
     ///| `token_id`            | The id of a supported type of token to filter on           |                                     |
     ///| `default_token_fee`   | The default transaction fee in Mob network                 |                                     |
-    ///| `assigned_subaddress_b58`  | The subaddress where the spendable Txos can be sourced from                    |                                      |
     ///| `conn`                | An reference to the pool connection of wallet database     |                                     |
     ///
     /// # Returns:
@@ -1874,7 +1874,7 @@ impl TxoModel for Txo {
         } = Txo::list_spendable(
             Some(account_id_hex),
             max_spendable_value,
-            only_from_subaddress_b58,
+            assigned_subaddress_b58,
             token_id,
             default_token_fee,
             conn,
@@ -3142,9 +3142,9 @@ mod tests {
             &account_id_hex.to_string(),
             300 * MOB as u128,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         )
         .unwrap();
@@ -3156,9 +3156,9 @@ mod tests {
             &account_id_hex.to_string(),
             (300 * MOB + Mob::MINIMUM_FEE) as u128,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         )
         .unwrap();
@@ -3173,9 +3173,9 @@ mod tests {
             &account_id_hex.to_string(),
             (300 * MOB + Mob::MINIMUM_FEE) as u128,
             Some(200 * MOB),
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         );
 
@@ -3191,9 +3191,9 @@ mod tests {
             &account_id_hex.to_string(),
             16800 * MOB as u128,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         )
         .unwrap();
@@ -3312,9 +3312,9 @@ mod tests {
                 &account_id_hex.to_string(),
                 42 * MOB as u128,
                 None,
+                Some(subaddress),
                 0,
                 Mob::MINIMUM_FEE,
-                Some(subaddress),
                 conn,
             )
             .unwrap();
@@ -3328,9 +3328,9 @@ mod tests {
             &account_id_hex.to_string(),
             (100 * MOB + Mob::MINIMUM_FEE) as u128,
             None,
+            Some(&alice_public_address_b58),
             0,
             Mob::MINIMUM_FEE,
-            Some(&alice_public_address_b58),
             conn,
         );
 
@@ -3380,9 +3380,9 @@ mod tests {
             &account_id_hex.to_string(),
             16800 * MOB as u128,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         )
         .unwrap();
@@ -3391,9 +3391,9 @@ mod tests {
             &account_id_hex.to_string(),
             16800 * MOB as u128,
             Some(100 * MOB),
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         );
 
@@ -3445,9 +3445,9 @@ mod tests {
             &account_id_hex.to_string(),
             1800 * MOB as u128,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         );
         match res {
@@ -4241,9 +4241,9 @@ mod tests {
             &account_id.to_string(),
             target_value,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         )
         .unwrap();
@@ -4260,9 +4260,9 @@ mod tests {
             &account_id.to_string(),
             201 * MOB as u128,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         );
 
@@ -4279,9 +4279,9 @@ mod tests {
             &account_id.to_string(),
             3,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         )
         .unwrap();
@@ -4296,9 +4296,9 @@ mod tests {
             &account_id.to_string(),
             500 * MOB as u128,
             None,
+            None,
             0,
             Mob::MINIMUM_FEE,
-            None,
             &mut wallet_db.get_pooled_conn().unwrap(),
         );
         assert!(result.is_err());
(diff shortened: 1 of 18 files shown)


full-service/src/db/txo.rs
```diff
@@ -3021,6 +3026,7 @@ mod tests {
             None,
             0,
             Mob::MINIMUM_FEE,
+            None,
```
sugargoat, line 3029: These tests do not cover the `only_from_subaddress_b58` so they all set the new field to None to test functionality from listing/spending from the main account (and sourcing from any txo no matter which subaddress it came through)

full-service/src/json_rpc/v2/api/request.rs
```diff
@@ -57,6 +57,7 @@ pub enum JsonCommandRequest {
         block_version: Option<String>,
         sender_memo_credential_subaddress_index: Option<String>,
         payment_request_id: Option<String>,
+        spend_only_from_subaddress: Option<String>,
```
sugargoat, line 60: Updated all the APIs for building transactions to include the `spend_only_from_subaddress` filter

full-service/src/service/transaction.rs
```diff
@@ -341,6 +341,7 @@ pub trait TransactionService {
     ///| `max_spendable_value`   | The maximum amount for an input TXO selected for this transaction |                                                                                                   |
     ///| `memo`                  | Memo for the transaction                                          |                                                                                                   |
     ///| `block_version`         | The block version to build this transaction for.                  | Defaults to the network block version                                                             |
+    ///| `spend_only_from_subaddress` | The subaddress index to spend from.                          | (optional)                                                                                        |
```
sugargoat, line 344: This is the main motivation for this PR

full-service/src/service/transaction.rs
```diff
@@ -512,6 +519,11 @@ where
             if let Some(inputs) = input_txo_ids {
                 builder.set_txos(conn, inputs)?;
             } else {
+                if let Some(subaddress) = spend_only_from_subaddress {
+                    // Ensure the builder will filter to txos only from the specified subaddress
+                    builder.set_spend_only_from_subaddress(subaddress)?;
```
sugargoat, line 524: We set this on a property of the builder so that when the builder is filtering txos, it knows to only select from txos received at the given subaddress

full-service/src/service/transaction.rs
```diff
@@ -1674,4 +1703,180 @@ mod tests {
             authenticated_sender_memo.payment_request_id()
         );
     }
+
+    // Test sending a transaction from only a specified subaddress, and that change
+    // arrives back to that subaddress.
+    #[async_test_with_logger]
+    async fn test_send_transaction_from_only_subaddress(logger: Logger) {
```
sugargoat, line 1710: This is the main unittest asserting the functionality we want

full-service/src/db/txo.rs
```diff
@@ -145,6 +145,7 @@ impl Txo {
     }
 }
 
+#[derive(Debug)]
```
sugargoat, line 148: Wanted to be able to print this structure in tests

full-service/src/db/txo.rs
```diff
@@ -635,6 +636,7 @@ pub trait TxoModel {
     ///| `max_spendable_value` | The upper limit for the spendable TxOut value to filter on |                                     |
     ///| `token_id`            | The id of a supported type of token to filter on           |                                     |
     ///| `default_token_fee`   | The default transaction fee in Mob network                 |                                     |
+    ///| `only_from_subaddress`| The subaddress from which spendable txos can be sourced    | (optional)                          |
```
sugargoat, line 639: The main motivation of this PR: the ability to select spendable txos _only_ from a give subaddress

full-service/src/json_rpc/v1/api/wallet.rs
```diff
@@ -181,6 +181,7 @@ where
                         subaddress_index: None,
                     },
                     None,
+                    None, // Note: Not including spend_only_from_subaddress in V1 API
```
sugargoat, line 184: Needed to update the called method from the v1 API, but we are not adding this functionality to V1.

full-service/src/service/models/tx_proposal.rs
```diff
@@ -547,6 +547,7 @@ mod tests {
                     subaddress_index: Some(alice_address_from_bob.subaddress_index as u64),
                 },
                 None,
+                None,
```
sugargoat, line 550: Not testing spend_only_from_subaddress at the tx_proposal level, as this is simply testing the construction of the tx_proposal is sound, not whether it selected the correct TXOs. That logic is tested in txo.rs

full-service/src/service/receipt.rs
```diff
@@ -433,6 +433,7 @@ mod tests {
                     subaddress_index: None,
                 },
                 None,
+                None,
```
sugargoat, line 436: Not testing whether receipts were constructed with txos selected only from a specified subaddress, as this is not functionality specific to receipts.

full-service/src/service/transaction_builder.rs
```diff
@@ -430,20 +446,43 @@ impl<FPR: FogPubkeyResolver + 'static> WalletTransactionBuilder<FPR> {
...
             let change_amount = Amount::new(change_value as u64, token_id);
-            let tx_out_context = transaction_builder.add_change_output(
-                change_amount,
-                &reserved_subaddresses,
-                &mut rng,
-            )?;
+            if let Some(spend_from_subaddress) = self.spend_only_from_subaddress.as_ref() {
+                // Send the change back to the subaddress that is spending the inputs.
+                // In the future, we may want to allow this to be a bit more configurable
+                let change_address = b58_decode_public_address(&spend_from_subaddress)?;
```
sugargoat, line 452: ** ⚠️  IMPORTANT IMPLEMENTATION DETAIL ⚠️ **

I made the decision to send the change back to the subaddress from which the transaction is being spent. 🚨 THIS BREAKS RECOVERABLE TRANSACTION HISTORY.

**Why do I think this is ok?**

* By choosing to spend only from a given subaddress, the full-service subaddress model is being broken. The subaddresses are meant to be turnstiles, so that you can track funds coming in, but you can spend from any txo. This is important to several design decisions made for this wallet, and is meant to lend to efficiency and simplicity of logic.
* This new model with spending only from subaddresses implies that a subaddress is more like a subaccount. The challenge here is that an ACCOUNT only has ONE change address, the default change address.

The options for where change could go:

1. Change goes to the default change subaddress. ❌ Then the balance for the subaddress would instantly be wrong, because typically there is only one TXO which is then spent from, and that change would no longer be associated with the spending subaddress.
2. Change goes back to the subaddress from which it was spent. 🟢 This is ok, BUT, there is no way currently to write a Destination memo for any output not going to the default change subaddress.
3. Create a new scheme where each subaddress has an associated change subaddress, and we increment by two subaddresses whenever a subaddress is created. ❌ This is more complicated than necessary at this point. If we find that people really want recoverable transaction history AND subaddresses that are actually behaving like subaccounts, we could revisit this. As it stands, the transaction history is preserved with transaction logs.

full-service/src/service/txo.rs
```diff
@@ -339,6 +339,8 @@ where
                 subaddress_index: None,
             },
             None,
+            None, /* FIXME: I don't believe we need to provide spend_only_from_subaddress for
```
sugargoat, line 342: Splitting TXOs is a very specific endpoint, and does not need to have `spend_only_from_subaddress` specified

full-service/src/service/transaction_builder.rs
holtzman: What technically precludes including a destination memo on the change txo when it goes back to the subaddress from which the spend is happening?

Does this trait implementation not allow overriding the change subaddress in a way that would cause the destination memo to be properly written? https://github.com/mobilecoinfoundation/mobilecoin/blob/39cdbb21b89e173ee7b3f1666a7f65b8ff7e58e9/transaction/builder/src/reserved_subaddresses.rs#L57

There is, and should be noted, a reason to distrust destination memos sent by anybody but the originator of the change, so a wallet implementer would want to evaluate the threat model and ramifications of possibly forged destination memos before deciding to make use of them.

full-service/src/service/transaction.rs
```diff
@@ -1674,4 +1703,345 @@ mod tests {
...
+        let change_shared_secret = get_tx_out_shared_secret(
+            exchange_account_key.view_private_key(),
+            &RistrettoPublic::try_from(&alice_to_bob_change_txo.public_key).unwrap(),
+        );
+
+        // FIXME: This is ideally how this works once we update the
+        // `transaction_builder.add_change_output` to accept a non-change
+        // subaddress. For now, we're just decrypting the memo as an authenticated
+        // sender memo.
+        let change_memo = alice_to_bob_change_txo.decrypt_memo(&change_shared_secret);
```
sugargoat, line 1917: Note: ⚠️ This is where the change memo logic is being tested, and flagged how it should be tested if DestinationMemo is properly constructed

full-service/src/service/transaction_builder.rs
sugargoat: Wow 🌮 That is exactly what I needed! So much emoji angst above 😅 when the API was already available to do exactly what we needed! Code is updated, and I'm going to clean up the builder call a little bit, but please take a look!

Thank you @holtzman!

full-service/src/service/transaction.rs
sugargoat: Destination memo is now being properly constructed, so resolving this comment

full-service/src/service/transaction.rs
```diff
@@ -1674,4 +1711,338 @@ mod tests {
...
+                WalletTransactionBuilderError::WalletDb(
+                    WalletDbError::InsufficientFundsUnderMaxSpendable(_),
+                ),
+            )) => {}
+            Ok(_) => panic!("Should error with InsufficientFundsUnderMaxSpendable"),
+            Err(e) => panic!(
+                "Should error with InsufficientFundsUnderMaxSpendable but got {:?}",
+                e
+            ),
+        }
```
nick-mobilecoin, line 2030: How is this different than the failed sign and submit above?
should this go away?
if really desired to have multiple values perhaps an iterator of the values
```suggestion
```

full-service/src/service/transaction_builder.rs
```diff
@@ -430,20 +453,52 @@ impl<FPR: FogPubkeyResolver + 'static> WalletTransactionBuilder<FPR> {
...
+
+                let change_txo = OutputTxo {
+                    tx_out: tx_out_context.tx_out,
+                    recipient_public_address: reserved_subaddresses.change_subaddress.clone(),
+                    confirmation_number: tx_out_context.confirmation,
+                    amount: change_amount,
+                    shared_secret: Some(tx_out_context.shared_secret),
+                };
+                change_txos.push(change_txo);
+            }
```
nick-mobilecoin, line 501: consider doing only the differences in different blocks
```suggestion
            let (change_address, change_destination) = match self.spend_only_from_subaddress {
                Some(ref subaddress) => {
                    let change_address = b58_decode_public_address(subaddress)?;
                    let reserved_subaddresses = ReservedSubaddresses::from_subaddress_index(
                        &account.account_key()?,
                        self.spend_only_from_subaddress_index,
                        None,
                    );
                    (change_address, reserved_subaddresses)
                }
                None => (
                    reserved_subaddresses.change_subaddress.clone(),
                    reserved_subaddresses.clone(),
                ),
            };
            let tx_out_context = transaction_builder.add_change_output(
                change_amount,
                &reserved_subaddresses,
                &mut rng,
            )?;

            let change_txo = OutputTxo {
                tx_out: tx_out_context.tx_out,
                recipient_public_address: change_address,
                confirmation_number: tx_out_context.confirmation,
                amount: change_amount,
                shared_secret: Some(tx_out_context.shared_secret),
            };
            change_txos.push(change_txo);
```

TBH the `let (..) match` logic should probably be moved outside of the for loop, this would alleviate one of the clones as well as thin up the logic that's critical to the for loop

I'm also under the impression the FIXME no longer applies with the latest change of using `ReservedSubaddresses::from_subaddress_index()`

nick-mobilecoin (commented):
I think the main hiccups are:
1. Ensuring the argument naming with respect  to `spend_only_from_subbaddress` and the pre-existing `assigned_subaddress`
2. ensuring we want to be using a b58 subaddress, when other apis for subaddresses seem to use the index

There are some comments with ":shrug: take it or leave it", I can't stress enough they're suggestions and I understand if one wants to resolve without making those changes.

full-service/src/db/txo.rs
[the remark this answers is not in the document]
sugargoat: The ambiguity I'm trying to avoid is that an `assigned_subaddress_b58` is some subaddress that was created (assigned) for an expected counterparty (e.g. an exchange assigning for one of their customers, where they expect to receive funds). Because an address could be used in many ways (and typically is used only for receiving), I wanted to make it clear that this address was being used _only for spending from_ during this operation.

Let me think about this one though, because it is a good point about consistency with naming.

full-service/src/db/txo.rs
```diff
@@ -645,6 +647,7 @@ pub trait TxoModel {
         max_spendable_value: Option<u64>,
```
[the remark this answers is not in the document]
sugargoat: Good catch, though the way the account/subaddress works is not quite an either/or - the account_id_hex is required, because the wallet supports multiple accounts. But then the subaddress is optional. So for example, as an exchange, I may have Account A (my main account where I custody our exchange funds) and Account B (the account where I custody customer's funds), and then in Account B, I assign subaddresses for my customers to send to, and for the purposes of this PR, at some point I want to spend from the funds received at one of those subaddresses.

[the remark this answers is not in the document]
sugargoat: I think you could potentially make it an either/or using lookups from `assigned_subaddress_b58` -> `account_id`, as that mapping is present in the `AssignedSubaddress` table schema. Though we typically re-derive from `account_keys` -> `subaddress_index` -> `public_address` -> `public_address_b58`.

full-service/src/db/txo.rs
```diff
@@ -2980,6 +2984,155 @@ mod tests {
...
+
+        // Assign an address for Alice
+        let (alice_public_address_b58, alice_subaddress_index) =
+            AssignedSubaddress::create_next_for_account(
+                &account_id_hex.to_string(),
+                "Alice's Subaddress",
+                &ledger_db,
+                &mut wallet_db.get_pooled_conn().unwrap(),
+            )
+            .unwrap();
```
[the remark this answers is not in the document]
sugargoat: Nice, I like that quite a lot - and I just learned about `collect_tuple` so I can assign all in one go 🙌

full-service/src/db/txo.rs
```diff
@@ -3094,6 +3251,210 @@ mod tests {
         );
     }
 
+    // The narrative for this test is that an exchange creates three assigned
```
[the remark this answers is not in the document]
sugargoat: I aspire to more parameterized tests, as I've seen you beautifully model. The tests are a little bloated right now with copy/pasta, and a refactor could make it easier to test new functionality (and ensure old functionality is being preserved)

full-service/src/db/txo.rs
```diff
@@ -3094,6 +3251,210 @@ mod tests {
...
+            )
+            .unwrap();
+        let (carol_public_address_b58, carol_subaddress_index) =
+            AssignedSubaddress::create_next_for_account(
+                &account_id_hex.to_string(),
+                "Carol's Subaddress",
+                &ledger_db,
+                &mut wallet_db.get_pooled_conn().unwrap(),
+            )
+            .unwrap();
```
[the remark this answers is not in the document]
sugargoat: Love it

full-service/src/db/txo.rs
```diff
@@ -3094,6 +3251,210 @@ mod tests {
...
+            Some(&carol_public_address_b58),
+            0,
+            Mob::MINIMUM_FEE,
+            conn,
+        )
+        .unwrap();
+        assert_eq!(
+            spendable_txos.max_spendable_in_wallet,
+            ((300 * MOB) - Mob::MINIMUM_FEE) as u128
+        );
```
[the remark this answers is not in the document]
sugargoat: Great point - removing

full-service/src/db/txo.rs
```diff
@@ -3094,6 +3251,210 @@ mod tests {
...
+            42 * MOB as u128,
+            None,
+            0,
+            Mob::MINIMUM_FEE,
+            Some(&carol_public_address_b58),
+            &mut wallet_db.get_pooled_conn().unwrap(),
+        )
+        .unwrap();
+        let result_set = HashSet::from_iter(txos_for_value.iter().map(|t| t.value as u64));
+        assert_eq!(result_set, HashSet::from_iter([300 * MOB]));
```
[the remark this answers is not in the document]
sugargoat: Great!

full-service/src/db/txo.rs
```diff
@@ -3094,6 +3251,210 @@ mod tests {
...
+
+        // Exceed the amount that Alice has, even though the full wallet or Bob or Carol
+        // could cover it
+        let res = Txo::select_spendable_txos_for_value(
+            &account_id_hex.to_string(),
+            (150 * MOB + Mob::MINIMUM_FEE) as u128,
```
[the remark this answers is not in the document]
sugargoat: Good ZOMBIES sug 🧟

full-service/src/db/txo.rs
```diff
@@ -3094,6 +3251,210 @@ mod tests {
...
+            Mob::MINIMUM_FEE,
+            Some(&alice_public_address_b58),
+            &mut wallet_db.get_pooled_conn().unwrap(),
+        );
+
+        match res {
+            Err(WalletDbError::InsufficientFundsUnderMaxSpendable(_)) => {}
+            Ok(_) => panic!("Should error with InsufficientFundsUnderMaxSpendable"),
+            Err(_) => panic!("Should error with InsufficientFundsUnderMaxSpendable"),
+        }
```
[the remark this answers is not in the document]
sugargoat: Nice

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_spend_from_subaddress.rs
```diff
@@ -0,0 +1,412 @@
...
+        let unspent = balance_mob["unspent"].as_str().unwrap();
+        let pending = balance_mob["pending"].as_str().unwrap();
+        let spent = balance_mob["spent"].as_str().unwrap();
+        let secreted = balance_mob["secreted"].as_str().unwrap();
+        let orphaned = balance_mob["orphaned"].as_str().unwrap();
+        assert_eq!(unspent, (142 * MOB).to_string(),); // FIXME: shouldn't this be 142?
+        assert_eq!(pending, "0");
+        assert_eq!(spent, (0).to_string(),); // FIXME: Why was it spent from Alice's balance?
+                                             //assert_eq!(spent, (100 * MOB).to_string(),); // FIXME: Why was it spent from
+                                             // Alice's balance?
```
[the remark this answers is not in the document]
sugargoat: Thank you! Fixed fixmes!

full-service/src/json_rpc/v2/api/request.rs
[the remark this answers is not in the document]
sugargoat: 1. Love - much clearer. Still considering the name throughout wrt `assigned_subaddress_b58` as well
2. The assumption is that exchanges will be storing the `assigned_address_b58` string, as that's what they'll be giving to their users, so we use that as an identifier. Maybe a more robust API would take either/or?
3. This is actually a very interesting question (and one we've considered for T3 as a feature, not a bug). For example, it is possible to set the sender memo as any subaddress you control, and you may want to do this for consistency across your customer base, and in T3 records. (For example, it benefits the institution to report the same subaddress for all of their records to make it easier for forensics to aggregate their activity)

full-service/src/db/txo.rs
```diff
@@ -2980,6 +2984,155 @@ mod tests {
...
+        );
+
+        let (_txo_hex, _txo, _key_image) = create_test_received_txo(
+            &account_key,
+            alice_subaddress_index as u64,
+            Amount::new(44 * MOB, Mob::ID),
+            146,
+            &mut rng,
+            &wallet_db,
+        );
```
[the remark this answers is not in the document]
sugargoat: NGL it took me a sec to grok the acc+value, but I think the clarity otherwise is worth it

full-service/src/db/txo.rs
```diff
@@ -2980,6 +2984,155 @@ mod tests {
...
+            None,
+            0,
+            Mob::MINIMUM_FEE,
+            conn,
+        )
+        .unwrap();
+        assert_eq!(
+            spendable_txos.max_spendable_in_wallet,
+            ((198 * MOB) - Mob::MINIMUM_FEE) as u128
+        );
```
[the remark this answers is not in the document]
sugargoat: Looks great - and I'll add a clarifying comment re: the overall account balance

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_spend_from_subaddress.rs
```diff
@@ -0,0 +1,412 @@
...
+        let unspent = balance_mob["unspent"].as_str().unwrap();
+        let pending = balance_mob["pending"].as_str().unwrap();
+        let spent = balance_mob["spent"].as_str().unwrap();
+        let secreted = balance_mob["secreted"].as_str().unwrap();
+        let orphaned = balance_mob["orphaned"].as_str().unwrap();
+        assert_eq!(unspent, (300 * MOB).to_string(),);
+        assert_eq!(pending, "0");
+        assert_eq!(spent, "0");
+        assert_eq!(secreted, "0");
+        assert_eq!(orphaned, "0");
```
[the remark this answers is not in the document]
sugargoat: Great ty

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_spend_from_subaddress.rs
```diff
@@ -0,0 +1,412 @@
...
+        // One change (and that change should be going back to Bob)
+        assert_eq!(tx_proposal.change_txos.len(), 1);
+        assert_eq!(
+            tx_proposal.change_txos[0].recipient_public_address_b58,
+            bob_b58_public_address
+        );
+        // Tombstone block = ledger height (12 to start + 3 new blocks + 10 default
+        // tombstone)
+        assert_eq!(tx_proposal.tombstone_block_index, "25");
+
```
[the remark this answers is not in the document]
sugargoat: Very true - copy/pasta - will clean up

full-service/src/db/txo.rs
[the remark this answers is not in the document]
nick-mobilecoin: oh [collect_tuple](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.collect_tuple) is interesting, I was hoping there was a more constant way to do it since `["alice", "bob"]` is a compile time size, but collect_tuple should work fairly well in tests with an "unwrap()"

full-service/src/service/transaction.rs
```diff
@@ -1674,4 +1711,338 @@ mod tests {
...
+        log::info!(
+            logger,
+            "Built and submitted transaction from Alice's Subaddress to Bob's Subaddress"
+        );
+
+        // NOTE: Submitting to the test ledger via propose_tx doesn't actually add the
+        // block to the ledger, because no consensus is occurring, so this is the
+        // workaround.
+        {
+            log::info!(logger, "Adding block from transaction log");
```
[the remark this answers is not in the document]
sugargoat: 💯 thank you

full-service/src/service/transaction.rs
sugargoat: Ah, I checked the narrative at the top - the goals for the two tests are:

1. Confirm that more than Alice OR Bob has, but enough that THE WALLET has (Alice + Bob + Carol)
2. More than Alice has, but enough that BOB has

Will tidy up!

sugargoat: Trying to iteratorize this, and not sure how to handle the await, if you have a sug, @nick-mobilecoin !

sugargoat: Ok, I think I got it with `join_all` - commit coming shortly

sugargoat: Actually I don't got it - pushed my changes for you to review, if you have a sug on how to fix it up!

holtzman (approved):
Looks great! Thanks

PR #989, opened 2024-05-21
Add Require Spend Subaddress

sugargoat:
### Motivation

We would like to provide a guard rail for the use of subaddress_to_spend_from, so that an account can enforce that transactions using the Txos within the account adhere to the scheme of spending from any Txo regardless of which subaddress it was sent to (the turnstile model), or requiring that a subaddress to spend from is always specified (so the Txos don’t get mixed together and create inconsistent subaddress balances)

### In this PR
- [x] Adds `require_spend_subaddresses` to the Account db as bool column
- [x] Adds database migration 
- [x] Adds `require_spend_subaddresses` to the APIs for create and import account
- [x] Adds logic to enforce building a transaction with an account with this enabled or not uses the correct params
- [x] Renames `subaddress_to_spend_from` to `spend_from_subaddress` for brevity & clarity
- [x] Adds tests for mode mismatches, and updates existing tests for `spend_from_subaddress` to correctly initialize the account with `require_spend_subaddress`
- [x] Adds toggle to enable or disable this mode for an account
- [x] Updates documentation for these API endpoints (Updates [here](https://mobilecoin.gitbook.io/full-service-api/~/changes/NoNZ9e0QGYDK8Uc5RmvJ?r=hQYgz6nipxmR5umItpAR))

### Test Plan

- [x] Update unittests for existing `spend_from_subaddress` tests
- [x] Update e2e test and add new e2e tests for mismatches to enforce failures
- [ ] ~Add new unittests for require subaddress functionality in `transaction` service~ Going with minimum viable e2e testing

### Future Work
* Add a `--strict-params` mode to prevent misspellings from accidentally omitting the parameter

### DIFF
--- a/full-service/migrations/2024-05-21-035622_subaddress-only-account/down.sql
+++ b/full-service/migrations/2024-05-21-035622_subaddress-only-account/down.sql
@@ -1,2 +1,2 @@
 -- This file should undo anything in `up.sql`
-ALTER TABLE accounts DROP COLUMN require_spend_subaddresses;
+ALTER TABLE accounts DROP COLUMN require_spend_subaddress;
--- a/full-service/migrations/2024-05-21-035622_subaddress-only-account/down.sql
+++ b/full-service/migrations/2024-05-21-035622_subaddress-only-account/down.sql
@@ -1,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE accounts DROP COLUMN spend_only_from_subaddress;
--- a/full-service/migrations/2024-05-21-035622_subaddress-only-account/down.sql
+++ b/full-service/migrations/2024-05-21-035622_subaddress-only-account/down.sql
@@ -1,2 +1,2 @@
 -- This file should undo anything in `up.sql`
-ALTER TABLE accounts DROP COLUMN spend_only_from_subaddress;
+ALTER TABLE accounts DROP COLUMN require_spend_subaddresses;
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -563,7 +563,7 @@ impl AccountModel for Account {
             fog_enabled,
             view_only: false,
             managed_by_hardware_wallet: false,
-            spend_only_from_subaddress: require_spend_subaddresses,
+            require_spend_subaddresses: require_spend_subaddresses,
         };
 
         diesel::insert_into(accounts::table)
@@ -676,7 +676,7 @@ impl AccountModel for Account {
             fog_enabled: false,
             view_only: true,
             managed_by_hardware_wallet,
-            spend_only_from_subaddress: false,
+            require_spend_subaddresses: false,
         };
 
         diesel::insert_into(accounts::table)
@@ -747,7 +747,7 @@ impl AccountModel for Account {
             fog_enabled: true,
             view_only: true,
             managed_by_hardware_wallet: true,
-            spend_only_from_subaddress: false,
+            require_spend_subaddresses: false,
         };
 
         diesel::insert_into(accounts::table)
--- a/full-service/src/db/account.rs
+++ b/full-service/src/db/account.rs
@@ -82,7 +82,7 @@ pub trait AccountModel {
     ///| `name`                  | The display name for the account.                                       | A label can have duplicates, but it is not recommended.               |
     ///| `fog_report_url`        | Fog Report server url.                                                  | Applicable only if user has Fog service, empty string otherwise.      |
     ///| `fog_authority_spki`    | Fog Authority Subject Public Key Info.                                  | Applicable only if user has Fog service, empty string otherwise.      |
-    ///| `spend_only_from_subaddress_mode` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
+    ///| `require_spend_subaddresses` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
     ///
     /// # Returns:
     /// * (account_id, main_subaddress_b58)
@@ -95,7 +95,7 @@ pub trait AccountModel {
         name: &str,
         fog_report_url: String,
         fog_authority_spki: String,
-        spend_only_from_subaddress_mode: bool,
+        require_spend_subaddresses: bool,
         conn: Conn,
     ) -> Result<(AccountID, String), WalletDbError>;
 
@@ -112,7 +112,7 @@ pub trait AccountModel {
     ///| `name`                  | The display name for the account.                                       | A label can have duplicates, but it is not recommended.               |
     ///| `fog_report_url`        | Fog Report server url.                                                  | Applicable only if user has Fog service, empty string otherwise.      |
     ///| `fog_authority_spki`    | Fog Authority Subject Public Key Info.                                  | Applicable only if user has Fog service, empty string otherwise.      |
-    ///| `spend_only_from_subaddress_mode` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
+    ///| `require_spend_subaddresses` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
     ///| `conn`                  | An reference to the pool connection of wallet database                  |                                                                       |
     ///
     /// # Returns:
@@ -126,7 +126,7 @@ pub trait AccountModel {
         name: &str,
         fog_report_url: String,
         fog_authority_spki: String,
-        spend_only_from_subaddress_mode: bool,
+        require_spend_subaddresses: bool,
         conn: Conn,
     ) -> Result<(AccountID, String), WalletDbError>;
 
@@ -144,7 +144,7 @@ pub trait AccountModel {
     ///| `next_subaddress_index`  | This index represents the next subaddress to be assigned as an address.                           | This is useful information in case the account is imported elsewhere. |
     ///| `name`                   | The display name for the account.                                                                 | A label can have duplicates, but it is not recommended.               |
     ///| `fog_enabled`            | Indicate if fog server is enabled or disabled                                                     |                                                                       |
-    ///| `spend_only_from_subaddress_mode` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
+    ///| `require_spend_subaddresses` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
     ///| `conn`                   | An reference to the pool connection of wallet database                                            |                                                                       |
     ///
     /// # Returns:
@@ -159,7 +159,7 @@ pub trait AccountModel {
         next_subaddress_index: Option<u64>,
         name: &str,
         fog_enabled: bool,
-        spend_only_from_subaddress_mode: bool,
+        require_spend_subaddresses: bool,
         conn: Conn,
     ) -> Result<(AccountID, String), WalletDbError>;
 
@@ -176,7 +176,7 @@ pub trait AccountModel {
     ///| `next_subaddress_index` | This index represents the next subaddress to be assigned as an address. | This is useful information in case the account is imported elsewhere. |
     ///| `fog_report_url`        | Fog Report server url.                                                  | Applicable only if user has Fog service, empty string otherwise.      |
     ///| `fog_authority_spki`    | Fog Authority Subject Public Key Info.                                  | Applicable only if user has Fog service, empty string otherwise.      |
-    ///| `spend_only_from_subaddress_mode` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
+    ///| `require_spend_subaddresses` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
     ///| `conn`                  | An reference to the pool connection of wallet database                  |                                                                       |
     ///
     /// # Returns:
@@ -190,7 +190,7 @@ pub trait AccountModel {
         next_subaddress_index: Option<u64>,
         fog_report_url: String,
         fog_authority_spki: String,
-        spend_only_from_subaddress_mode: bool,
+        require_spend_subaddresses: bool,
         conn: Conn,
     ) -> Result<Account, WalletDbError>;
 
@@ -207,7 +207,7 @@ pub trait AccountModel {
     ///| `next_subaddress_index` | This index represents the next subaddress to be assigned as an address. | This is useful information in case the account is imported elsewhere. |
     ///| `fog_report_url`        | Fog Report server url.                                                  | Applicable only if user has Fog service, empty string otherwise.      |
     ///| `fog_authority_spki`    | Fog Authority Subject Public Key Info.                                  | Applicable only if user has Fog service, empty string otherwise.      |
-    ///| `spend_only_from_subaddress_mode` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
+    ///| `require_spend_subaddresses` | If enabled, this mode requires all transactions to spend from a provided subaddress |                                                      |
     ///| `conn`                  | An reference to the pool connection of wallet database                  |                                                                       |
     ///
     /// # Returns:
@@ -221,7 +221,7 @@ pub trait AccountModel {
         next_subaddress_index: Option<u64>,
         fog_report_url: String,
         fog_authority_spki: String,
-        spend_only_from_subaddress_mode: bool,
+        require_spend_subaddresses: bool,
         conn: Conn,
     ) -> Result<Account, WalletDbError>;
 
@@ -463,7 +463,7 @@ impl AccountModel for Account {
         name: &str,
         fog_report_url: String,
         fog_authority_spki: String,
-        spend_only_from_subaddress_mode: bool,
+        require_spend_subaddresses: bool,
         conn: Conn,
     ) -> Result<(AccountID, String), WalletDbError> {
         let fog_enabled = !fog_report_url.is_empty();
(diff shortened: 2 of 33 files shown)


full-service/src/db/account.rs
```diff
@@ -964,6 +986,7 @@ mod tests {
                 "Alice's Main Account",
                 "".to_string(),
                 "".to_string(),
+                None,
```
lnsiegel, line 989: Should these fields be commented to indicate which values they are initializing?

lnsiegel (commented):
All the initial pass plumbing for subaddress_to_spend_from and subaddress_to_spend_from_mode looks good.  

The description says that the PR "Adds logic to enforce building a transaction with an account with this enabled or not uses the correct params".  But I'm not seeing the code that implements that logic.  Maybe I'm missing it somehow.  If not, should it be part of the PR?

full-service/src/service/transaction.rs
```diff
@@ -468,6 +468,22 @@ where
...
+                    return Err(TransactionServiceError::TransactionBuilder(WalletTransactionBuilderError::InvalidArgument(
+                        "This account is configured to spend only from a specific subaddress. Please provide a subaddress to spend from.".to_string()
+                    )));
+                }
+            } else {
+                if subaddress_to_spend_from.is_some() {
+                    return Err(TransactionServiceError::TransactionBuilder(WalletTransactionBuilderError::InvalidArgument(
+                        "This account is not configured to spend only from a specific subaddress. Please do not provide a subaddress to spend from.".to_string()
+                    )));
+                }
```
nick-mobilecoin, line 484: Is this a limitation we actually want to impose?
I could understand as a potential early warning that someone didn't set the spend from subaddress up correctly. 
but inherently it doesn't seem like this is an issue if someone for some reason wanted to drain their change account or similar.

full-service/migrations/2024-05-21-035622_subaddress-only-account/down.sql
```diff
@@ -0,0 +1,2 @@
+-- This file should undo anything in `up.sql`
```
sugargoat, line 1: A migration is required because we are adding a new field to Account, in order to allow a clear delineation of the Account's mode

full-service/src/json_rpc/v2/api/request.rs
```diff
@@ -57,7 +57,7 @@ pub enum JsonCommandRequest {
         block_version: Option<String>,
         sender_memo_credential_subaddress_index: Option<String>,
         payment_request_id: Option<String>,
-        subaddress_to_spend_from: Option<String>,
+        spend_from_subaddress: Option<String>,
```
sugargoat, line 60: Renaming from previous PR that introduced these, assuming this feature has not been adopted yet, so no need for backward compatibility

full-service/src/json_rpc/v2/api/request.rs
```diff
@@ -121,6 +121,8 @@ pub enum JsonCommandRequest {
     create_account {
         name: Option<String>,
         fog_info: Option<FogInfo>,
+        #[serde(default = "bool::default")] // default is false
+        require_spend_subaddresses: bool,
```
sugargoat, line 125: 📌  This is the main motivation for this PR

full-service/src/json_rpc/v2/api/wallet.rs
```diff
@@ -172,7 +172,7 @@ where
             block_version,
             sender_memo_credential_subaddress_index,
             payment_request_id,
-            subaddress_to_spend_from,
+            spend_from_subaddress,
```
sugargoat, line 175: Renaming from previous PR that introduced these. Assuming no need for backward compatibility

full-service/src/json_rpc/v2/api/wallet.rs
```diff
@@ -504,11 +504,20 @@ where
                 txo: txo_status_and_memo.map(|txo_info| (&txo_info).into()),
             }
         }
-        JsonCommandRequest::create_account { name, fog_info } => {
+        JsonCommandRequest::create_account {
+            name,
+            fog_info,
+            require_spend_subaddresses,
```
sugargoat, line 510: 📌 This is the main motivation for this PR

full-service/src/service/transaction.rs
```diff
@@ -468,6 +468,22 @@ where
         let conn = pooled_conn.deref_mut();
 
         exclusive_transaction(conn, |conn| {
+            if Account::get(&AccountID(account_id_hex.to_string()), conn)?
```
sugargoat, line 471: 📌  This is the main logic enforcing the match between `requires_spend_subaddresses` on an account, and providing `spend_from_subaddress` to the `build` methods

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_with_spend_from_subaddress.rs
```diff
@@ -307,4 +308,182 @@ mod e2e_transaction {
         assert_eq!(secreted, "0");
         assert_eq!(orphaned, "0");
     }
+
+    #[test_with_logger]
+    fn test_build_and_submit_transaction_with_require_spend_subaddress_mismatch_fails_if_set(
```
sugargoat, line 313: Both these tests need a little clean up 📋

full-service/src/service/transaction.rs
holtzman: Good point. I think that it would be ok to spend from a subaddress even when it is not required.

holtzman:
@sugargoat I like the renaming, and while we are at it, what about if we make it:

- `spend_subaddress` for the field that specifies which subaddress to spend from, and
- `require_spend_subaddress` for the mode that requires it?

This would be consistent and concise.  If you are ok with this, I'd be happy to do the search-replace and push a commit.

full-service/src/db/account.rs
holtzman, line 1: import_view_only also needs require_spend_subaddress added.  build_unsigned_transaction operates on a view only account for offline transaction signing and there is no reason why offline transaction signing and spending from subaddresses wouldn't be used together.

lnsiegel (commented):
Renaming and init defaults look good to me!

full-service/src/service/transaction.rs
```diff
@@ -468,6 +468,22 @@ where
...
+                .require_spend_subaddresses
+            {
+                if spend_from_subaddress.is_none() {
+                    return Err(TransactionServiceError::TransactionBuilder(WalletTransactionBuilderError::InvalidArgument(
```
nick-mobilecoin, line 475: ❓ Does `InvalidArgument()` make sense for a missing argument?  Kind of depends how one looks at it, as it's "invalid json request" so perhaps it's sufficient.

I see the following, but it doesn't appear to be used in the code base
```rust
    /// Attempting to build a transaction from a TXO without a subaddress: {0}
    NullSubaddress(String),
```

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_with_spend_from_subaddress.rs
```diff
@@ -307,4 +308,182 @@ mod e2e_transaction {
...
+                "amount": { "value": "42000000000000", "token_id": "0" }, // 42.0 MOB
+            }
+        });
+        let res = dispatch(&client, body, &logger);
+        let error = res.get("error").unwrap();
+        let data = error.get("data").unwrap();
+        let details = data.get("details").unwrap();
+        assert!(details
+            .to_string()
+            .contains(&transaction_error("This account is configured to spend only from a specific subaddress. Please provide a subaddress to spend from.".to_string()).to_string()));
```
nick-mobilecoin, line 398: tl;dr Nothing actionable here, just me capturing thoughts from the squirrel that runs around in my skull

I think it's good that this is testing the error condition as the error cases often get skipped in testing and they're the ones that may have the most impact on the end user.
Saying all that, checking for a large substring feels like a future pain point in that anytime someone wants to update the error message they have to update the test. Maybe that's a good thing since in many ways this error message is the public API to the user.

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_with_spend_from_subaddress.rs
```diff
@@ -307,4 +308,182 @@ mod e2e_transaction {
...
+        let (client, mut ledger_db, db_ctx, _network_state) = setup(&mut rng, logger.clone());
+
+        // Add an account
+        let body = json!({
+            "jsonrpc": "2.0",
+            "id": 1,
+            "method": "create_account",
+            "params": {
+                "name": "Exchange Main Account",
+            }
```
nick-mobilecoin, line 416: 👍 I like that this is testing the default via the omission of the argument.

full-service/src/service/transaction.rs
```diff
@@ -468,6 +468,22 @@ where
...
+                        "This account is configured to spend only from a specific subaddress. Please provide a subaddress to spend from.".to_string()
+                    )));
+                }
+            } else {
+                if spend_from_subaddress.is_some() {
+                    return Err(TransactionServiceError::TransactionBuilder(WalletTransactionBuilderError::InvalidArgument(
+                        "This account is not configured to spend only from a specific subaddress. Please do not provide a subaddress to spend from.".to_string()
+                    )));
+                }
+            }
```
nick-mobilecoin, line 485: I might lean a bit hard on preferring match statements over if/else so feel free to take no action on this suggestion.

The below suggestion came from co-pilot.

me: is there a more concise way to implement the highlighted code

co-pilot:
Yes, the highlighted code can be made more concise by using the `match` statement in Rust. Here's how you can do it:

```rust
match (Account::get(&AccountID(account_id_hex.to_string()), conn)?.require_spend_subaddresses, spend_from_subaddress) {
    (true, None) => {
        return Err(TransactionServiceError::TransactionBuilder(WalletTransactionBuilderError::InvalidArgument(
            "This account is configured to spend only from a specific subaddress. Please provide a subaddress to spend from.".to_string()
        )));
    }
    (false, Some(_)) => {
        return Err(TransactionServiceError::TransactionBuilder(WalletTransactionBuilderError::InvalidArgument(
            "This account is not configured to spend only from a specific subaddress. Please do not provide a subaddress to spend from.".to_string()
        )));
    }
    _ => {}
}
```

This code does the same thing as the original code but in a more concise way. It checks the combination of `require_spend_subaddresses` and `spend_from_subaddress` and returns an error if necessary.

I modified a bit to make it easier to grok once cargo fmt has it's way
```suggestion
            let require_spend_subaddresses =
                Account::get(&AccountID(account_id_hex.to_string()), conn)?
                    .require_spend_subaddresses;
            match (require_spend_subaddresses, &spend_from_subaddress) {
                (true, None) => {
                    return Err(TransactionServiceError::TransactionBuilder(WalletTransactionBuilderError::InvalidArgument(
                        "This account is configured to spend only from a specific subaddress. Please provide a subaddress to spend from.".to_string()
                    )));
                }
                (false, Some(_)) => {
                    return Err(TransactionServiceError::TransactionBuilder(WalletTransactionBuilderError::InvalidArgument(
                        "This account is not configured to spend only from a specific subaddress. Please do not provide a subaddress to spend from.".to_string()
                    )));
                }
                _ => {}
            }
```

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_with_spend_from_subaddress.rs
```diff
@@ -307,4 +308,182 @@ mod e2e_transaction {
...
+
+        // Imagine that Bob is sending 42.0 MOB to Alice
+        let body = json!({
+            "jsonrpc": "2.0",
+            "id": 1,
+            "method": "build_and_submit_transaction",
+            "params": {
+                "account_id": account_id,
+                "recipient_public_address": alice_b58_public_address,
+                "amount": { "value": "42000000000000", "token_id": "0" }, // 42.0 MOB
```
nick-mobilecoin, line 477: nit:
```suggestion
                "amount": { "value": (42 * MOB).to_string(), "token_id": "0" },
```

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_with_spend_from_subaddress.rs
```diff
@@ -307,4 +308,182 @@ mod e2e_transaction {
...
```
nick-mobilecoin, line 389: nit:
```suggestion
                "amount": { "value": (42 * MOB).to_string(), "token_id": "0" },
```

nick-mobilecoin (commented):
Everything looks good, just need:
1. The build_unsigned_transaction @holtzman mentioned
2. Be sure we want to prevent one from spending from a subaddress if the account is not a require subaddress spend

full-service/src/service/transaction.rs
sugargoat: Very good point - I was also thinking about the case of an onramp...

If I create a subaddress for someone, as part of my account establishment, but the first thing they want to do is onramp, then I'm going to need to send eUSD from somewhere...that's not their account. Or even if they had some TXOs, but they're onramping more than they've deposited, I still may presumably want to send them funds from another subaddress (maybe that I set up specifically for this purpose)

full-service/src/db/account.rs
sugargoat: I'm ok with leaving these uncommented, with the assumption that most of us are using an IDE that will expand the field name in the text view

full-service/src/service/transaction.rs
sugargoat: But actually, both of those cases would be fine even if `require_spend_from_subaddress` was set. Disregard :)

sugargoat:
> @sugargoat I like the renaming, and while we are at it, what about if we make it:
> 
> * `spend_subaddress` for the field that specifies which subaddress to spend from, and
> * `require_spend_subaddress` for the mode that requires it?
> 
> This would be consistent and concise. If you are ok with this, I'd be happy to do the search-replace and push a commit.

No worries - happy to get the renaming in order!

full-service/src/db/account.rs
sugargoat: Done!

full-service/src/db/schema.rs
```diff
@@ -1,5 +1,3 @@
-// @generated automatically by Diesel CLI.
```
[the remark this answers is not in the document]
sugargoat: Hmm I wonder if I accidentally removed this - I had to go back to the file to add in `diesel::` prefix to each `table!` declaration.

I'm on diesel 1.4.1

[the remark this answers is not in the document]
sugargoat: Yeah, hm I added the line back in, ran the migration again, and it deleted it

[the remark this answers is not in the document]
sugargoat: I'll put the line back in though - no need to change it for this PR

full-service/src/db/schema.rs
```diff
@@ -116,6 +115,7 @@ diesel::table! {
 
 diesel::joinable!(assigned_subaddresses -> accounts (account_id));
 diesel::joinable!(authenticated_sender_memos -> txos (txo_id));
+diesel::joinable!(destination_memos -> txos (txo_id));
```
sugargoat, line 119: Running the migration picked up on this change - seems legit, and like it should be in here

full-service/src/service/transaction.rs
sugargoat: Oh wow good find

sugargoat: I'm stealing this error for our purposes. From context, I imagine maybe this was plumbed originally so you couldn't spend an orphaned TXO, but since it's not used anywhere, we'll commandeer it.

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_and_submit_with_spend_from_subaddress.rs
sugargoat: Agreed - I think we could use an error overhaul - @holtzman and I were talking about how that might take shape. Will leave as future work.

full-service/src/service/transaction.rs
sugargoat: In removing the error state if you have `NOT require` and still provide a `spend_subaddress`, there's only one condition now, so I'll leave it as `if`

full-service/src/db/account.rs
```diff
@@ -212,6 +221,7 @@ pub trait AccountModel {
         next_subaddress_index: Option<u64>,
         fog_report_url: String,
         fog_authority_spki: String,
+        require_spend_subaddresses: bool,
```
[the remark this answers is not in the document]
sugargoat: I _think_ the v1 here is referring to the legacy (v1) entropy format, as opposed to the v1 API - let me double check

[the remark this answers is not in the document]
sugargoat: I feel pretty confident that's the case, looking at line 133 `Create an account from either mnemonic phrase (v2) or root entropy (v1).`

full-service/src/service/transaction.rs
sugargoat: Addressed! You can now spend from a subaddress even if the `require spend subaddress` mode was not enabled for that account

holtzman (approved):
Looks good. Thanks!

PR #993, opened 2024-05-28
add bip39_entropy as alternative to mnemonic for signer-service

holtzman:
### Motivation

Offline signers have expressed a desire to store and manage the secret entropy of their account as a 32 byte value rather than as a 24 word mnemonic phrase for consistency with other blockchain wallets that their software manages this way.

Herein we introduce `bip39_entropy`, a 32 byte, hex-encoded string (64 characters) to the `signer-service` API to satisfy this need.  When outputting account keys, signing txos, or syncing txos for a `view-only` wallet, users of the `signer-service` can now chose  between providing a 24 word phrase, `mnemonic`, or a 64 character hex-encoded 32 byte secret, using `bip39_entropy`.

### In this PR
* json rpc request data structures updated to make `mnemonic` optional and to optionally allow `bip39_entropy`.
* service calls updated to require exactly one of `mnemonic` or `bip39_entropy` and to use whichever is provided.

### DIFF
--- a/signer/src/service/api/mod.rs
+++ b/signer/src/service/api/mod.rs
@@ -72,31 +72,98 @@ fn signer_service_api_inner(command: JsonCommandRequest) -> Result<JsonCommandRe
                 account_info,
             }
         }
-        JsonCommandRequest::get_account { mnemonic } => {
-            let account_info = service::get_account(&mnemonic)?;
-            JsonCommandResponse::get_account { account_info }
-        }
+        JsonCommandRequest::get_account {
+            mnemonic,
+            bip39_entropy,
+        } => match mnemonic {
+            Some(mnemonic) => match bip39_entropy {
+                Some(_) => {
+                    return Err(anyhow!(
+                        "Only one of mnemonic or bip39_entropy can be provided"
+                    ));
+                }
+                None => {
+                    let account_info = service::get_account_by_mnemonic(&mnemonic)?;
+                    JsonCommandResponse::get_account { account_info }
+                }
+            },
+            None => match bip39_entropy {
+                Some(bip39_entropy) => {
+                    let account_info = service::get_account_by_bip39_entropy(&bip39_entropy)?;
+                    JsonCommandResponse::get_account { account_info }
+                }
+                None => {
+                    return Err(anyhow!("Either mnemonic or bip39_entropy must be provided"));
+                }
+            },
+        },
         JsonCommandRequest::sign_tx {
             mnemonic,
+            bip39_entropy,
             unsigned_tx_proposal,
-        } => {
-            let signed_tx = service::sign_tx(
-                &mnemonic,
-                (&unsigned_tx_proposal)
-                    .try_into()
-                    .map_err(|e: String| anyhow!(e))?,
-            )?;
-            JsonCommandResponse::sign_tx {
-                tx_proposal: (&signed_tx).try_into().map_err(|e: String| anyhow!(e))?,
-            }
-        }
+        } => match mnemonic {
+            Some(mnemonic) => match bip39_entropy {
+                Some(_) => {
+                    return Err(anyhow!(
+                        "Only one of mnemonic or bip39_entropy can be provided"
+                    ));
+                }
+                None => {
+                    let signed_tx = service::sign_tx_with_mnemonic(
+                        &mnemonic,
+                        (&unsigned_tx_proposal)
+                            .try_into()
+                            .map_err(|e: String| anyhow!(e))?,
+                    )?;
+                    JsonCommandResponse::sign_tx {
+                        tx_proposal: (&signed_tx).try_into().map_err(|e: String| anyhow!(e))?,
+                    }
+                }
+            },
+            None => match bip39_entropy {
+                Some(bip39_entropy) => {
+                    let signed_tx = service::sign_tx_with_bip39_entropy(
+                        &bip39_entropy,
+                        (&unsigned_tx_proposal)
+                            .try_into()
+                            .map_err(|e: String| anyhow!(e))?,
+                    )?;
+                    JsonCommandResponse::sign_tx {
+                        tx_proposal: (&signed_tx).try_into().map_err(|e: String| anyhow!(e))?,
+                    }
+                }
+                None => {
+                    return Err(anyhow!("Either mnemonic or bip39_entropy must be provided"));
+                }
+            },
+        },
         JsonCommandRequest::sync_txos {
             mnemonic,
+            bip39_entropy,
             txos_unsynced,
-        } => {
-            let txos_synced = service::sync_txos(&mnemonic, txos_unsynced)?;
-            JsonCommandResponse::sync_txos { txos_synced }
-        }
+        } => match mnemonic {
+            Some(mnemonic) => match bip39_entropy {
+                Some(_) => {
+                    return Err(anyhow!(
+                        "Only one of mnemonic or bip39_entropy can be provided"
+                    ));
+                }
+                None => {
+                    let txos_synced = service::sync_txos_by_mnemonic(&mnemonic, txos_unsynced)?;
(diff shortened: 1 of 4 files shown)


signer/src/service/api/mod.rs
```diff
@@ -72,31 +72,98 @@ fn signer_service_api_inner(command: JsonCommandRequest) -> Result<JsonCommandRe
...
+        JsonCommandRequest::get_account {
+            mnemonic,
+            bip39_entropy,
+        } => match mnemonic {
+            Some(mnemonic) => match bip39_entropy {
+                Some(_) => {
+                    return Err(anyhow!(
+                        "Only one of mnemonic or bip39_entropy can be provided"
+                    ));
+                }
```
nick-mobilecoin, line 84: This could probably be done at a later time, but if we're going to force only one of mnemonic or bip39 entropy then I would suggest that we create an enum for them and handle it at the decoding level so that we don't need to have this logic triplicated

I _think_ it would be something like the following, I'm not fond of the name I chose.
Again we can come back later, and this PR can focus on getting this change out for a fast turnaround.
```rust
#[derive(Deserialize, Serialize, EnumIter, Debug)]
#[serde(tag = "method", content = "params")]
#[allow(non_camel_case_types)]
pub enum Derivation {
    Mnemonic(String),
    Bip39Entropy(String),
}

/// Requests to the Signer Service.
#[derive(Deserialize, Serialize, EnumIter, Debug)]
#[serde(tag = "method", content = "params")]
#[allow(non_camel_case_types)]
pub enum JsonCommandRequest {
    create_account {},
    get_account {
        derivation: Derivation
    },
    sign_tx {
        derivation: Derivation,
        unsigned_tx_proposal: UnsignedTxProposal,
    },
    sync_txos {
        derivation: Derivation,
        txos_unsynced: Vec<TxoUnsynced>,
    },
}
```

signer/src/service/api/mod.rs
```diff
@@ -72,31 +72,98 @@ fn signer_service_api_inner(command: JsonCommandRequest) -> Result<JsonCommandRe
...
+            },
+            None => match bip39_entropy {
+                Some(bip39_entropy) => {
+                    let account_info = service::get_account_by_bip39_entropy(&bip39_entropy)?;
+                    JsonCommandResponse::get_account { account_info }
+                }
+                None => {
+                    return Err(anyhow!("Either mnemonic or bip39_entropy must be provided"));
+                }
+            },
```
nick-mobilecoin, line 98: suggest passing both values in to the match for  a more concise approach
```suggestion
        } => match (mnemonic, bip39_entropy) {
            (Some(mnemonic), None) => {
                let account_info = service::get_account_by_mnemonic(&mnemonic)?;
                JsonCommandResponse::get_account { account_info }
            }
            (None, Some(bip39_entropy)) => {
                let account_info = service::get_account_by_bip39_entropy(&bip39_entropy)?;
                JsonCommandResponse::get_account { account_info }
            }
            _ => {
                return Err(anyhow!(
                    "Only one of mnemonic or bip39_entropy can be provided"
                ))
            }
```

signer/src/service/api/mod.rs
```diff
@@ -72,31 +72,98 @@ fn signer_service_api_inner(command: JsonCommandRequest) -> Result<JsonCommandRe
...
+                            .map_err(|e: String| anyhow!(e))?,
+                    )?;
+                    JsonCommandResponse::sign_tx {
+                        tx_proposal: (&signed_tx).try_into().map_err(|e: String| anyhow!(e))?,
+                    }
+                }
+                None => {
+                    return Err(anyhow!("Either mnemonic or bip39_entropy must be provided"));
+                }
+            },
```
nick-mobilecoin, line 138: similar to above
```suggestion
            (Some(mnemonic), None) => {
                let signed_tx = service::sign_tx_with_mnemonic(
                    &mnemonic,
                    (&unsigned_tx_proposal)
                        .try_into()
                        .map_err(|e: String| anyhow!(e))?,
                )?;
                JsonCommandResponse::sign_tx {
                    tx_proposal: (&signed_tx).try_into().map_err(|e: String| anyhow!(e))?,
                }
            }
            (None, Some(bip39_entropy)) => {
                let signed_tx = service::sign_tx_with_bip39_entropy(
                    &bip39_entropy,
                    (&unsigned_tx_proposal)
                        .try_into()
                        .map_err(|e: String| anyhow!(e))?,
                )?;
                JsonCommandResponse::sign_tx {
                    tx_proposal: (&signed_tx).try_into().map_err(|e: String| anyhow!(e))?,
                }
            }
            _ => {
                return Err(anyhow!(
                    "Only one of mnemonic or bip39_entropy can be provided"
                ))
            }
```

signer/src/service/api/mod.rs
```diff
@@ -72,31 +72,98 @@ fn signer_service_api_inner(command: JsonCommandRequest) -> Result<JsonCommandRe
...
+            None => match bip39_entropy {
+                Some(bip39_entropy) => {
+                    let txos_synced =
+                        service::sync_txos_by_bip39_entropy(&bip39_entropy, txos_unsynced)?;
+                    JsonCommandResponse::sync_txos { txos_synced }
+                }
+                None => {
+                    return Err(anyhow!("Either mnemonic or bip39_entropy must be provided"));
+                }
+            },
```
nick-mobilecoin, line 165: similar to above
```suggestion
            (Some(mnemonic), None) => {
                let txos_synced = service::sync_txos_by_mnemonic(&mnemonic, txos_unsynced)?;
                JsonCommandResponse::sync_txos { txos_synced }
            }
            (None, Some(bip39_entropy)) => {
                let txos_synced =
                    service::sync_txos_by_bip39_entropy(&bip39_entropy, txos_unsynced)?;
                JsonCommandResponse::sync_txos { txos_synced }
            }
            _ => {
                return Err(anyhow!(
                    "Only one of mnemonic or bip39_entropy can be provided"
                ))
            }
```

signer/src/service/mod.rs
```diff
@@ -24,8 +24,19 @@ pub fn create_account() -> (Mnemonic, AccountInfo) {
...
 
-pub fn get_account(mnemonic: &str) -> Result<AccountInfo> {
+pub fn get_account_by_mnemonic(mnemonic: &str) -> Result<AccountInfo> {
     let mnemonic = Mnemonic::from_phrase(mnemonic, Language::English)?;
+    get_account(mnemonic)
+}
+
+pub fn get_account_by_bip39_entropy(bip39_entropy: &str) -> Result<AccountInfo> {
+    let mut entropy = [0u8; 32];
+    hex::decode_to_slice(bip39_entropy, &mut entropy)?;
```
nick-mobilecoin, line 34: 🤦 I've been decoding to a vec and then using `try_into()`, this is much more concise and less likely for someone to use `clone_from_slice` which panics if the sizes are wrong.

signer/src/service/mod.rs
```diff
@@ -24,8 +24,19 @@ pub fn create_account() -> (Mnemonic, AccountInfo) {
     (mnemonic, account_info)
 }
 
-pub fn get_account(mnemonic: &str) -> Result<AccountInfo> {
+pub fn get_account_by_mnemonic(mnemonic: &str) -> Result<AccountInfo> {
```
nick-mobilecoin, line 27: 👍 on the rename instead of just adding the bip39 version

signer/src/service/api/mod.rs
nick-mobilecoin: We should definitly hold off here, thinking about it more I think it will just let the last one win, if both show up in the json.

nick-mobilecoin (approved):
main suggestion is to consider matching on a tuple.
I can understand if it feels too dense, but I would still encourage one to move that direction as I think it's more readable than the nested match statements

sugargoat (approved):
LGTM & thank you @nick-mobilecoin for a great review too!

PR #998, opened 2024-06-06
Introduce webhook callback in account sync

sugargoat:
### Motivation

Some ramps providers would like to receive a callback when TXOs are found for their accounts during syncing.

### In this PR
- [x] adds an optional configuration to include a webhook URL
- [x] Adds a WebhookThread being managed by the WalletService alongside account syncing
- [x] Adds a shared state via `AtomicUsize` to track how many new txos since last webhook call 

### Test Plan

- [x] new e2e test hitting webhook through httpmock during account sync


full-service/src/db/transaction_log.rs
```diff
@@ -892,7 +892,12 @@ mod tests {
         let mut ledger_db = get_test_ledger(5, &known_recipients, 12, &mut rng);
 
         // Start sync thread
-        let _sync_thread = SyncThread::start(ledger_db.clone(), wallet_db.clone(), logger.clone());
+        let _sync_thread = SyncThread::start(
+            ledger_db.clone(),
+            wallet_db.clone(),
+            "/webhook",
```
sugargoat, line 898: ⚠️ Changing the signature on `SyncThread::start` to include the URL - so that ultimately the config can launch the thread with the proper URL

full-service/src/error.rs
```diff
@@ -231,6 +231,9 @@ pub enum SyncError {
 
     /// Error with the B58 Util: {0}
     B58(B58Error),
+
+    /// Error calling webhook: {0}
+    Webhook(reqwest::Error),
```
sugargoat, line 236: New error type wrapping reqwest - only for Webhook right now

full-service/src/json_rpc/v2/api/test_utils.rs
```diff
@@ -58,6 +59,21 @@ pub struct TestWalletState {
...
+#[derive(Deserialize, Debug)]
+struct WebhookPayload {
+    // Define the expected fields in the webhook payload
+    field1: String,
+    field2: i32,
+}
+
+// Dummy webhook handler to verify we're sending the webhook
+#[get("/webhook", format = "json", data = "<payload>")]
+async fn handle_webhook(payload: Json<WebhookPayload>) -> Status {
```
sugargoat, line 72: Proof of concept to add webhook handler as part of our test scaffolding

full-service/src/json_rpc/v2/e2e_tests/webhook.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) &2020-2022 MobileCoin Inc.
+
+//! End-to-end tests for the Full Service Wallet API.
+
+#[cfg(test)]
+mod e2e_webhook {
```
sugargoat, line 6: ☝️ The main motivation for this change is tested here: during `account_sync`, the webhook fires that receives how many txos were received & spent in the scanned chunk

full-service/src/service/sync.rs
```diff
@@ -52,7 +56,12 @@ pub struct SyncThread {
 }
 
 impl SyncThread {
-    pub fn start(ledger_db: LedgerDB, wallet_db: WalletDb, logger: Logger) -> Self {
+    pub fn start(
+        ledger_db: LedgerDB,
+        wallet_db: WalletDb,
+        webhook_url: &str,
```
sugargoat, line 62: ⚠️  Signature change here - for strong typing, I would prefer to use `Url`, but at the same time, the reqwest library just takes a string

full-service/src/service/sync.rs
```diff
@@ -136,7 +145,24 @@ pub fn sync_all_accounts(
...
         }
-        sync_account_next_chunk(ledger_db, conn, &account.id, logger)?;
+        let (num_received_txos, num_spent_txos) =
+            sync_account_next_chunk(ledger_db, conn, &account.id, logger)?;
+        if num_received_txos > 0 || num_spent_txos > 0 {
+            // call webhook
+            // FIXME move this somewhere and configured by config
+            // FIXME: should we be using rocket? or we've already got reqwest throughout
+            // anyway?
+            let client = Client::builder().gzip(true).use_rustls_tls().build()?;
```
sugargoat, line 155: 🚧 Under construction 🚧 

Plan is to have the client instantiated at the top - keeping everything together for now

full-service/src/service/sync.rs
```diff
@@ -30,6 +30,10 @@ use mc_transaction_core::{
 };
 use rayon::prelude::*;
 
+use reqwest::{
+    blocking::Client, // FIXME: if we don't want this to be blocking, we'll be awaiting futures
```
sugargoat, line 34: ⚠️ Note - currently implemented with blocking client. When not using blocking client, everything needs to move to async. I wasn't ready to evaluate and make that change

full-service/src/service/sync.rs
```diff
@@ -147,7 +173,7 @@ pub fn sync_account_next_chunk(
     conn: Conn,
     account_id_hex: &str,
     logger: &Logger,
-) -> Result<(), SyncError> {
+) -> Result<(usize, usize), SyncError> {
```
sugargoat, line 176: 🚧 Under Construction 🚧 

Using usize right now because we're working with lengths of vectors, but also because usize = u32, and that should be safe sending as numbers through reqwest body

full-service/src/service/sync.rs
```diff
@@ -186,11 +212,11 @@ pub fn sync_account_next_chunk(
 
         // If no blocks were found, exit.
         if end_block_index.is_none() {
-            return Ok(());
+            return Ok((0, 0));
```
sugargoat, line 215: 🚧 Under Construction 🚧 

Not a huge fan of early outs in rust - I prefer to have one exit point at the end of the function. Partly because I'm also setting magic numbers here - why zero? Any reason this would change and not be 0 here?

full-service/src/service/wallet_service.rs
```diff
@@ -98,6 +98,7 @@ impl<
             Some(SyncThread::start(
                 ledger_db.clone(),
                 wallet_db,
+                "/wallet", // FIXME
```
sugargoat, line 101: ☝️ This is the main invocation of the sync thread for the running program

full-service/src/service/sync.rs
```diff
@@ -136,7 +145,24 @@ pub fn sync_all_accounts(
 
             continue;
         }
-        sync_account_next_chunk(ledger_db, conn, &account.id, logger)?;
+        let (num_received_txos, num_spent_txos) =
```
holtzman, line 148: Do we want to fire the webhook on spending? I was imagining only firing the webhook on deposits (new txos).

practically speaking, there may be little difference as RTH proscribes sending a 0 value change txp whenever a txo is spent.

sugargoat: great point - will just do received

full-service/src/service/sync.rs
holtzman: agreed that we should be leery of full-service having its core scanning loop blocked by a web get or post to a webhook.

full-service/src/config.rs
```diff
@@ -89,6 +89,10 @@ pub struct APIConfig {
     /// T3 Server to connect to and the api key to use for authorization.
     #[clap(flatten)]
     pub t3_sync_config: T3Config,
+
+    /// Webhook configuration.
+    #[clap(long, env = "ACCOUNT_SYNC_WEBHOOK_URL")]
+    pub account_sync_webhook_url: Option<String>, // FIXME: URL
```
sugargoat, line 95: ☝️ New configuration option: provide the webhook URL on full-service command line or in env var

full-service/src/config.rs
```diff
@@ -89,6 +89,10 @@ pub struct APIConfig {
     /// T3 Server to connect to and the api key to use for authorization.
     #[clap(flatten)]
     pub t3_sync_config: T3Config,
+
+    /// Webhook configuration.
+    #[clap(long, env = "ACCOUNT_SYNC_WEBHOOK_URL")]
```
holtzman, line 94: All our of our env variables start with MC_ currently, so adding that... Also this is a webhook that will be fired when there are new deposits, so suggesting alternate name for the argument.
```suggestion
    #[clap(long, env = "MC_DEPOSITS_WEBHOOK_URL")]
```

full-service/src/config.rs
holtzman:
```suggestion
    pub deposits_webhook_url: Option<String>, // FIXME: URL
```

full-service/src/db/transaction_log.rs
sugargoat: Went a different direction

full-service/src/error.rs
sugargoat: Went a different direction & eliminated

full-service/src/json_rpc/v2/api/test_utils.rs
sugargoat: Went a different direction & eliminated

full-service/src/service/sync.rs
sugargoat: Went a different direction & eliminated

full-service/src/service/sync.rs
sugargoat: Moved the webhook into its own WebhookThread - we may want to use this as a touchpoint to evaluate the async and whether we're happy with the concurrency model we're using for account sync as well

full-service/src/service/sync.rs
sugargoat: Moved to AtomicUsize

full-service/src/config.rs
sugargoat: Done! And added some documentation for this new feature

Cargo.lock
```diff
@@ -163,6 +163,58 @@ version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
```
sugargoat, line 163: ⚠️ There's a lot being pulled in here...will look into this

full-service/src/json_rpc/v2/e2e_tests/webhook.rs
```diff
@@ -0,0 +1,123 @@
...
+                &mut ledger_db,
+                &vec![public_address.clone()],
+                100 * MOB + (i as u64),
+                &[KeyImage::from(rng.next_u64())],
+                &mut rng,
+            );
+        }
+
+        // Wait for the account sync thread to finish syncing
+        let wallet_db = &db_ctx.get_db_instance(logger.clone());
```
sugargoat, line 103: @holtzman - doing this instead of `manually_sync_account`.

full-service/src/service/sync.rs
```diff
@@ -51,12 +55,111 @@ pub struct SyncThread {
...
 
+pub struct WebhookThread {
+    /// The main sync thread handle.
+    join_handle: Option<thread::JoinHandle<()>>,
+
+    /// Stop trigger, used to signal the thread to terminate.
+    stop_requested: Arc<AtomicBool>,
+}
+
+impl WebhookThread {
```
sugargoat, line 66: 🆕  This is the new Webhook Thread that checks the shared `AtomicUsize` `num_txos_synced` and fires the webhook callback

full-service/src/service/sync.rs
```diff
@@ -51,12 +55,111 @@ pub struct SyncThread {
...
+                            .unwrap()
+                            .swap(0, Ordering::Relaxed);
+                        log::debug!(
+                            logger,
+                            "checked num_txos_to_send atomicusize = {:?}",
+                            num_txos_to_send
+                        );
+
+                        if num_txos_to_send > 0 {
+                            let get_url = format!("{server_url}?num_txos={num_txos_to_send}");
```
sugargoat, line 111: 🚧 FIXME: URL

full-service/src/service/wallet_service.rs
```diff
@@ -91,17 +94,35 @@ impl<
...
+                Some(SyncThread::start(
+                    ledger_db.clone(),
+                    wallet_db,
+                    num_txos_synced.clone(),
+                    logger.clone(),
+                )),
+                // As a companion to the account syncing, start the webhook syncing
+                // if configured
+                if let Some(wh_config) = webhook_config {
+                    Some(WebhookThread::start(
```
sugargoat, line 115: 🆕 Where the webhook thread is started to be held by the WalletService

full-service/src/bin/main.rs
```diff
@@ -117,17 +117,30 @@ fn rocket() -> Rocket<Build> {
         tx_sources,
     };
 
+    let webhook_config = config.deposits_webhook_url.clone().map(|wu| WebhookConfig {
+        url: wu,
+        poll_interval: config.poll_interval.clone(),
```
sugargoat, line 122: 🆕 Now setting poll_interval to match overall config for polling consensus

full-service/src/config.rs
```diff
@@ -89,6 +90,33 @@ pub struct APIConfig {
...
     #[clap(flatten)]
     pub t3_sync_config: T3Config,
+
+    /// Webhook configuration to notify an external server listening for
+    /// deposit notifications.
+    ///
+    /// The format of the webhook is a POST request with the following query
+    /// parameters:
+    ///
+    /// POST /webhook -H "Content-Type: application/json" \
```
sugargoat, line 100: 🆕  Updated documentation for POST & accounts

full-service/src/db/account.rs
```diff
@@ -29,9 +29,10 @@ use mc_core::slip10::Slip10KeyGenerator;
 use mc_crypto_digestible::{Digestible, MerlinTranscript};
 use mc_crypto_keys::{RistrettoPrivate, RistrettoPublic};
 use mc_transaction_core::{get_tx_out_shared_secret, TokenId};
+use serde_derive::Serialize;
 use std::fmt;
 
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
```
sugargoat, line 34: Added `Serialize` so we can easily send the AccountIDs into `json!`

full-service/src/service/sync.rs
```diff
@@ -52,11 +53,18 @@ pub struct SyncThread {
 }
 
 impl SyncThread {
-    pub fn start(ledger_db: LedgerDB, wallet_db: WalletDb, logger: Logger) -> Self {
+    pub fn start(
+        ledger_db: LedgerDB,
+        wallet_db: WalletDb,
+        accounts_with_deposits: Arc<Mutex<HashMap<AccountID, bool>>>,
+        _restart: Arc<Mutex<AtomicBool>>, // FIXME: when does this get set and by what?
```
sugargoat, line 60: ⚠️ I'm not sure about the actual implementation of this piece of it cc @holtzman

full-service/src/service/sync.rs
```diff
@@ -110,14 +124,16 @@ impl Drop for SyncThread {
...
+    accounts_with_deposits: Arc<Mutex<HashMap<AccountID, bool>>>,
     logger: &Logger,
 ) -> Result<(), SyncError> {
     // Get the current number of blocks in ledger.
     let num_blocks = ledger_db
         .num_blocks()
         .expect("failed getting number of blocks");
     if num_blocks == 0 {
-        return Ok(());
+        return Ok(()); // FIXME: we want it to fire in this case with empty
```
sugargoat, line 135: ⚠️ I believe this is the case where we want it to fire with empty accounts - need to figure out how to make that actually happen

sugargoat:
This PR is now ready for review, and I would recommend that we leave the restart logic for a fast follow.

full-service/src/service/webhook.rs
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2018-2024 MobileCoin Inc.
+
+//! Manages sending a webhook for synced accounts that have received deposits
```
sugargoat, line 3: 🆕 Moved this to its own file - this is the main change: introduces the WebhookThread

full-service/src/service/sync.rs
holtzman: I think this is just a spot where the local ledger is empty and so there is nothing to do.

holtzman (approved):
Looks good! Some todos left but let's merge!

PR #999, opened 2024-06-07
update help for --deposits-webhook-url

holtzman:
### In this PR

The help text for --deposits-webhook-url is updated to match the latest behavior

### DIFF
--- a/full-service/src/config.rs
+++ b/full-service/src/config.rs
@@ -94,22 +94,18 @@ pub struct APIConfig {
     /// Webhook configuration to notify an external server listening for
     /// deposit notifications.
     ///
-    /// The format of the webhook is a POST request with the following query
+    /// The format of the webhook is a POST request with the following
     /// parameters:
     ///
     /// POST /webhook -H "Content-Type: application/json" \
-    ///     -d '{"accounts": [A,B,C], "restart": false}'
-    ///
-    /// The first time full-service is caught up with the network ledger,
-    /// it will send a webhook with {"restart": true, "accounts": [A,]}
-    /// where "accounts" may be empty.
-    ///
-    /// Where the num_txos provided indicate how many txos were received
-    /// in the last scan period for any account in the wallet.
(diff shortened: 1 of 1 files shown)


full-service/src/config.rs
```diff
@@ -94,23 +94,19 @@ pub struct APIConfig {
...
-    /// in the last scan period for any account in the wallet.
+    ///     -d '{"accounts": [A,B,C]}'
     ///
     /// The expected action to take in response to the webhook is to call
     /// the `get_txos` API endpoint for the given accounts to retrieve more
     /// details about the TXOs received.
     ///
+    /// It is also expected for the client to call get_txos on startup and
+    /// periodically to ensure that no TXOs are missed.
+    ///
```
nick-mobilecoin, line 109: does this contradict the webhook intent?
Or is the idea that a client could manually do this instead of leaving a something running, like the polling that was being done before?

nick-mobilecoin (approved):

full-service/src/config.rs
holtzman: The intention of the webhook is more to notify the larger system into which full-service is integrated that there has been activity in the accounts that full-service is watching.

This will allow the system to poll much less frequently while also be more responsive to deposits.

full-service is not:
- delivering details, beyond an account_id for the account for which 0 or more units of activity has occurred, where that account_id is a full-service specific hash of pubkey info such that it is a contextual pseudonymous identifier of the account.
 - retrying notifications upon errors
 - requesting acknowledgements that notifications were received
 - saving state across restarts as to whether or not an event fired a notification
 - promising to only notify once for any given event
 
Overall the expectation is that the client continues to be fully responsible for retrieving information from full-service and for synchronizing its local data with data from full-service.

It is recommended for client developers, therefore, to not entirely rely on the webhook to poll full-service and synchronize, but to also poll/synchronize at start-up to catch any deposits that happened while the client was offline, and from time-to-time to catch any missed notifications, which could occur because of networking issues or full-service outages.

PR #1001, opened 2024-07-18
FS-154 New TransactionId Derivation

lnsiegel:
### Motivation

`transaction_log_ids` created by full-service are used to lookup records in the `transaction_log` within the local wallet database.

Many third-party full-service clients expose this id to their users, expecting it to be the equivalent of other blockchain's transaction ids -- a hash that can be used to find evidence of the transaction on the blockchain itself.

Up until now, this has not been the case. While it was a 32 byte hash, represented as a 64 character hex string, the information being hashed included information private to the wallet database and so could not be used to find evidence of the transaction on the blockchain itself.

### In this PR

This PR updates the algorithm for creating new `transaction_log_ids` (historical ids will remain unchanged) to use public blockchain information only. Specifically, we now use the `public_key` of one of the transaction's `payload_txos` as the `transaction_log.id` for newly created `transaction_log` entries. This will result in improved experiences for when the `transaction_log.id` is shared with end users and used by them to, for example, search for the transaction in the block-explorer.

Meanwhile clients that use the `transaction_log.id` to look up `transaction_logs` can still use it the same as always.

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -63,6 +63,9 @@ impl TryFrom<Vec<OutputTxo>> for TransactionId {
         println!("{:?}", backtrace);
         println!("\nTRACK ====================== END =========================\n\n");
 
+        // Two ways to propagate an empty payload_txo vector error:
+        // return Err("Payload Txo Vector is empty");
+        // panic!("Payload Txo Vector is empty");
         if _payload_txos.is_empty() {
             return Err("Payload Txo Vector is empty");
         }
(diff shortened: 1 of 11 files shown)


full-service/src/db/transaction_log.rs
```diff
@@ -34,34 +30,34 @@ impl From<&TransactionLog> for TransactionId {
...
-        let temp: [u8; 32] = src.digest32::<MerlinTranscript>(b"transaction_data");
-        Self(hex::encode(temp))
+impl TryFrom<Vec<OutputTxo>> for TransactionId {
+    type Error = &'static str;
+    fn try_from(_payload_txos: Vec<OutputTxo>) -> Result<Self, Self::Error> {
+        Ok(Self(
+            HexFmt(
+                _payload_txos
+                    .iter()
+                    .min_by_key(|txo| txo.tx_out.public_key)
```
holtzman, line 54: build_transaction and submit_transaction should share the same TransactionId but it is possible for the payload_txos (when there are more than one) to be present in a different order.  using min_by_key here (instead of e.g. picking the first element of the Vec) guarantees that even if the payload_txos within the TxProposal get sorted between calls to this impl, the TransactionId will remain the same.

holtzman (approved):
LGTM

PR #1002, opened 2024-07-23
Unit tests for FS-154 New TransactionId Derivation

lnsiegel:
### Motivation

Added rust unit tests to confirm the new method of `transaction_log_id` generation for vectors of `OutputTxos` passed to the `TryFrom<Vec<OutputTxo>>`  implementation for `TransactionId`.

### In this PR

There are two new unit tests added:

1. `test_try_from_vec_output_txo_for_transaction_id`
2. `test_try_from_empty_vec_output_txo_for_transaction_id`

The first test creates different sized vectors of `OutputTxo` structures.  The vectors are of size 1 through 10.  The embedded `TxOut` structs are initialized from entropy, so the public keys are unique on each run.  Each vector is first scanned for its minimum public key.  The vector is then passed to `TransactionId::try_from()` and the resulting `transaction_log_id` is asserted to be equal to the vector's minimum public key.

The second test checks that the expected error message is returned when an empty vector is passed to `TransactionId::try_from()`

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -2024,8 +2024,7 @@ mod tests {
 
         use std::assert_matches::assert_matches;
 
-        let output_vec: Vec<OutputTxo> = Vec::new();
-        let transaction_log_id = TransactionId::try_from(output_vec);
+        let transaction_log_id = TransactionId::try_from(vec![]);
 
         assert_matches!(transaction_log_id, Err("no valid payload_txo"));
     }
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -53,7 +53,7 @@ impl TryFrom<Vec<OutputTxo>> for TransactionId {
                     .iter()
                     .map(|txo| txo.tx_out.public_key)
                     .min()
-                    .ok_or("no valid payload_txo")?
+                    .ok_or("no valid payload_txo")?,
             )
             .to_string(),
         ))
@@ -1980,11 +1980,9 @@ mod tests {
         let num_loops = 10;
 
         for loop_number in 1..=num_loops {
-
             let mut output_vec: Vec<OutputTxo> = Vec::new();
 
             for _ in 1..=loop_number {
-
                 let subaddress_index = 0;
                 let (tx_out, _) = create_test_txo_for_recipient(
                     &recipient_account_key,
@@ -2017,7 +2015,6 @@ mod tests {
 
     #[test]
     fn test_try_from_empty_vec_output_txo_for_transaction_id() {
-
         use std::assert_matches::assert_matches;
 
         let transaction_log_id = TransactionId::try_from(vec![]);
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -51,10 +51,9 @@ impl TryFrom<Vec<OutputTxo>> for TransactionId {
             HexFmt(
                 _payload_txos
                     .iter()
-                    .min_by_key(|txo| txo.tx_out.public_key)
+                    .map(|txo| txo.tx_out.public_key)
+                    .min()
                     .ok_or("no valid payload_txo")?
-                    .tx_out
-                    .public_key,
             )
             .to_string(),
         ))
@@ -2008,10 +2007,9 @@ mod tests {
 
             let min_public_key = output_vec
                 .iter()
-                .min_by_key(|txo| txo.tx_out.public_key)
-                .unwrap()
-                .tx_out
-                .public_key;
+                .map(|txo| txo.tx_out.public_key)
+                .min()
+                .unwrap();
             let transaction_log_id = TransactionId::try_from(output_vec).unwrap();
 
             assert_eq!(min_public_key.to_string(), transaction_log_id.0);
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -848,6 +848,7 @@ impl TransactionLogModel for TransactionLog {
 #[cfg(test)]
 mod tests {
     use std::{
+        assert_matches::assert_matches,
         collections::HashMap,
         ops::DerefMut,
         sync::{Arc, Mutex},
@@ -2015,8 +2016,6 @@ mod tests {
 
     #[test]
     fn test_try_from_empty_vec_output_txo_for_transaction_id() {
-        use std::assert_matches::assert_matches;
-
         let transaction_log_id = TransactionId::try_from(vec![]);
 
         assert_matches!(transaction_log_id, Err("no valid payload_txo"));
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -1972,7 +1972,7 @@ mod tests {
     }
 
     #[test]
-    fn try_from_vec_output_txo_for_transaction_id() {
+    fn test_try_from_vec_output_txo_for_transaction_id() {
 
         let mut rng: StdRng = SeedableRng::from_entropy();
         let root_id = RootIdentity::from_random(&mut rng);
@@ -2016,7 +2016,6 @@ mod tests {
 
             assert_eq!(min_public_key.to_string(), transaction_log_id.0);
         }
-        assert_eq!(num_txos, num_loops );
     }
 }
 
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -854,10 +854,12 @@ mod tests {
         sync::{Arc, Mutex},
     };
 
-    use mc_account_keys::{PublicAddress, CHANGE_SUBADDRESS_INDEX};
+    use mc_account_keys::{AccountKey, PublicAddress, RootIdentity, CHANGE_SUBADDRESS_INDEX};
     use mc_common::logger::{async_test_with_logger, Logger};
     use mc_ledger_db::Ledger;
+    use mc_util_from_random::FromRandom;
     use mc_transaction_core::{ring_signature::KeyImage, tokens::Mob, tx::Tx, Token};
+    use mc_transaction_extra::TxOutConfirmationNumber;
     use rand::{rngs::StdRng, SeedableRng};
 
     use crate::{
@@ -867,8 +869,8 @@ mod tests {
             transaction_builder::WalletTransactionBuilder,
         },
         test_utils::{
-            add_block_with_tx_outs, builder_for_random_recipient, get_resolver_factory,
-            get_test_ledger, manually_sync_account, random_account_with_seed_values,
+            add_block_with_tx_outs, builder_for_random_recipient, create_test_txo_for_recipient,
+            get_resolver_factory, get_test_ledger, manually_sync_account, random_account_with_seed_values,
             WalletDbTestContext, MOB,
         },
         util::b58::b58_encode_public_address,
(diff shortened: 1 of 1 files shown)


full-service/src/db/transaction_log.rs
```diff
@@ -1968,4 +1970,63 @@ mod tests {
...
+        let num_loops = 5; // number of times to run test
+        let amount = 77;
+
+        for _ in 1..=num_loops {
+            // run test loop
+
+            let mut output_vec: Vec<OutputTxo> = Vec::new();
+
+            for _ in 0..num_txos {
+                // loop to build vector and test try_from()
```
nick-mobilecoin, line 1990: ⛏️ I lean pretty hard in the camp of comments should communicate why and code should communicate what
These comments, got formatted funky due to cargo fmt, but also I don't think they provide much benefit
```suggestion
        let txo_vector_size = 5;
        let num_test_runs = 5;
        let amount = 77;

        for _ in 1..=num_test_runs {
            let mut output_vec: Vec<OutputTxo> = Vec::new();

            for _ in 0..txo_vector_size {
```

full-service/src/db/transaction_log.rs
```diff
@@ -1968,4 +1970,63 @@ mod tests {
...
+
+                let (tx_out, _) = create_test_txo_for_recipient(
+                    &recipient_account_key,
+                    0, // subaddress_index
+                    Amount::new(amount * MOB, Mob::ID),
+                    &mut rng,
+                );
```
nick-mobilecoin, line 1997: ⛏️ consider alternative to comment
I don't want to say "use X IDE", at the same time I think it's important to consider those that may use X IDE
image
Often times IDEs and language servers will give argument inlay hints for constants. commenting these constants results in noise for those developers.
Probably not too likely but still possible is a refactor could rename this argument and then the comment is no longer correct.

If one wants to help clarify what the 0 is consider:
```suggestion
                let subaddress_index = 0;
                let (tx_out, _) = create_test_txo_for_recipient(
                    &recipient_account_key,
                    subaddress_index,
                    Amount::new(amount * MOB, Mob::ID),
                    &mut rng,
                );
```

full-service/src/db/transaction_log.rs
```diff
@@ -1968,4 +1970,63 @@ mod tests {
...
+
+                output_vec.push(output_txo);
+            }
+
+            let min_public_key = output_vec
+                .iter()
+                .min_by_key(|txo| txo.tx_out.public_key)
+                .unwrap()
+                .tx_out
+                .public_key;
```
nick-mobilecoin, line 2015: ⛏️ 
```suggestion
            let min_public_key = output_vec
                .iter()
                .map(|txo| txo.tx_out.public_key)
                .min()
                .unwrap();
```

full-service/src/db/transaction_log.rs
```diff
@@ -1968,4 +1970,63 @@ mod tests {
...
+            let transaction_log_id = TransactionId::try_from(output_vec).unwrap();
+
+            assert_eq!(min_public_key.to_string(), transaction_log_id.0);
```
nick-mobilecoin, line 2018: ⛏️ I see that there is a `TransactionLog`, but it doesn't appear to be used in this test so perhaps the `_log` should be dropped
```suggestion
            let transaction_id = TransactionId::try_from(output_vec).unwrap();

            assert_eq!(min_public_key.to_string(), transaction_id.0);
```

full-service/src/db/transaction_log.rs
```diff
@@ -1968,4 +1970,63 @@ mod tests {
...
+        let num_txos = 5; // txo vector size
+        let num_loops = 5; // number of times to run test
+        let amount = 77;
+
+        for _ in 1..=num_loops {
+            // run test loop
+
+            let mut output_vec: Vec<OutputTxo> = Vec::new();
+
+            for _ in 0..num_txos {
```
nick-mobilecoin, line 1989: 🔧 I'm not fully understanding the purpose for `1..=num_loops` I don't think it gains much.
Ideally one should have test cases for  zero, one, multiple, the "zom" of [zombies](https://medium.com/@simontoth/daily-bit-e-of-c-testing-zombies-504238a18ef9).

consider:
```suggestion
        for loop_number in 1..=num_loops {

            let mut output_vec: Vec<OutputTxo> = Vec::new();

            for _ in 1..=loop_number {
```
this would get the one and many scenarios, with a few "many" scenarios
but I "personally" would probably lean toward parameterized tests 
The advantage of parameterized tests is each one runs independently so you could get more info about whats failing. 
With a loop the first failure stops the test, and thus all subsequent loop iterations
```rust
    #[parameterized(
        one = {1},
        two = {2},
        three = {3},
        ten = {10}
    )]
    fn test_try_from_vec_output_txo_for_transaction_id(num_txos: usize) {
        let mut rng: StdRng = SeedableRng::from_entropy();
        let root_id = RootIdentity::from_random(&mut rng);
        let recipient_account_key = AccountKey::from(&root_id);
        let amount = 77;
        let mut output_vec: Vec<OutputTxo> = Vec::new();

        for _ in 0..num_txos {
```

nick-mobilecoin (approved):
not sure what you all were hoping for
most of the comments are nits, ⛏️ , don't take them personal and feel free to ignore
the one thing I question is the looping inside of the test, 🔧

full-service/src/db/transaction_log.rs
```diff
@@ -1968,4 +1970,64 @@ mod tests {
...
+    #[test]
+    fn test_try_from_empty_vec_output_txo_for_transaction_id() {
+
+        let output_vec: Vec<OutputTxo> = Vec::new();
+        let transaction_log_id = TransactionId::try_from(output_vec);
+
+        match transaction_log_id {
+            Ok(_) => assert!(false, "Expected error, but got Ok"),
+            Err(err) => assert_eq!(err, "no valid payload_txo"),
+        }
```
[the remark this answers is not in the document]
lnsiegel: I did feel that that Err(err) assertion I had there was a weak point, and raised it in a separate discussion with Sara and Henry.  Last night Henry went deep on the best way implement the match statement or equivalent, and I followed along behind him.  Henry did significantly improve what I had originally.  But what you are showing there, and on a single line no less, outshines us both I think.

@holtzman : I expect to go with what Nick has, but before I resolve this convo, do you have any thoughts?

full-service/src/db/transaction_log.rs
```diff
@@ -1968,4 +1970,64 @@ mod tests {
...
+                .tx_out
+                .public_key;
+            let transaction_log_id = TransactionId::try_from(output_vec).unwrap();
+
+            assert_eq!(min_public_key.to_string(), transaction_log_id.0);
+        }
+    }
+
+    #[test]
+    fn test_try_from_empty_vec_output_txo_for_transaction_id() {
```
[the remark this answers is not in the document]
lnsiegel: Thanks!

full-service/src/db/transaction_log.rs
```diff
@@ -1968,4 +1970,64 @@ mod tests {
...
+
+        let output_vec: Vec<OutputTxo> = Vec::new();
+        let transaction_log_id = TransactionId::try_from(output_vec);
```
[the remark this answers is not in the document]
lnsiegel: Learnings!!  For now I will leave it as-is, but if I find myself back in the code I will certainly switch to this.

full-service/src/db/transaction_log.rs
lnsiegel: Eh, I don't think it was the cargo formatter.  I just wasn't being very disciplined with my comments, so _your_ comments here are well taken.

full-service/src/db/transaction_log.rs
lnsiegel: Agreed, the comment is unnecessary, and the descriptive variable name an improvement.

full-service/src/db/transaction_log.rs
lnsiegel: LGTM!

full-service/src/db/transaction_log.rs
lnsiegel: Makes sense.

full-service/src/db/transaction_log.rs
lnsiegel: Because the test txos are being initialized from entropy, I wanted to run the test more than once.   The `1..=num_loops` was just a simple way to do that.  But I see what you are showing is much more powerful.  I took a quick look at **yare** just now, it seems pretty straightforward.  And the 🧟‍♂️ mnemonic is good too!

full-service/src/db/transaction_log.rs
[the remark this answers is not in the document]
holtzman: looks great to me.

full-service/src/db/transaction_log.rs
holtzman: Let's change the corresponding block in`TransactionId::try_from` as well.

holtzman: lines 54-57 could use a similar change

lnsiegel: Got it, thanks Henry.

full-service/src/db/transaction_log.rs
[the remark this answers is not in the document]
holtzman: tried out assert_matches! and found that it requires
`#![feature(assert_matches)]`

to be added to the top of the file, as it is considered an `unstable feature`. There seems to be some debate about its semantics.  Further reading here:

https://github.com/rust-lang/rust/issues/82775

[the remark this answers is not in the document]
lnsiegel: I was able to get assert_matches to work by adding a `use` statement into the body of the test function:

`use std::assert_matches::assert_matches;`

At the moment I'm not clear if that's resolves to the same version as Henry gets with the `feature` attribute.  Would the version in the `std` crate be considered stable?  If not, should we avoid using this?

full-service/src/db/transaction_log.rs
lnsiegel: I tried to go the `yare` route on this one.  I was able to get the crate loaded, apply the code above, and build cleanly.  But I couldn't get the test to run after that.  I had the name of the test on the command line, and as I'm writing this I'm wondering if maybe I need to either change the test name in some way, or just give no test name and see if it runs along with all the tests.

In any case, I'm falling back to the nested looping in the form Nick suggested prior to mentioning `yare` above.  That seems to work quite well, even without being parameterized.

full-service/src/db/transaction_log.rs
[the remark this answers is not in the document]
nick-mobilecoin: > tried out assert_matches! and found that it requires
#![feature(assert_matches)] to be added to the top of the file

I believe this should only be required, once, at the crate level in `foo/src/lib.rs` or `foo/src/main.rs`

I've also been using this create at times, https://docs.rs/assert_matches/latest/assert_matches/

I can't decide if the feature gating is a good thing or a bad thing. For instance we have one project where the use of feature gating is the only reason the project is using nightly. The use of nightly is forcing that project to stay back on some dependencies, because those dependencies use nightly for some of their beta work and it causes compilation errors in our usage.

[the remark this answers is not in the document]
lnsiegel: Thanks Nick.  I'm going to look at moving the feature into `into full-service/src/lib.rs`.  But I'm still unclear as to whether it is safe to go ahead with this.  I will defer to you and to @holtzman.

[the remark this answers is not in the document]
lnsiegel: So it turns out that `#![feature(assert_matches)]` is already in `full-service/src/lib.rs`.  But if I don't declare it somewhere in `transaction_log.rs` then rustc tells me it can't find it.  

Still looking at that right now.  If you want to replicate this, you can comment out the `use` statement I added on line `2018` of `transaction_log.rs` and recompile.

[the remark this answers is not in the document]
nick-mobilecoin: I think there maybe some confusion with the `use` versus `#![feature]` nomenclature.

The `#![feature(assert_matches)]` tells the compiler that you know `assert_matches` is unstable and that you still want to work with it. This should only need to be done at the crate level and it's already done for full-service.

the `use std::assert_matches::assert_matches;` brings the `assert_matches` behavior into scope so that it can be used in that test. This is namespacing in case you have a local function or macro called `assert_matches`. The `use` statement needs to be made in every module that one is going to use the `assert_matches!()` macro.
I'd suggest putting this use statement at the top of the test module
```rust
#[cfg(test)]
mod tests {
    use itertools::Itertools;
    use mc_account_keys::{
        AccountKey, PublicAddress, RootIdentity, ShortAddressHash, CHANGE_SUBADDRESS_INDEX,
    };
    ...
    use std::{assert_matches::assert_matches, iter::FromIterator, ops::DerefMut, time::Duration};
```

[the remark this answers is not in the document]
lnsiegel: Ah ok.  I did go through that in rust book, but it's clearer now that I'm working through an actual use case.

I happened to be on with Sara and we took a look at the PR together, she let me know that it's not unusual for unstable features to be used.  So I'm going to go ahead and merge.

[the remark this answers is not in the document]
lnsiegel: Sorry, update the placement of the `use` statement and then merge.

PR #1008, opened 2024-09-19
Fix list_spendable() to not return inflight txos

nick-mobilecoin:
Previously `list_spendable()` would return back an inflight txo if it
previously existed in a transaction log that failed. Now
`list_spendable()` correctly excludes txos that are in flight.

### DIFF
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -2,6 +2,21 @@
 
 //! DB impl for the Txo model.
 
+use crate::{
+    db::{
+        account::{AccountID, AccountModel},
+        assigned_subaddress::AssignedSubaddressModel,
+        models::{
+            Account, AssignedSubaddress, AuthenticatedSenderMemo as AuthenticatedSenderMemoModel,
+            DestinationMemo as DestinationMemoModel, NewAuthenticatedSenderMemo,
+            NewDestinationMemo, NewTransactionOutputTxo, NewTxo, TransactionOutputTxo, Txo,
+        },
+        transaction_log::TransactionId,
+        Conn, WalletDbError,
+    },
+    service::models::tx_proposal::OutputTxo,
+    util::b58::b58_encode_public_address,
+};
 use diesel::{
     dsl::{count, exists, not},
     prelude::*,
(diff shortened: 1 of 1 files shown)


full-service/src/db/txo.rs
```diff
@@ -1799,14 +1799,21 @@ impl TxoModel for Txo {
...
-        query = query
-            .filter(transaction_logs::id.is_null())
-            .or_filter(transaction_logs::failed.eq(true))
-            .or_filter(
-                transaction_logs::id
+        let inflight_txo_ids = transaction_logs::table
+            .into_boxed()
+            .inner_join(transaction_input_txos::table)
+            .filter(
+                transaction_logs::submitted_block_index
```
holtzman, line 1806: currently, the build_transaction method is actually reserving its inputs because it gets logged when full-service does the signing, and only build_unsigned_transaction receives no logs.  I believe that this will change some client's to have their build_transaction no longer reserve the input txos as it is making sure that the transaction is submitted when previously that check did not happen.

Do we want to make this change as part of this PR?

nick-mobilecoin: Do we know what the intent of this was
```rust
            .or_filter(
                transaction_logs::id
                    .is_not_null()
                    .and(transaction_logs::submitted_block_index.is_null()),
```
That seems like it's looking for transactions logs with a non null id, but not submitted.

nick-mobilecoin: Maybe you're saying the same thing and suggesting further improvements.

I say we keep this PR focused on the current bug at hand. If we want to further improve we do in a follow on pr, even if we want it in this release.

Especially for bug fixes I think it's best to have the PR only contain the bug fix so it's easier to share, patch or revert as necessary

holtzman (approved):
except for the issue of changing the semantics of build_transaction to not reserving its input txos until the tombstone block, looks good!

approving in advance of resolving that discussion.

full-service/src/db/txo.rs
holtzman: agreed. was also trying to not change the semantics here, and think that I got lost in the cascading logic and thought we'd reversed things.

PR #1009, opened 2024-09-19
Omit finalized txo logs from list_spendable() query

nick-mobilecoin:


holtzman (approved):
this improvement on the calculation of `inflight_txo_ids` make sense to me.

PR #1011, opened 2024-09-20
Only fail transaction logs for specified account.

nick-mobilecoin:
Previously when syncing multiple accounts the first account synced would
fail expired transaction logs from other accounts. This would result in
failing transaction logs that may have actually succeeded. Now only
expired transaction logs for the currently processing account will be
failed.

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -352,6 +352,7 @@ pub trait TransactionLogModel {
     /// # Returns
     /// * unit
     fn update_pending_exceeding_tombstone_block_index_to_failed(
+        account_id: &AccountID,
         block_index: u64,
         conn: Conn,
     ) -> Result<(), WalletDbError>;
@@ -802,6 +803,7 @@ impl TransactionLogModel for TransactionLog {
     }
 
     fn update_pending_exceeding_tombstone_block_index_to_failed(
+        account_id: &AccountID,
         block_index: u64,
         conn: Conn,
     ) -> Result<(), WalletDbError> {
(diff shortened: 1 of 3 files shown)


full-service/src/db/transaction_log.rs
```diff
@@ -802,13 +803,15 @@ impl TransactionLogModel for TransactionLog {
...
     fn update_pending_exceeding_tombstone_block_index_to_failed(
+        account_id: &AccountID,
         block_index: u64,
         conn: Conn,
     ) -> Result<(), WalletDbError> {
         use crate::db::schema::transaction_logs;
 
         diesel::update(
             transaction_logs::table
+                .filter(transaction_logs::account_id.eq(&account_id.0))
```
nick-mobilecoin, line 814: one can comment out this line to see the new test fail

full-service/src/service/sync.rs
```diff
@@ -284,7 +285,8 @@ pub fn sync_account_next_chunk(
             }
 
             TransactionLog::update_pending_exceeding_tombstone_block_index_to_failed(
-                end_block_index + 1,
+                &account_id,
+                end_block_index,
```
holtzman, line 289: Why is this end_block_index here, on the view_only account clause, while below on the regular account (line 379) it is end_block_index + 1?

holtzman (approved):
lgtm

full-service/src/service/sync.rs
holtzman: I think + 1 is correct in both places.

holtzman: yeah. the view_only account code isn't covered by the unit test, and if one takes away the + 1 on the code that is covered, the unit test fails.

holtzman, line 289:
```suggestion
                end_block_index + 1,
```

nick-mobilecoin: good catch, not sure how I accidentally dropped that
Thanks!

PR #1012, opened 2024-10-08
ci/cd refactor: on-pr workflow

jgreat:
### Motivation

Refactoring ci/cd workflows.  First steps `on-pr` workflow.

- linters: rust, helm, docker, actions
- Run Tests 
- Build code for testnet/mainnet
  - MacOS (arm64, x64)
  - Linux (x64)
 - Build Linux containers and publish to dockerhub
 - Build helm charts and publish to harbor

Each build step caches results so compiles should only happen on rust/cargo changes.
Each Run will create a set of artifacts.  Zip files with binaries in the GH Actions interface and Container/Charts in their respective repos.

### Next steps
- `on-tag` workflow for publishing tagged binaries, containers and creating the github release.
- enhance builds with Linux ARM64 artifacts
- Integration testing for full-service against mainnet/testnet networks. 
- Clean up/standardize repo shell scripts so they will pass `shellcheck` linting.


sugargoat (approved):
Thank you!

PR #1016, opened 2024-11-05
make path in tar consistent, fix missing artifact in release

jgreat:
### Motivation

- looks like --strip-components isn't working on tarball creation in linux.  Just create desired directory structure instead. 
- fix typo in release upload and require all files to be present for successful release.


holtzman (approved):
LGTM

PR #1017, opened 2024-11-05
python cli status cmd should only display fees for tokens on chain

holtzman:
### Motivation

python CLI used on mainnet was getting a fatal error when used with the `status` command due to trying to display fees for token 8192, which is testnet only.
### In this PR
* when iterating over the defined tokens in order to display their minimum fees, only try and display fees for tokens that are present in the fee map.

### DIFF
--- a/python/mobilecoin/cli.py
+++ b/python/mobilecoin/cli.py
@@ -246,9 +246,10 @@ class CommandLineInterface:
         print()
         print('Transaction Fees:')
         for token in TOKENS:
-            if str(token.token_id) in network_status['fees']:
+            fee_storage_units = network_status['fees'].get(str(token.token_id))
+            if fee_storage_units is not None:
                 amount = Amount.from_storage_units(
-                    network_status['fees'][str(token.token_id)],
+                    fee_storage_units,
                     token
                 )
                 print(indent(amount.format(), '  '))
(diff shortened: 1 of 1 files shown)


python/mobilecoin/cli.py
```diff
@@ -246,11 +246,12 @@ def status(self):
...
-                network_status['fees'][str(token.token_id)],
-                token
-            )
-            print(indent(amount.format(), '  '))
+            if str(token.token_id) in network_status['fees']:
+                amount = Amount.from_storage_units(
+                    network_status['fees'][str(token.token_id)],
+                    token
+                )
+                print(indent(amount.format(), '  '))
```
nick-mobilecoin, line 254: The change is good enough as is

python does not do *any* optimizations to code logic
this means it will:
- `str(token.token_id)` twice
- likely look for the string in `network_status['fees']` twice

I doubt this code path is a performance bottleneck to need the change, just an FYI if one isn't aware
```suggestion
            token_fee = network_status['fees'].get(str(token.token_id))
            if token_fee is not None:
                amount = Amount.from_storage_units(
                    token_fee,
                    token
                )
                print(indent(amount.format(), '  '))
```

nick-mobilecoin (approved):

PR #1025, opened 2024-12-01
Various mob CLI improvements.

christian-oudard:


holtzman:
@christian-oudard you already proposed the first commit in #1013, where it was commented on and then fixed in #1018 for all full-service clients, with the original PR closed.

Please remove that change from this PR.

Thanks!

PR #1026, opened 2024-12-02
Consolidate logic for `sync_account_next_chunk()`

nick-mobilecoin:
Previously over 90 lines was duplicated in `sync_account_next_chunk()`
between regular accounts and view only accounts. This logic has now been
consolidated such that only the minor differences have been
consolidated.

### DIFF
--- a/full-service/src/service/sync.rs
+++ b/full-service/src/service/sync.rs
@@ -222,172 +222,93 @@ pub fn sync_account_next_chunk(
         }
         let end_block_index = end_block_index.unwrap();
 
-        let num_received_txos = if account.view_only {
+        let (view_private_key, account_key) = if account.view_only {
             let view_account_key: ViewAccountKey = mc_util_serial::decode(&account.account_key)?;
-
-            // Attempt to decode each transaction as received by this account.
-            let received_txos: Vec<_> = tx_outs
-                .into_par_iter()
-                .filter_map(|(block_index, tx_out)| {
-                    let amount = decode_amount(&tx_out, view_account_key.view_private_key())?;
-                    Some((block_index, tx_out, amount))
-                })
-                .collect();
-
-            let mut received_txos_with_subaddresses = Vec::new();
-            for (block_index, tx_out, amount) in received_txos {
-                let subaddress_index =
-                    decode_subaddress_index(&tx_out, view_account_key.view_private_key(), conn);
-
-                received_txos_with_subaddresses.push((
-                    block_index,
-                    tx_out,
-                    amount,
-                    subaddress_index,
-                ));
-            }
-
-            let num_received_txos = received_txos_with_subaddresses.len();
-
-            // Write received transactions to the database.
-            for (block_index, tx_out, amount, subaddress_index) in received_txos_with_subaddresses {
-                Txo::create_received(
-                    tx_out.clone(),
-                    subaddress_index,
-                    None,
-                    amount,
-                    block_index,
-                    account_id_hex,
-                    conn,
-                )?;
-            }
-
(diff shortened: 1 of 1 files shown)


full-service/src/service/sync.rs
```diff
@@ -222,172 +222,93 @@ pub fn sync_account_next_chunk(
         }
         let end_block_index = end_block_index.unwrap();
 
-        let num_received_txos = if account.view_only {
+        let (view_private_key, account_key) = if account.view_only {
```
nick-mobilecoin, line 225: whether or not we have access to a full `AccountKey` was the main difference between the two previously large blocks of text

full-service/src/service/sync.rs
```diff
@@ -222,172 +222,93 @@ pub fn sync_account_next_chunk(
...
+            );
+
+            received_txos_with_subaddresses_and_key_images.push((
+                block_index,
+                tx_out,
+                amount,
+                subaddress_index,
+                key_image,
+            ));
+        }
```
nick-mobilecoin, line 258: it's unfortunate that github's diff logic started spreading out here as this is more or less the same
doing a diff locally in something like beyondcompare might be easier to visualize
image

full-service/src/service/sync.rs
```diff
@@ -446,28 +365,30 @@ pub fn decode_subaddress_index(
...
-            account_key.view_private_key(),
-            &account_key.subaddress_spend_private(subaddress_i),
-        );
-        Some(KeyImage::from(&onetime_private_key))
-    } else {
-        None
-    };
+    let subaddress_index = decode_subaddress_index(tx_out, view_private_key, conn);
+
+    let key_image = account_key.and_then(|account_key| {
```
nick-mobilecoin, line 378: TBH i'm not fond of this logic and extension to this function, but it seemed like the minimal change to reduce the excessive code duplication in `sync_account_next_chunk()(`

full-service/src/service/sync.rs
```diff
@@ -222,172 +222,93 @@ pub fn sync_account_next_chunk(
...
+                    .map(|txo_id_hex| (block_index, txo_id_hex.clone()))
+            })
+            .collect();
+
+        for (block_index, txo_id_hex) in &spent_txos {
+            Txo::update_spent_block_index(txo_id_hex, *block_index, conn)?;
+            TransactionLog::update_pending_associated_with_txo_to_succeeded(
+                txo_id_hex,
+                *block_index,
                 conn,
```
nick-mobilecoin, line 294: The logic and arguments in `update_pending_associated_with_txo_to_succeeded()` will need to be updated to address the issue of multiple transaction logs with the same `txo_id_hex` being marked as succeeded when only one should. 
instead of trying to mirror the change between the two different code branches this function had, I thought it better to consolidate first before applying the fix

full-service/src/service/sync.rs
holtzman: I can't get the above image to display as more than a thumbnail.  Am I missing something good? 😅

full-service/src/service/sync.rs
holtzman: thanks for taking this on. The places where view account logic was largely duplicative was a bit of tech debt that is wonderful to be addressing.

holtzman (approved):
lgtm

full-service/src/service/sync.rs
nick-mobilecoin: sorry that's the sidebar from BeyondCompare, unfurtunatly the changes were so long that I couldn't get a nice screenshot of it

PR #1027, opened 2024-12-03
Fix transaction logs incorrectly marked successful

nick-mobilecoin:
Previously `transaction_log`s were marked successful if the input TXO(s)
were spent according to the block chain. This was problematic when
multiple `transaction_log`s were created that utilized the same input
TXOs. These `transaction_log`s using the same input TXOs could happen
when two clients were talking to the same Full-Service instance and
those clients were using two step build and submit:

1. Client_a sends `build_transaction`, input TXO_n is used
2. Client_b sends `build_transaction`, input TXO_n is also used
3. Client_a `submit_transaction`
4. Client_b `submit_transaction` before the block chain knows that the
   Client_a transaction exists and will consume `TXO_n`. The blockchain
   will still reject this transaction eventually.
5. Previously, Full-Service would see that input TXO_n has been consumed
   and will mark **all** submitted `transaction_log`s that used TXO_n as
   successful.

Now `transaction_log`s are marked successful based on if their output
TXO(s) land on the blockchain. This ensures that only the transaction
log that the blockchain accepted will be marked as successful.

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -777,16 +777,16 @@ impl TransactionLogModel for TransactionLog {
     }
 
     fn update_pending_associated_with_txo_to_succeeded(
-        txo_id_hex: &str,
+        txo_out_id_hex: &str,
         finalized_block_index: u64,
         conn: Conn,
     ) -> Result<(), WalletDbError> {
-        use crate::db::schema::{transaction_input_txos, transaction_logs};
+        use crate::db::schema::{transaction_logs, transaction_output_txos};
         // Find all submitted transaction logs associated with this txo that have not
         // yet been finalized (there should only ever be one).
         let transaction_log_ids: Vec<String> = transaction_logs::table
-            .inner_join(transaction_input_txos::table)
-            .filter(transaction_input_txos::txo_id.eq(txo_id_hex))
+            .inner_join(transaction_output_txos::table)
+            .filter(transaction_output_txos::txo_id.eq(txo_out_id_hex))
             .filter(transaction_logs::submitted_block_index.is_not_null()) // we actually sent this transaction
             .filter(transaction_logs::failed.eq(false)) // non-failed transactions
             .filter(transaction_logs::finalized_block_index.is_null()) // non-completed transactions
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -777,7 +777,7 @@ impl TransactionLogModel for TransactionLog {
     }
 
     fn update_pending_associated_with_txo_to_succeeded(
-        txo_out_id_hex: &str,
+        transaction_output_txo_id_hex: &str,
         finalized_block_index: u64,
         conn: Conn,
     ) -> Result<(), WalletDbError> {
@@ -786,7 +786,7 @@ impl TransactionLogModel for TransactionLog {
         // yet been finalized (there should only ever be one).
         let transaction_log_ids: Vec<String> = transaction_logs::table
             .inner_join(transaction_output_txos::table)
-            .filter(transaction_output_txos::txo_id.eq(txo_out_id_hex))
+            .filter(transaction_output_txos::txo_id.eq(transaction_output_txo_id_hex))
             .filter(transaction_logs::submitted_block_index.is_not_null()) // we actually sent this transaction
             .filter(transaction_logs::failed.eq(false)) // non-failed transactions
             .filter(transaction_logs::finalized_block_index.is_null()) // non-completed transactions
--- a/full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_then_submit.rs
+++ b/full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_then_submit.rs
@@ -9,7 +9,7 @@ mod e2e_transaction {
         json_rpc::v2::{
             api::test_utils::{dispatch, dispatch_expect_error, setup},
             models::{
-                amount::Amount as AmountJSON, transaction_log::TransactionLog,
+                account::Account, amount::Amount as AmountJSON, transaction_log::TransactionLog,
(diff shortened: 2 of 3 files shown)


full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_then_submit.rs
```diff
@@ -717,10 +717,18 @@ mod e2e_transaction {
 
     #[test_with_logger]
     fn test_reuse_input_txo_then_submit_transaction(logger: Logger) {
+        // This test mimics a situation where multiple services are using the same
```
nick-mobilecoin, line 720: I decided to modify the test that was added as part of #738 since I think that the new code logic better captures the intended behavior and it seems like it's very similar if not the same problem that #738 was trying to solve.

I'm a big proponent of tests, but I'm also a big proponent of, just enough tests.

full-service/src/db/transaction_log.rs
```diff
@@ -777,16 +777,16 @@ impl TransactionLogModel for TransactionLog {
     }
 
     fn update_pending_associated_with_txo_to_succeeded(
-        txo_id_hex: &str,
+        txo_out_id_hex: &str,
```
nick-mobilecoin, line 780: 🤷 not sure on the addition of `out` here. It's nicer to have it in the signature than to look at the implementation to see the `transaction_output_txos`

full-service/src/service/sync.rs
```diff
@@ -230,6 +230,17 @@ pub fn sync_account_next_chunk(
             (*account_key.view_private_key(), Some(account_key))
         };
 
+        tx_outs.iter().try_for_each(|(block_index, tx_out)| {
```
nick-mobilecoin, line 233: Moved the logic for marking transaction logs as succeeded so that the `txo_outs` didn't need to be cloned, as they are consumed by the `into_par_iter()` below.

I don't think there is any intentional rhyme or reason on the order of operations, in particular, with regard to potential errors that may occur.

full-service/src/json_rpc/v2/e2e_tests/transaction/build_submit/build_then_submit.rs
```diff
@@ -749,126 +755,102 @@ mod e2e_transaction {
...
-        let transaction_log: TransactionLog =
-            serde_json::from_value(result.get("transaction_log").unwrap().clone()).unwrap();
-        assert_eq!(transaction_log.id, txlog_id_b);
-        assert_eq!(transaction_log.status, TxStatus::Built.to_string());
+        // The first transaction log will have succeeded, because it's outputs
+        // show up on the ledger.
+        // The second transaction log will still be pending, because it's outputs
+        // don't exist in the ledger. This second transaction log will
+        // eventually fail once the tombstone block is reached.
+        let expected_statuses = [TxStatus::Succeeded, TxStatus::Pending];
```
nick-mobilecoin, line 833: If one reverts the code changes in `full-service/src/db/transaction_log.rs` and `full-service/src/service/sync.rs` this test will fail with both status coming back as `Succeeded`

full-service/src/db/transaction_log.rs
holtzman: Given the reason for the PR, it makes sense to be more clear that this function should only be used for a `transaction_output` `txo`.

What do you think of keeping consistent with that notation, which is used already used -- prefixing with `transaction_output` -- i.e. `transaction_output_txo_id_hex`? 

I don't have a strong opinion here, but I prefer avoiding making new terminology when there is an existing pattern, and especially putting words in the middle of existing patterns, such as `txo_id` and `txo_id_hex`.

nick-mobilecoin: The main problem I see is it comes across as redundant
`txo` stands for transaction output so to have `transaction_output_txo_id_hex` is communicating `transaction_output_transaction_output_id_hex`

Though the table is named `transaction_output_txos`... so it's probably not inconsistent, i'll go ahead and make the change

nick-mobilecoin: done in [ebd3a46](https://github.com/mobilecoinofficial/full-service/pull/1027/commits/ebd3a4617a647ebcec4c1d655e4d80205179fc7e)

full-service/src/db/transaction_log.rs
```diff
@@ -777,16 +777,16 @@ impl TransactionLogModel for TransactionLog {
...
-        use crate::db::schema::{transaction_input_txos, transaction_logs};
+        use crate::db::schema::{transaction_logs, transaction_output_txos};
         // Find all submitted transaction logs associated with this txo that have not
         // yet been finalized (there should only ever be one).
         let transaction_log_ids: Vec<String> = transaction_logs::table
-            .inner_join(transaction_input_txos::table)
-            .filter(transaction_input_txos::txo_id.eq(txo_id_hex))
+            .inner_join(transaction_output_txos::table)
+            .filter(transaction_output_txos::txo_id.eq(txo_out_id_hex))
             .filter(transaction_logs::submitted_block_index.is_not_null()) // we actually sent this transaction
```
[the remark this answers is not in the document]
holtzman: I think making the minimal change is the correct, conservative approach.

The finalized_block_index check might also be unnecessary at this point.

However, if we go stripping those things out and it turns out that there are ways to get multiple transaction_log entries that have a common output txo, this may prevent issues for any "shadow" log entries that are not the actual submitted transaction.

There should only be 1 transaction_log entry that matches, and so perhaps a check for that and some sort of log message would be the change for the next PR on the topic?

holtzman (approved):
Looks good!

I think failing a tranasction_log where the input is spent and the output is not present in the same block would be good for the case where 2x full-service with the same account keys are running together, which is a situation that inspired this PR.

That said, this PR fixes a bug whereas that would be an improvement, allowing the full-service with the failed tx to recognize it sooner, so could be the subject of a future PR.

PR #1028, opened 2024-12-03
Fail transaction logs when input txos are consumed

nick-mobilecoin:
Previously a transaction log would need to wait for the tombstone block
to occur. Now when an input txo is spent any transaction logs that
haven't been finalized and use that input txo will be marked failed.

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -15,6 +15,11 @@ use crate::{
             Account, NewTransactionInputTxo, NewTransactionLog, TransactionInputTxo,
             TransactionLog, TransactionOutputTxo, Txo,
         },
+        schema::{
+            transaction_input_txos, transaction_logs,
+            transaction_logs::dsl::{id as dsl_id, transaction_logs as dsl_transaction_logs},
+            transaction_output_txos, txos,
+        },
         txo::{TxoID, TxoModel},
         Conn, WalletDbError,
     },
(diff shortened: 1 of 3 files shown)


full-service/src/db/transaction_log.rs
```diff
@@ -322,24 +327,38 @@ pub trait TransactionLogModel {
...
-    /// # Returns
-    /// * unit
+    /// * `transaction_output_txo_id_hex` - The txo ID for which to get all transaction logs
+    ///   associated with this txo ID.
+    /// * `finalized_block_index` - The block index at which the transaction will be completed and
+    ///   finalized.
+    /// * `conn` - A reference to the pool connection of wallet database
     fn update_pending_associated_with_txo_to_succeeded(
-        txo_id_hex: &str,
+        transaction_output_txo_id_hex: &str,
```
nick-mobilecoin, line 340: noticed when I went to add the new method, I failed to update the trait docs for the previous PR.

full-service/src/db/transaction_log.rs
```diff
@@ -322,24 +327,38 @@ pub trait TransactionLogModel {
...
-    ///
-    ///| Name                    | Purpose                                                                      | Notes |
-    ///|-------------------------|------------------------------------------------------------------------------|-------|
-    ///| `txo_id_hex`            | The txo ID for which to get all transaction logs associated with this txo ID.|       |
-    ///| `finalized_block_index` | The block index at which the transaction will be completed and finalized.    |       |
-    ///| `conn`                  | An reference to the pool connection of wallet database                       |       |
-    ///
-    /// # Returns
-    /// * unit
+    /// * `transaction_output_txo_id_hex` - The txo ID for which to get all transaction logs
```
nick-mobilecoin, line 334: I'm not sure the history on the doc format, if perhaps it was an early Cargo docs version or something, but the bullet + dash style is what's current
I mainly changed this as I was going to have to update the table format due to the argument rename, one of the reasons aligned table formats are avoided for these kind of docs.

full-service/src/db/transaction_log.rs
```diff
@@ -407,9 +426,7 @@ impl TransactionLogModel for TransactionLog {
     }
 
     fn get(id: &TransactionId, conn: Conn) -> Result<TransactionLog, WalletDbError> {
-        use crate::db::schema::transaction_logs::dsl::{id as dsl_id, transaction_logs};
```
nick-mobilecoin, line 410: I'm not sure if there was a reason the use statements were in each function, it seems to cause a bit of code duplication and detracts from focusing on the function implementations.

full-service/src/db/transaction_log.rs
```diff
@@ -802,13 +803,32 @@ impl TransactionLogModel for TransactionLog {
         Ok(())
     }
 
+    fn update_consumed_txo_to_failed(
+        transaction_input_txo_id_hex: &str,
+        conn: Conn,
+    ) -> Result<(), WalletDbError> {
+        let transaction_log_ids: Vec<String> = transaction_logs::table
+            .inner_join(transaction_input_txos::table)
+            .filter(transaction_input_txos::txo_id.eq(transaction_input_txo_id_hex))
```
holtzman, line 812: should there be a filter here to make sure that the transaction_log was submitted so that logs for unsubmitted transactions don't become confused with ones that were submitted and failed?

Or is this only reached for submitted transaction_logs already?

nick-mobilecoin: since this is coming in via input txos, not transaction logs, this will grab any and all transaction logs that have the input txo
¿Do we have a use case where one would build a transaction log and want to see it in the built state even though it can never succeed?

In particular when we saw the spent key image error, it was because 
1. the transaction log was built.
2. A different transaction log with the same input TXOs landed on the block chain
3. the first transaction log was submitted, because it was showing up as `built` not `failed`
4. The consensus nodes kept rejecting the transaction log due to key images already spent

holtzman (approved):
lgtm

full-service/src/db/transaction_log.rs
holtzman: The use case that I'm thinking through is when a client, e.g. Desktop Wallet, does a build_transaction prior to asking the user to confirm the transaction, and then the user declines to confirm.

This will leave a transaction_log with a submitted_block_index of null

I *think* for error reporting in the client, it would be good for the client to be able to tell the difference between unsubmitted transactions and failed transactions.

holtzman: do we already set these to failed when the tombstone block passes?

nick-mobilecoin: it looks like those transaction logs will be marked failed once the tombstone block is exceeded, [update_pending_exceeding_tombstone_block_index_to_failed](https://github.com/mobilecoinofficial/full-service/blob/cb2f46e3d6c03e9e06edc15bd98fa08bf9c0b90c/full-service/src/db/transaction_log.rs#L805)
```rust
        diesel::update(
            transaction_logs::table
                .filter(transaction_logs::account_id.eq(&account_id.0))
                .filter(transaction_logs::tombstone_block_index.lt(block_index as i64))
                .filter(transaction_logs::failed.eq(false))
                .filter(transaction_logs::finalized_block_index.is_null()),
        )
        .set((transaction_logs::failed.eq(true),))
        .execute(conn)?;
```

nick-mobilecoin: Which jives with the experience we had where we kept trying to submit the transaction, and the consensus node kept saying already spent
our service saw the transaction log in the build state, until the tombstone block was exceeded

PR #1033, opened 2025-01-08
Improved wallet sync performance

holtzman:
### Motivation

Since PR #1027, a performance issue has been noted when syncing a wallet-db that has a large number of `txos` tracked within. This is due to looking up the `txo_id` of each txo as it is synced from the ledger via the `txos` table using the `public_key` column, which is not indexed.

### In this PR
* when scanning the `transaction_logs` to see if a `txo` appearing in the ledger means a corresponding `transaction_log` entry should be transitioned from `pending` to `succeeded`, calculate the `txo_id` (which is a digest of the `txo` data that appeared in the ledger, rather than looking it up in the `txos` table by `public_key`. This significantly relieved the performance issue, lowering sync times by several orders of magnitude.
* to further increase sync throughput for imported (and re-syncing) accounts, a check was added to only do the above part of the sync when the account has pending transactions in `transaction_log`, and at least one of those transactions was submitted prior to the chunk of the ledger being synced was "mined."
* during debugging, it was also noted that a good deal of time was being spent in the part of the sync loop where unspent `txos` were being checked against incoming `key_images` to see if they should be marked as spent.  Further exploration indicated that an index of (`account_id`, `spent_block_index`) would greatly speed up the SQL query that returned all of the unspent `key_images` for an account, and this PR adds that index via a migration.

### DIFF
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -2336,12 +2336,18 @@ mod tests {
 
         assert_eq!(ledger_db.num_blocks().unwrap(), 18);
 
+        // make sure the tests were set up correctly
+        assert_ne!(
+            first_account_first_tx_log.submitted_block_index,
+            first_account_second_tx_log.submitted_block_index
+        );
         assert_ne!(
             first_account_first_tx_log.submitted_block_index,
             second_account_tx_log.submitted_block_index
         );
         assert_ne!(first_account_id.to_string(), second_account_id.to_string());
 
+        // chceck that the results are as expected prior to syncing
         assert_eq!(
             TransactionLog::lowest_pending_block_index(&first_account_id, conn).unwrap(),
             Some(first_account_first_tx_log.submitted_block_index.unwrap() as u64)
@@ -2351,9 +2357,11 @@ mod tests {
             Some(second_account_tx_log.submitted_block_index.unwrap() as u64)
         );
 
+        // sync the accounts
         let _sync = manually_sync_account(&ledger_db, &wallet_db, &first_account_id, &logger);
         let _sync = manually_sync_account(&ledger_db, &wallet_db, &second_account_id, &logger);
 
+        // check that the results are as expected after syncing
         assert_eq!(
             TransactionLog::lowest_pending_block_index(&first_account_id, conn).unwrap(),
             None
@@ -2362,7 +2370,6 @@ mod tests {
             TransactionLog::lowest_pending_block_index(&second_account_id, conn).unwrap(),
             None
         );
-
         let updated_tx_log =
             TransactionLog::get(&TransactionId::from(&first_account_first_tx_log), conn).unwrap();
         assert_eq!(updated_tx_log.status(), TxStatus::Succeeded);
--- a/full-service/src/db/transaction_log.rs
+++ b/full-service/src/db/transaction_log.rs
@@ -342,6 +342,22 @@ pub trait TransactionLogModel {
         conn: Conn,
     ) -> Result<(), WalletDbError>;
 
+    /// Return the lowest block_index associated with an account's pending transactions, as tracked in this wallet db.
+    /// 
+    /// # Arguments
+    ///
+    ///| Name                   | Purpose                                                 | Notes                               |
+    ///|------------------------|---------------------------------------------------------|-------------------------------------|
+    ///| `account_id`           | The account id to scan for transaction logs.            | Account must exist in the database. |
+    ///| `conn`                 | An reference to the pool connection of wallet database  |                                     |
+    /// 
+    /// # Returns
+    /// * None if no pending transactions from account_id, or
+    ///   Some<u64> that is the submitted_block_index of the oldest pending tx.
+    fn lowest_pending_block_index(
+        account_id: &AccountID,
+        conn: Conn,
+    ) -> Result<Option<u64>, WalletDbError>;
 
     /// Update all transaction logs that have an input transaction corresponding to
     /// `transaction_input_txo_id_hex` to failed.
@@ -776,6 +792,27 @@ impl TransactionLogModel for TransactionLog {
         .execute(conn)?;
 
         Ok(())
+    }
+
+    fn lowest_pending_block_index(
+        account_id: &AccountID,
+        conn: Conn,
+    ) -> Result<Option<u64>, WalletDbError> {
+        let lowest_pending_submitted_block_index: Vec<Option<i64>> = transaction_logs::table
+            .filter(transaction_logs::account_id.eq(&account_id.0))
+            .filter(transaction_logs::submitted_block_index.is_not_null())
+            .filter(transaction_logs::failed.eq(false)) // non-failed transactions
+            .filter(transaction_logs::finalized_block_index.is_null())
+            .select(diesel::dsl::min(transaction_logs::submitted_block_index))
+            .load(conn)?;
+
+        if lowest_pending_submitted_block_index.len() == 0
+            || lowest_pending_submitted_block_index[0].is_none()
+        {
+            Ok(None)
+        } else {
+            Ok(Some(lowest_pending_submitted_block_index[0].unwrap() as u64))
+        }
     }
 
     fn update_pending_associated_with_txo_to_succeeded(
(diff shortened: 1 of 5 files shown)


full-service/src/service/receipt.rs
```diff
@@ -591,7 +592,7 @@ mod tests {
         // Land the Txo in the ledger - only sync for the sender
         TransactionLog::log_submitted(
             &tx_proposal,
-            14,
```
holtzman, line 594: this tx actually lands in block_index 13 and the test conditions never create a block_index 14. Improvements to the ledger->wallet sync code cause this test to fail now due to the off-by-one error in the logging call.

The update to the test makes a real-time determination of which block_index to use.

There are two more identical updates below for the same reason.

full-service/src/db/transaction_log.rs
```diff
@@ -2123,4 +2160,226 @@ mod tests {
...
+            &mut ledger_db,
+            &[80 * MOB],
+            &mut rng,
+            &logger,
+        );
+        let second_account_id = AccountID::from(&second_account_key);
+
+        // Build a first transaction from first_account
+        let mut pooled_conn = wallet_db.get_pooled_conn().unwrap();
+        let conn = pooled_conn.deref_mut();
```
nick-mobilecoin, line 2193: consider this comment and the lines immediately following it.
The comment talks about building a transaction, but the next two lines deal with creating a database connection.
Then there is a blank line, which often communicates a change in thought  or logic

one will mentally group these 3 lines together as the same idea, _somewhat_ independent of the rest of the code
```suggestion
        let mut pooled_conn = wallet_db.get_pooled_conn().unwrap();
        let conn = pooled_conn.deref_mut();
```

full-service/src/db/transaction_log.rs
```diff
@@ -2123,4 +2160,226 @@ mod tests {
...
+            &mut ledger_db,
+            &[
+                tx_proposal.change_txos[0].tx_out.clone(),
+                tx_proposal.payload_txos[0].tx_out.clone(),
+            ],
+            &key_images,
+            &mut rng,
+        );
+
+        assert_eq!(ledger_db.num_blocks().unwrap(), 18);
```
nick-mobilecoin, line 2337: I think this can be DRYed up some
```suggestion
        let recipient_account_key = AccountKey::random(&mut rng);
        let recipient = recipient_account_key.subaddress(rng.next_u64());
        let mut logs = vec![];
        for (account_key, value) in [(&account_keys[0], 50), (&account_keys[0], 45), (&account_keys[1], 55)] {
            let (log, tx_proposal) = create_test_minted_and_change_txos(account_key.clone(), recipient.clone(), value * MOB, wallet_db.clone(), ledger_db.clone()).await;
            let key_images: Vec<KeyImage> = tx_proposal
                .input_txos
                .iter()
                .map(|txo| txo.key_image)
                .collect();

            // Note: This block doesn't contain the fee output.
            add_block_with_tx_outs(
                &mut ledger_db,
                &[
                    tx_proposal.change_txos[0].tx_out.clone(),
                    tx_proposal.payload_txos[0].tx_out.clone(),
                ],
                &key_images,
                &mut rng,
            );
            logs.push(log);
        }
        assert_eq!(ledger_db.num_blocks().unwrap(), 18);
```

full-service/src/db/transaction_log.rs
```diff
@@ -2123,4 +2160,226 @@ mod tests {
...
+            TransactionLog::get(&TransactionId::from(&first_account_first_tx_log), conn).unwrap();
+        assert_eq!(updated_tx_log.status(), TxStatus::Succeeded);
+
+        let updated_tx_log =
+            TransactionLog::get(&TransactionId::from(&first_account_second_tx_log), conn).unwrap();
+        assert_eq!(updated_tx_log.status(), TxStatus::Succeeded);
+
+        let updated_tx_log =
+            TransactionLog::get(&TransactionId::from(&second_account_tx_log), conn).unwrap();
+        assert_eq!(updated_tx_log.status(), TxStatus::Succeeded);
```
nick-mobilecoin, line 2383: IFF the change to use of arrays is done above
```suggestion
        // make sure the tests were set up correctly
        assert_ne!(
            logs[0].submitted_block_index,
            logs[1].submitted_block_index
        );
        assert_ne!(
            logs[0].submitted_block_index,
            logs[2].submitted_block_index
        );
        assert_ne!(account_ids[0].to_string(), account_ids[1].to_string());

        // check that the results are as expected prior to syncing
        assert_eq!(
            TransactionLog::lowest_pending_block_index(&account_ids[0], conn).unwrap(),
            Some(logs[0].submitted_block_index.unwrap() as u64)
        );
        assert_eq!(
            TransactionLog::lowest_pending_block_index(&account_ids[1], conn).unwrap(),
            Some(logs[2].submitted_block_index.unwrap() as u64)
        );

        for account_id in &account_ids {
            manually_sync_account(&ledger_db, &wallet_db, &account_id, &logger);
            assert_eq!(
                TransactionLog::lowest_pending_block_index(&account_id, conn).unwrap(),
                None
            );
        }

        for log in logs {
            let updated_log = TransactionLog::get(&TransactionId::from(&log), conn).unwrap();
            assert_eq!(updated_log.status(), TxStatus::Succeeded);
        }
```

full-service/src/service/sync.rs
```diff
@@ -230,17 +230,23 @@ pub fn sync_account_next_chunk(
...
 
-        tx_outs.iter().try_for_each(|(block_index, tx_out)| {
-            let txos = Txo::select_by_public_key(&[&tx_out.public_key], conn).unwrap();
-            txos.iter().try_for_each(|txo| {
+        // Mark pending txs as succeeded when one of their output txos is found on the
+        // chain.
+        let lowest_pending_block_index =
+            TransactionLog::lowest_pending_block_index(&account_id, conn)?;
+        if lowest_pending_block_index.is_some()
+            && lowest_pending_block_index.unwrap() <= end_block_index
```
nick-mobilecoin, line 238: My understanding is that the `lowest_pending_block_index` is an optimization intended to speed up re-syncing the full block chain for a client.  I'm not sure how much we've regressed performance wise compared to pre 2.10.5 and only adding the `TxoID::from()` logic.

I want to be sure this logic is sound and isn't prone to re-submission issues

1. Client calls [submit_transaction](https://mobilecoin.gitbook.io/full-service-api/api-endpoints/v2/transaction/transaction/submit_transaction) when full service sees the block 
2. Full service sends the request to a consensus node, the node says I'm at block 10, and the tx looks good so far
3. full service sets the submitted block based on 10
4. Client calls submit_transaction again
5. Full service sends the request to a consensus node, the node says I'm at block 11 and the txo looks good so far
6. full service will set the submitted block based on 11

My understanding is the consensus node would return an error in an instance where the initial submit transaction landed and the client returned a newer block index

nick-mobilecoin (approved):
looks good, just want to be sure about the logic of only updating logs that have a submitted index after the current range.

full-service/src/db/transaction_log.rs
holtzman: removed this code from the pr

full-service/src/db/transaction_log.rs
holtzman: removed this code from the PR

full-service/src/db/transaction_log.rs
holtzman: removed this code from the PR

full-service/src/db/transaction_log.rs
```diff
@@ -2123,4 +2160,226 @@ mod tests {
...
+        let first_account_id = AccountID::from(&first_account_key);
+
+        let second_account_key = random_account_with_seed_values(
+            &wallet_db,
+            &mut ledger_db,
+            &[80 * MOB],
+            &mut rng,
+            &logger,
+        );
+        let second_account_id = AccountID::from(&second_account_key);
```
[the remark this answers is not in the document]
holtzman: removed this code from the PR

full-service/src/db/transaction_log.rs
```diff
@@ -2123,4 +2160,226 @@ mod tests {
...
     }
+
+    #[async_test_with_logger]
+    async fn test_lowest_pending_block_index(logger: Logger) {
+        let mut rng: StdRng = SeedableRng::from_seed([20u8; 32]);
+
+        let db_test_context = WalletDbTestContext::default();
+        let wallet_db = db_test_context.get_db_instance(logger.clone());
+        let known_recipients: Vec<PublicAddress> = Vec::new();
+        let mut ledger_db = get_test_ledger(5, &known_recipients, 12, &mut rng);
```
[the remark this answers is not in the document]
holtzman: removed from PR -- test no longer needed due to removing the function being tested.

full-service/src/db/transaction_log.rs
```diff
@@ -778,6 +794,27 @@ impl TransactionLogModel for TransactionLog {
...
+            .filter(transaction_logs::finalized_block_index.is_null())
+            .select(diesel::dsl::min(transaction_logs::submitted_block_index))
+            .load(conn)?;
+
+        if lowest_pending_submitted_block_index.len() == 0
+            || lowest_pending_submitted_block_index[0].is_none()
+        {
+            Ok(None)
+        } else {
+            Ok(Some(lowest_pending_submitted_block_index[0].unwrap() as u64))
```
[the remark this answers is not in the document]
holtzman: removed

full-service/src/db/transaction_log.rs
```diff
@@ -778,6 +794,27 @@ impl TransactionLogModel for TransactionLog {
...
+        }
```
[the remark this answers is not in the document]
holtzman: removed from PR

eranrund (approved):
Nice work!

PR #1034, opened 2025-03-05
fix a bug that prevented change output fog hint from being properly created

eranrund:
### Motivation

This PR fixes a bug that prevents the fog hint from being properly created when adding change outputs.
Prior to this fix, the change output would not get a fog hint even for fog-enabled accounts.

### In this PR
* Fixes the aforementioned bug

The fix is not particularly elegant, see comments in code.

### Test Plan
Added unit tests that ensure correct fog hints are being generated for both the change address and recipients.
I opted to not add a unit test for ensuring no fog hint gets generated for a change address for non-fog accounts, since I don't think that's an interesting thing to test because in real life there's no public key available for encrypting the fog hint if the account does not have fog enabled. Adding that test would just be a lot of copy-paste and it doesn't seem worth the code bloat to me.

I also tested that this works by importing my Sentz app entropy and sending a transaction, and after this fix I can continue to see my balance in the Sentz app. Prior to this fix, my Sentz balance would show $0 once I sent a tx using `full-service`.

### DIFF
--- a/full-service/src/service/transaction_builder.rs
+++ b/full-service/src/service/transaction_builder.rs
@@ -671,11 +671,14 @@ mod tests {
         service::sync::SyncThread,
         test_utils::{
             builder_for_random_recipient, get_test_ledger, random_account_with_seed_values,
-            WalletDbTestContext, MOB,
+            random_fog_enabled_account_with_seed_values, WalletDbTestContext, MOB,
         },
     };
     use mc_account_keys::AccountKey;
     use mc_common::logger::{async_test_with_logger, test_with_logger, Logger};
+    use mc_crypto_keys::RistrettoPublic;
+    use mc_transaction_core::fog_hint::FogHint;
+    use mc_util_from_random::FromRandom;
     use rand::{rngs::StdRng, SeedableRng};
 
     #[async_test_with_logger]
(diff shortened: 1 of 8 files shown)


full-service/src/service/transaction_builder.rs
```diff
@@ -249,7 +249,87 @@ impl<FPR: FogPubkeyResolver + 'static> WalletTransactionBuilder<FPR> {
 
         let view_account_key = account.view_account_key()?;
         let view_private_key = account.view_private_key()?;
-        let reserved_subaddresses = ReservedSubaddresses::from(&view_account_key);
+        let reserved_subaddresses = match account.account_key() {
+            Ok(account_key) => ReservedSubaddresses::from(&account_key),
+            Err(_) if !account.fog_enabled => ReservedSubaddresses::from(&view_account_key),
+            Err(_) => {
```
eranrund, line 255: This is the inelegant part. If we want to clean this up, we can start storing the fog parameters inside the `Account` row, but that would require a migration, so maybe this is acceptable.
I haven't tested this with a hardware wallet - I am not sure what the flow for that is at the moment, but I did make sure this code appears to be working by commenting out the `match` statement and just running it on my fog-enabled non-view-only account - so at least in that scenario it works as expected.

holtzman (approved):
Looks good to me; I committed changes that make it work with hardware wallets too.

full-service/src/service/transaction_builder.rs
holtzman: hardware wallets were borked due `import_view_only_from_hardware_wallet_with_fog` not creating a fog-enabled subaddress for the change; fixed that and pushed the commit to his branch.

PR #1036, opened 2025-03-12
Validate proof of reserve SCI

eranrund:
### Motivation

We want a way for users to prove they are in possession of specific TxOuts that are unspent. For a proof of reserve, we want to be able to reveal the following about a TxOut in a verifiable way:
1. The amount of the TxOut
2. The fact that the user constructing the proof is able to spend the TxOut
3. The key image of the TxOut, which lets us know whether it has been spent or not.
A valid Signed Contingent Input (e.g. one that has a valid MLSAG signature and contains a TxOut that appears in the ledger) is one possible way to provide such a proof. An SCI is constructed such that it has a ring size of 1 and is made unspendable (due to tombstone block limit or a required output that can never be satisfied, e.g. u64::MAX) allows us to achieve the 3 goals mentioned above.

We need the ability for full-service to verify such specially-constructed SCIs, so this is what this PR achieves.

### In this PR
A new full-service command called `validate_proof_of_reserve` that receives the protobuf-encoded bytes of a SignedContingentInput, and makes sure that it is both valid and constructed in a way that reveals a specific TxOut (by having ring size = 1), is unspendable and is actually present in the ledger.

### Test Plan
* Added unit tests
* Tested manually using `curl` and an SCI that I generated through Sentz as part of some previous work I was doing on proof of reserves.

### Future Work
* It could be nice to add an endpoint for generating proof of reserve SCIs for accounts managed by full-service, but since this is not needed immediately, I opted not to spend time on that.

### DIFF
--- a/full-service/src/service/signed_contingent_input.rs
+++ b/full-service/src/service/signed_contingent_input.rs
@@ -153,3 +153,416 @@ where
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_utils::{append_test_block, get_test_ledger, setup_wallet_service};
+    use mc_account_keys::{AccountKey, DEFAULT_SUBADDRESS_INDEX};
+    use mc_blockchain_types::BlockContents;
+    use mc_common::logger::{async_test_with_logger, Logger};
+    use mc_crypto_keys::RistrettoPublic;
+    use mc_crypto_ring_signature_signer::{NoKeysRingSigner, OneTimeKeyDeriveData};
+    use mc_fog_report_validation_test_utils::MockFogResolver;
+    use mc_transaction_builder::{
+        test_utils::get_ring, EmptyMemoBuilder, InputCredentials, SignedContingentInputBuilder,
+    };
+    use mc_transaction_core::{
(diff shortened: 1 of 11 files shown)


full-service/src/service/signed_contingent_input.rs
```diff
@@ -0,0 +1,572 @@
...
+        if sci.required_output_amounts.len() != 1 {
+            return Ok(ValidateProofOfReserveSciResult::NotProofOfReserveSci {
+                error: "Proof of reserve SCI must have exactly one required output".to_string(),
+            });
+        }
+
+        let required_output_amount = &sci.required_output_amounts[0];
+        if required_output_amount.token_id != sci.pseudo_output_amount.token_id {
+            return Ok(ValidateProofOfReserveSciResult::NotProofOfReserveSci {
+                error: "Proof of reserve SCI must have the same token id as the pseudo-output"
```
holtzman, line 130: nit: for consistency in error wording, does it make sense to align this and the next error?
```suggestion
                error: "Proof of reserve SCI must have the same token id for the required output as for the pseudo-output"
```

holtzman (approved):
Looks good!

PR #1040, opened 2025-03-27
fix secreted txos showing orphaned status

holtzman:
### Motivation

Before this PR `secreted` `txos` were showing a status of `orphaned` unless both counterparties (sender and recipient) were accounts within this instance of full-service, in which case they would show as `secreted` instead of showing their true status, as known to full-service.  `Orphaned` is a different status and showing `secreted` `txos` as `orphaned` is a bug.
### In this PR
* cause `txos` minted  by this instance of full-service to show as `secreted` when they appear in a block and the counterparty is not an account with keys in this instance of full-service
* cause `txos` where the counter parties have different account keys but both sets are known to this instance of full-service to show their true status (`unspent`, `spent`, `unverified`, or `orphaned`) rather than `secreted`, when that txo's status is returned via the `get_txo` or `get_txos` methods.
* fix the `list_secreted` function, which is used by the `get_account_status` and `get_txos` methods to return the `secreted` `txos` for the minting `account_id`, when and `account_id` is provided, including minted `txos` where the wallet has the keys for the counterparty as well as minted `txos` where the wallet does not have the keys of the `txo's` owner.  Previously there was a bug such that `list_secreted` was only including `txos` that went to a counterparty who's private keys were known to this instance of full-service.
* expanded test coverage on `txo::status` and `txo:list_secreted` with the `test_received_txo_lifecycle` test

### Future Work
API methods that return a list of `txos` and their values have always had situations where the API method was requesting `txos` with a particular `status`, and the returned `txos` might have a `status` value that differed from the requested filter. for example, when asking for `txos` minted by a particular `account_id` that were successfully included in the block chain, which is the meaning of `secreted`, `get_txos` could return `txos` with a`status` value with the `txo`'s json struct of `spent` because they are `secreted` from the point of view of the `account_id` passed as a param to this method, this `account_id` having minted the `txo`.  However, this `account_id` is not plumbed down to the call to txo::status used in composing the json response, and if the wallet also has the account keys of the recipient, aka owner of the `txo`, it knowns more about the `txo` and depending on the specific `txo` details would sometimes give a status taking advantage of this knowledge and sometimes not. This PR fixes that half-way. Since `txo::status` still does not have the context of the `account_id` who's perspective should be used in returning a `status` value, it now consistently chooses to reveal the "true" `status` if full-service knows it and only return `secreted` when it does not have those keys and only knows about the `txo` because it minted it and saw it appear on the block chain.

To reiterate, in such cases, the list of returned `secreted` `txos` is correct, but within that list there will be `txo` json structured data with a `status` value other than `secreted` when further status information is available to full-service due to having the account keys of the account that owns that `txo`. This is true both before and after this PR.  Prior to this PR, this would be true only `txos` that full-service knew to be `spent`.  After this PR, `unspent`, `unverified`, and `orphaned` statuses will also appear in `secreted` `txo` json structured data when the wallet knowns those to be the true status of the `txo`.

Future work might plumb a contextual `account_id` to `txo::status` so that full-service can always include the query specific contextual `status` in json structured response data.  This would also allow DRYing up `txo` `status` determination when making lists of `txos` with a particular `status`.  The list functions currently have an `account_id` parameter have mostly redundant logic with `txo::status` so as to apply the context of the supplied `account_id` to filter the list of `txos` known to the wallet to be correct for the combination of `status` and `account_id`.

### DIFF
--- a/full-service/src/db/txo.rs
+++ b/full-service/src/db/txo.rs
@@ -1548,9 +1548,11 @@ impl TxoModel for Txo {
             .filter(transaction_logs::failed.eq(false))
             .filter(transaction_logs::finalized_block_index.is_not_null())
             .filter(transaction_logs::submitted_block_index.is_not_null())
-            .filter(not(transaction_logs::account_id
-                .nullable()
-                .eq(txos::account_id)));
+            .filter(
+                txos::account_id
+                    .is_null()
+                    .or(transaction_logs::account_id.nullable().ne(txos::account_id)),
+            );
 
         if let Some(account_id_hex) = account_id_hex {
             query = query.filter(transaction_logs::account_id.eq(account_id_hex))
@@ -2056,13 +2058,10 @@ impl TxoModel for Txo {
             .left_join(transaction_output_txos::table)
             .left_join(txos::table.on(transaction_output_txos::txo_id.eq(txos::id)))
             .filter(transaction_output_txos::txo_id.eq(&self.id))
-            .filter(transaction_output_txos::is_change.eq(false))
             .filter(transaction_logs::failed.eq(false))
             .filter(transaction_logs::finalized_block_index.is_not_null())
             .filter(transaction_logs::submitted_block_index.is_not_null())
-            .filter(not(transaction_logs::account_id
-                .nullable()
-                .eq(txos::account_id)))
+            .filter(txos::account_id.is_null())
             .select(count(transaction_logs::id))
             .first(conn)?;
 
@@ -2795,9 +2794,6 @@ mod tests {
         .unwrap();
         assert_eq!(unspent.len(), 1);
         assert_eq!(unspent[0].received_block_index.unwrap(), 13);
-        // Store the key image for when we spend this Txo below
-        let for_bob_key_image: KeyImage =
-            mc_util_serial::decode(&unspent[0].key_image.clone().unwrap()).unwrap();
 
         // Note: To receive at Subaddress 4, we need to add an assigned subaddress
         // (currently this Txo is be orphaned). We add thrice, because currently
@@ -2865,6 +2861,22 @@ mod tests {
             .collect();
         assert_eq!(change.len(), 1);
 
+        // Alice shouldn't have any secreted txos
+        let secreted = Txo::list_secreted(
+            Some(&alice_account_id.to_string()),
+            &mut wallet_db.get_pooled_conn().unwrap(),
+        )
+        .unwrap();
+        assert_eq!(secreted.len(), 0);
+
+        // Store the key image for when we spend this Txo below
+        let for_bob_key_images: Vec<KeyImage> = unspent
+            .iter()
+            .filter_map(|txo| txo.key_image.clone())
+            .map(|encoded_key_image| mc_util_serial::decode(&encoded_key_image).unwrap())
+            .collect();
+        assert_eq!(for_bob_key_images.len(), 2);
+
         // Create a new account and send some MOB to it
         let bob_root_id = RootIdentity::from_random(&mut rng);
         let bob_account_key = AccountKey::from(&bob_root_id);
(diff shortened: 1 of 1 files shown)


full-service/src/db/txo.rs
```diff
@@ -1548,9 +1548,11 @@ impl TxoModel for Txo {
...
             .filter(transaction_logs::finalized_block_index.is_not_null())
             .filter(transaction_logs::submitted_block_index.is_not_null())
-            .filter(not(transaction_logs::account_id
-                .nullable()
-                .eq(txos::account_id)));
+            .filter(
+                txos::account_id
+                    .is_null()
+                    .or(transaction_logs::account_id.nullable().ne(txos::account_id)),
+            );
```
holtzman, line 1555: SQL  A != B does not match when one of A or B are NULL. An explicit check for a NULL txos.account_id was needed.

full-service/src/db/txo.rs
```diff
@@ -2056,13 +2058,10 @@ impl TxoModel for Txo {
...
             .left_join(txos::table.on(transaction_output_txos::txo_id.eq(txos::id)))
             .filter(transaction_output_txos::txo_id.eq(&self.id))
-            .filter(transaction_output_txos::is_change.eq(false))
             .filter(transaction_logs::failed.eq(false))
             .filter(transaction_logs::finalized_block_index.is_not_null())
             .filter(transaction_logs::submitted_block_index.is_not_null())
-            .filter(not(transaction_logs::account_id
-                .nullable()
-                .eq(txos::account_id)))
+            .filter(txos::account_id.is_null())
```
holtzman, line 2064: this filter for secreted txos also suffered from the lack of an explicit check for txos.acount_id IS NULL.  We also decide for txo::status, which does not include the context of the account_id from which to consider the status of the txo, to return the status based on all that we know.  When txos.account_id is not null, then we have the account keys for this txo and we can differentiate between spent, unspent, unverified and orphaned.

Better to err on the side of the caller having the perspective of the txos owner rather than minter, and giving them the data they need.

full-service/src/db/txo.rs
eranrund: I think this can be replaced with:
`.filter(transaction_logs::account_id.is_distinct_from(txos::account_id))`
See https://docs.diesel.rs/1.4.x/diesel/expression_methods/trait.PgExpressionMethods.html#method.is_distinct_from (`This behaves identically to the != operator, except that NULL is treated as a normal value.`)

full-service/src/db/txo.rs
eranrund: It might be worth commenting that this query results in secreted transaction logs specifically because the `account_id` field is the **recipient account id** (which I don't find intuitive and am likely to forget this specific definition if I see this code a month from now...). And given that its the recipient account, if it is null and the row indicates the txo was submitted+finalized+not failed, then it means it reached the blockchain and we don't know any more information about it (which is what Secreted stands for)

eranrund (approved):
Nice work and good documentation around these changes.
They are aligned with the conversations we've been having around this and match the PR description.

Btw, you might want to consider using [exists](https://docs.diesel.rs/master/diesel/dsl/fn.exists.html) instead of count. It might be faster.

full-service/src/db/txo.rs
holtzman: thanks for the pointer to this comparison operator! However, it seems to be postgres only or at least not available when the database layer is sqlite, as sqlite lacks the underlying SQL IS DISTINCT FROM.

eranrund: Oh you're right! I forgot full-service is sqlite 🤦

holtzman: yes AND diesel doesn't natively use it -- we could mix in some raw SQL but maybe better to just use the compound query.

PR #1044, opened 2025-04-10
Add TxBlueprint APIs and hardware wallet memo signing.

eranrund:
### Motivation

Right now, we do not include RTH memos in transactions signed by the Ledger hardware wallet. We'd like to start including them.
This originally was not feasible, but with https://github.com/mobilecoinfoundation/mobilecoin/pull/4030 it is now possible.

### In this PR
* Introduce a new `HardwareWalletAuthenticatedMemoHmacSigner` object, which implements the new `AuthenticatedMemoHmacSigner` trait introduced in https://github.com/mobilecoinfoundation/mobilecoin/pull/4030.
* Use  `HardwareWalletAuthenticatedMemoHmacSigner` when converting `TxBlueprint`s to `UnsignedTx`s on hardware-wallet accounts.
* Two new API v2 API endpoints: `build_tx_blueprint` and `sign_and_submit_tx_blueprint`

### Test Plan
* I've tested this locally by hacking the code to print out the `TxOut` it generates inside the `build` function in `full-service/src/service/transaction_builder.rs`, and then used a small utility that I wrote to verify that the memo is correctly formed and that the signature is valid. I then made manual calls to `build_and_submit_transaction`, `build_tx_blueprint` and `sign_and_submit_tx_blueprint` and verified everything behaved correctly.
* I've ran `desktop-wallet` against a build of this `full-service`, set it up with my Ledger hardware account, and similarly verified it can send a transaction (and that I get a print in the logs with the right memo data).
* I've tested both `signer-service` and `hardware-service` locally by calling the `build_tx_blueprint` endpoint on `full-service`, then calling `sign_tx_blueprint` on `signer-service`/`hardware-service`, then calling `submit_transaction`.


eranrund:
This is a script I used to test the `hardware-service` signer:
```python
import requests

data = {
    "method": "build_tx_blueprint",
    "jsonrpc": "2.0",
    "id": 1,
    "params": {
        "account_id": "c20b9bcc5717c3d4666dfefc5648e0d028b1fecbbfd1cd19d9396863b35704c7",
        "addresses_and_amounts": [
            [
                "65G5DuLxeRPp8ben1Tf7vdcj26gWrH4J6NXFDwZkZxAQYy2AVdnRkoXHuFDeSdBSGc5biNkhQbX6VfurVX8mE7AUXziB8oXNMMYHfRWwogy",
                {"value": "10000", "token_id": "1"}
            ]
        ]
    }
}
response = requests.post('http://localhost:9090/wallet/v2', json=data)
json_response = response.json()
tx_blueprint_proposal = json_response['result']['tx_blueprint_proposal']

data = {
    "method": "sign_tx_blueprint",
    "jsonrpc": "2.0",
    "id": 1,
    "params": {
        "tx_blueprint_proposal": tx_blueprint_proposal,
    }
}
response = requests.post('http://localhost:9092/api', json=data)
json_response = response.json()
tx_proposal = json_response['result']['tx_proposal']
account_id = tx_blueprint_proposal['account_id_hex']

data = {
    "method": "submit_transaction",
    "jsonrpc": "2.0",
    "id": 1,
    "params": {
        "tx_proposal": tx_proposal,
        "account_id": account_id,
    }
}
response = requests.post('http://localhost:9090/wallet/v2', json=data)
json_response = response.json()
print(json_response)
```

.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "mobilecoin"]
 	path = mobilecoin
 	url = https://github.com/mobilecoinfoundation/mobilecoin.git
-	
+
 [submodule "ledger-mob"]
 	path = ledger-mob
-	url = https://github.com/mobilecoinofficial/ledger-mob.git
+	url = git@github.com:eranrund/ledger-mob.git
```
eranrund, line 7: This needs to be reverted, I have some small changes to the `ledger-mob` repository but they are currently gated by https://github.com/mobilecoinfoundation/mobilecoin/pull/4030 getting merged, so I am punting on creating the `ledger-mob` PR.

full-service/src/json_rpc/v2/api/wallet.rs
```diff
@@ -446,15 +460,20 @@ where
...
                     input_txo_ids.as_ref(),
                     fee_value,
                     fee_token_id,
                     tombstone_block,
                     max_spendable_value,
+                    // TODO - We can check if the account has an account key to sign with, and if
+                    // so, we can include an RTH memo.
+                    // For hardware-wallet or view-only accounts, we can't include an RTH memo.
+                    // For now the behavior is to just not include a memo when doing offline
+                    // signing.
```
eranrund, line 476: The existing behavior (before this PR) is to not include memos in unsigned transactions, so I kept it. There are a few options here:
1. Leave as-is
2. Include RTH memos only if we're building an unsigned transaction for a non-view-only account
3. Include RTH memos in either case, which will result in the destination memo being generated but not the authenticated sender memo for view-only-accounts.
I don't have a strong feeling one way or the other since I'm not sure how the `build_unsigned_transaction` API is used in the wild.

eranrund:
This has been rebased on top of https://github.com/mobilecoinofficial/full-service/pull/1052/ but it is enormous and difficult to review. I'm going to split it into several smaller PRs.

PR #1045, opened 2025-05-25
(SENTZ-3883) fix get_account_status to calculate secreted balance per-token_id

holtzman:
### Motivation

The get_account_status method is
1. adding together the secreted txo balances across token_ids
2. reporting that same, blended balance repeatedly under each token_id’s block in the response

Instead, it should report a separate secreted total amount for each token_id as that token_id’s secreted balance.

It especially doesn’t make sense to sum this quantity across token_ids as the amounts are in native units for each token_id, which differs from token_id to token_id.

### In this PR
* add optional `token_id` parameter to `list_secreted` txo lookup
* pass through the `token_id` from functions calling `list_secreted` as appropriate
* update `test_received_txo_lifecycle` detect regression

### Test Plan

1. updated `db::txo::tests::test_received_txo_lifecycle` test as per above
2. did test on prod build on account with both MOB and eUSD
   * previously combined secreted balances and showed same secreted value for both tokens
   * now shows different, correct, secreted balances for each
   * still returns all secreted txos via` get_txos` method

### Other
Updated MacOS CI to:
* pin cargo sort to 1.0.9 because newer versions require update to rustc
* pin cmake to 3.31.7 for compatibility with mbedtls crates

### DIFF
--- a/.github/actions/bootstrap-macos/action.yaml
+++ b/.github/actions/bootstrap-macos/action.yaml
@@ -77,3 +77,7 @@ runs:
       echo "AUTH_KEY_PATH=${AUTH_KEY_PATH}" >> "${GITHUB_ENV}"
       echo "CERTIFICATE_PATH=${CERTIFICATE_PATH}" >> "${GITHUB_ENV}"
       echo "KEYCHAIN_PATH=${KEYCHAIN_PATH}" >> "${GITHUB_ENV}"
+
+      # debugging -- list a fragment of code signing identities
+      security find-identity -v -p codesigning | head -1 | awk '{print substr($2, 1, 4) "..." substr($2, length($2)-3)}'
+      
--- a/.github/actions/bootstrap-macos/action.yaml
+++ b/.github/actions/bootstrap-macos/action.yaml
@@ -79,5 +79,5 @@ runs:
       echo "KEYCHAIN_PATH=${KEYCHAIN_PATH}" >> "${GITHUB_ENV}"
 
       # debugging -- list a fragment of code signing identities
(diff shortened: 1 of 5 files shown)


.github/actions/bootstrap-macos/action.yaml
```diff
@@ -25,6 +25,12 @@ runs:
       rm -rf "$(brew --prefix)/var/homebrew/locks"
       brew bundle --quiet
 
+      echo "-- install cmake"
+      curl -LO https://github.com/Kitware/CMake/releases/download/v3.31.7/cmake-3.31.7-macos-universal.tar.gz
+      tar -xzf cmake-3.31.7-macos-universal.tar.gz
+      sudo mv cmake-3.31.7-macos-universal /opt/cmake-3.31.7
+      echo "/opt/cmake-3.31.7/bin" >> "${GITHUB_PATH}"
```
eranrund, line 32: Wondering why this was needed, supposedly `cmake` is installed via brew (https://github.com/mobilecoinofficial/full-service/blob/main/Brewfile#L2)

eranrund: And supposedly there is a new cmake on homebrew: https://formulae.brew.sh/formula/cmake

holtzman: mbedtls-sys-auto v2.26.1 doesn't built with the latest CMake (or any version >= 4.0. CMake removed compatibility with versions of CMake older than 3.5 with the 4.0 release: https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features

PR #1057, opened 2025-10-05
Rename some functions in preparation for a JSON tx blueprint API

eranrund:
People using full-service with a hardware wallet might want to have an offline signing flow for sending transactions that include memos.

Currently, the offline signing flow is done by calling `build_unsigned_transaction` which builds an `UnsignedTxProposal` that is then signed using the standalone signer service (https://github.com/mobilecoinofficial/full-service/blob/main/signer/src/service/api/request.rs#L39). Memos are generated by `build_unsigned_transaction`, which means they do not have access to the account key when a hardware wallet is used.

Using the new `TxBlueprint`, we are able to provide full-service users with an API that produces a `TxBlueprint`, which they will then be able to hand off to the standalone signer service, at which point an `UnsignedTxProposal` followed by a signed `Tx` can be produced.

This PR is the first step in enabling that flow - it contains changes to the `TransactionService` that is used by the JSON API implementation to add support for working with `TxBlueprint`s. This gets used in follow-up PRs.

### DIFF
--- a/full-service/src/json_rpc/v2/api/wallet.rs
+++ b/full-service/src/json_rpc/v2/api/wallet.rs
@@ -23,7 +23,9 @@ use crate::{
                 public_address::PublicAddress,
                 receiver_receipt::ReceiverReceipt,
                 transaction_log::TransactionLog,
-                tx_proposal::{TxProposal as TxProposalJSON, UnsignedTxProposal},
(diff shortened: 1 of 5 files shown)


eranrund:
@nick-mobilecoin oops, I accidentally merged this one before it got reviewed.

